PR #185, opened 2014-09-24
SNMP Alarm Support

plfrancis:


rkd-msw:
@plfrancis, I think this has all been merged by other pull requests, and so we can close this without losing anything. Is that right?

plfrancis:
That should be true. Closing.

PR #188, opened 2014-10-08
Add configuration support for target latency

steelyvoid:
Configuration support for target latency.  The cpp-common change is to add an informational log of the values passed to LoadMonitor.

Tested live, verifying that the configured value for target latency does appear in the sprout/homestead logs


bossmc:
Superseded by #205

PR #225, opened 2014-12-30
[Reviewer: Ellie] Discard DNS responses other than A/AAAA/SRV/NAPTR with a warning

rkd-msw:
Improves the behaviour of #224. Tested by setting my Bono node's upstream_hostname to news.bbc.co.uk and checking the logs:

```
30-12-2014 11:46:40.513 UTC Debug dnscachedresolver.cpp:154: Wait for query responses
30-12-2014 11:46:40.513 UTC Debug dnscachedresolver.cpp:384: Received DNS response for news.bbc.co.uk type A
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:90: Parsing DNS message
000000: c7ff8180 00010003 00000000 046e6577 73036262 6302636f 02756b00 00010001    .... .... .... .new s.bb c.co .uk. ....
000020: c00c0005 00010000 02db0014 076e6577 73777777 03626263 036e6574 02756b00    .... .... .... .new swww .bbc .net .uk.
000040: c02c0001 00010000 00830004 d43af43b c02c0001 00010000 00830004 d43af43a    .,.. .... .... .:.; .,.. .... .... .:.:

30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:95: Parsing header at offset 0x0
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:98: 1 questions, 3 answers, 0 authorities, 0 additional records
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:103: Parsing question 1 at offset 0xc
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:229: Parsed domain name = news.bbc.co.uk, encoded length = 16
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:112: Parsing answer 1 at offset 0x20
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:229: Parsed domain name = news.bbc.co.uk, encoded length = 2
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:282: Resource Record NAME=news.bbc.co.uk TYPE=CNAME CLASS=IN TTL=731 RDLENGTH=20
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:112: Parsing answer 2 at offset 0x40
30-12-2014 11:46:40.514 UTC Debug dnsparser.cpp:229: Parsed domain name = newswww.bbc.net.uk, encoded length = 2
[...]

PR #226, opened 2014-12-30
[Reviewer: Ellie] Further CNAME work

### LINKED ISSUE #224: CNAME DNS responses are interpreted as an IPv6 address
Seen on the mailing list: http://lists.projectclearwater.org/mail-archive/msg01196.html

I did a bit of investigation.

* parse_rr just creates a generic DnsRRecord object if a response is not A, AAAA, SRV or NAPTR (https://github.com/Metaswitch/cpp-common/blob/345dcc6ab28e670ef121a6a7b288f43e1205a7db/src/dnsparser.cpp#L351)

* We only call add_record_to_cache on an A/AAAA result if its domain matches the question (https://github.com/Metaswitch/cpp-common/blob/345dcc6ab28e670ef121a6a7b288f43e1205a7db/src/dnscachedresolver.cpp#L411). So if we got a CNAME for example.com pointing at example.net, and an additional A record for example.net, we'd report getting two records but only add one to the cache - the A record for example.net won't match so gets silently thrown away

* If a record isn't A or AAAA, we assume it's SRV or NAPTR (https://github.com/Metaswitch/cpp-common/blob/345dcc6ab28e670ef121a6a7b288f43e1205a7db/src/dnscachedresolver.cpp#L423) and cache it. It can actually be CNAME.

* At https://github.com/Metaswitch/cpp-common/blob/345dcc6ab28e670ef121a6a7b288f43e1205a7db/src/baseresolver.cpp#L184, we do an A/AAAA lookup and assume all our responses will be A/AAAA responses. One is actually a CNAME response - but we don't spot this, call to_ip46 on it, see that it's not an A record, assume it's an AAAA record (https://github.com/Metaswitch/cpp-common/blob/345dcc6ab28e670ef121a6a7b288f43e1205a7db/src/baseresolver.cpp#L454) and try and read its IPv6 address - which is actually just garbage memory.

I think the right behaviour here is:
* if we get a CNAME saying that example.com is really example.net, we shouldn't throw away any additional A record responses for example.net.
* in fact, if we get a CNAME saying that example.com is really example.net, but don't get any additional records for it, we should probably go and look example.net up.
* we should be a bit more cautious in the parts where we say "if it isn't A/AAAA, assume it's SRV/NAPTR" or "if it isn't A, assume it's AAAA" - we should be checking those assumptions explicitly.

rkd-msw:
Builds on #225 to fix #224 more fully (but I think it's a bit more controversial which is why I pulled #225 out separately). With this change, if we get a DNS answer that contains a CNAME pointing at Y, and then some A/AAAA records for Y, we'll use them. For example, this will work:

```
news.bbc.co.uk          599     IN      CNAME   newswww.bbc.net.uk
newswww.bbc.net.uk      300     IN      A       212.58.246.130
newswww.bbc.net.uk      300     IN      A       212.58.246.129
```

but 

```
newswww.bbc.net.uk      300     IN      A       212.58.246.130
newswww.bbc.net.uk      300     IN      A       212.58.246.129
news.bbc.co.uk          599     IN      CNAME   newswww.bbc.net.uk
```

will result in the two A records being thrown away with a LOG_WARNING. And if we just get:

`news.bbc.co.uk          599     IN      CNAME   newswww.bbc.net.uk`

we won't go away and do further A record lookups on the record pointed to by CNAME, we'll just report finding 0 A/AAAA records. Fortunately, the first style of answer is the only one I've seen, and I think RFC 1034 mandates it:

```
If recursive service is requested and available, the recursive response
to a query will be one of the following:

   - The answer to the query, possibly preface by one or more CNAME
     RRs that specify aliases encountered on the way to an answer.
```

so I think this is a good enough fix for the real world, and even in weird edge cases the behaviour will be reasonable (but not perfect).

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -383,6 +383,9 @@ void DnsCachedResolver::dns_response(const std::string& domain,
   LOG_DEBUG("Received DNS response for %s type %s",
             domain.c_str(), DnsRRecord::rrtype_to_string(dnstype).c_str());
 
+  // Stores the domain pointed to by a CNAME record
+  std::string canonical_domain;
+
   // Find the relevant node in the cache.
   DnsCacheEntryPtr ce = get_cache_entry(domain, dnstype);
 
@@ -407,8 +410,10 @@ void DnsCachedResolver::dns_response(const std::string& domain,
         if ((rr->rrtype() == ns_t_a) ||
             (rr->rrtype() == ns_t_aaaa))
         {
-          // A/AAAA record, so check that RRNAME matches the question.
-          if (strcasecmp(rr->rrname().c_str(), domain.c_str()) == 0)
+          // A/AAAA record, so check that RRNAME matches the question
+          // (or a CNAME).
+          if ((strcasecmp(rr->rrname().c_str(), domain.c_str()) == 0) ||
+              (strcasecmp(rr->rrname().c_str(), canonical_domain.c_str()) == 0))
           {
             // RRNAME matches, so add this record to the cache entry.
             add_record_to_cache(ce, rr);
@@ -425,9 +430,14 @@ void DnsCachedResolver::dns_response(const std::string& domain,
           // SRV or NAPTR record, so add it to the cache entry.
           add_record_to_cache(ce, rr);
         }
+        else if (rr->rrtype() == ns_t_cname)
+        {
+          canonical_domain = ((DnsCNAMERecord*)rr)->target();
+          LOG_DEBUG("CNAME record pointing at %s - treating this as equivalent to %s", canonical_domain.c_str(), domain.c_str());
+        }
         else
         {
-          LOG_WARNING("Ignoring %s record in DNS answer - only A, AAAA, NAPTR and SRV are supported", DnsRRecord::rrtype_to_string(rr->rrtype()).c_str());
+          LOG_WARNING("Ignoring %s record in DNS answer - only CNAME, A, AAAA, NAPTR and SRV are supported", DnsRRecord::rrtype_to_string(rr->rrtype()).c_str());
           delete rr;
         }
       }
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -432,8 +432,20 @@ void DnsCachedResolver::dns_response(const std::string& domain,
         }
         else if (rr->rrtype() == ns_t_cname)
         {
+          // Store off the CNAME value, so that if we see subsequent A
+          // records for the pointed-to name, we'll recognise them.
+          //
+          // This only works for responses that contain a CNAME
+          // followed by other valid records, e.g.
+          // example.net CNAME example.com
+          // example.com A 10.0.0.1
+          //
+          // RFC 1034 mandates this format, so this should be fine.
+          
           canonical_domain = ((DnsCNAMERecord*)rr)->target();
-          LOG_DEBUG("CNAME record pointing at %s - treating this as equivalent to %s", canonical_domain.c_str(), domain.c_str());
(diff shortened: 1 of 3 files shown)


src/dnsparser.cpp
```diff
@@ -346,7 +346,14 @@ int DnsParser::parse_rr(unsigned char* rptr, DnsRRecord*& rr)
...
   }
-  else
+  else if ((rrclass == ns_c_in) && (rrtype == ns_t_cname))
+  {
+    LOG_DEBUG("Parse CNAME record RDATA");
+    std::string target;
+    int target_len = parse_domain_name(rdata, target);
+    rr = new DnsCNAMERecord(rrname, ttl, target);
+  }
+    else
```
eleanor-merry: alignment

src/dnsparser.cpp
```diff
@@ -346,7 +346,14 @@ int DnsParser::parse_rr(unsigned char* rptr, DnsRRecord*& rr)
     }
     rr = (DnsRRecord*)new DnsNaptrRecord(rrname, ttl, order, preference, flags, services, regexp, replacement);
   }
-  else
+  else if ((rrclass == ns_c_in) && (rrtype == ns_t_cname))
+  {
+    LOG_DEBUG("Parse CNAME record RDATA");
+    std::string target;
+    int target_len = parse_domain_name(rdata, target);
```
eleanor-merry: Won't this cause unused warnings?

src/dnscachedresolver.cpp
```diff
@@ -407,22 +410,36 @@ void DnsCachedResolver::dns_response(const std::string& domain,
...
+        else if ((rr->rrtype() == ns_t_srv) ||
+                 (rr->rrtype() == ns_t_naptr))
         {
           // SRV or NAPTR record, so add it to the cache entry.
           add_record_to_cache(ce, rr);
         }
+        else if (rr->rrtype() == ns_t_cname)
+        {
+          canonical_domain = ((DnsCNAMERecord*)rr)->target();
+          LOG_DEBUG("CNAME record pointing at %s - treating this as equivalent to %s", canonical_domain.c_str(), domain.c_str());
```
eleanor-merry: Line length

src/dnscachedresolver.cpp
```diff
@@ -383,6 +383,9 @@ void DnsCachedResolver::dns_response(const std::string& domain,
   LOG_DEBUG("Received DNS response for %s type %s",
             domain.c_str(), DnsRRecord::rrtype_to_string(dnstype).c_str());
 
+  // Stores the domain pointed to by a CNAME record
+  std::string canonical_domain;
```
eleanor-merry: Can you put in a comment about the expected form of the DNS response? I'd like it to be clear that the CNAME result is first and so will be populated before the checks are done on the A/AAAA records.

eleanor-merry:
Code looks good. I'd also like to see a SAS log for the CNAME case (or at least adding into the existing SAS log for A/AAAA lookups that they came via a CNAME).

rkd-msw:
Thanks! I agree with and will apply all your comments, but I'm unlikely to have time to do so for a couple of weeks (and the mailing list user who hit this agrees that it's not needed urgently).

rkd-msw:
I've applied all your suggestions except the SAS log - that's quite complicated because the SAS log in baseresolver.cpp doesn't have access to the information in dnscachedresolver.cpp about whether this was a CNAME lookup.

I think a better and more general approach would be to store the original DNS answer in the cache and SAS log a text form of that (like with ENUM) - that would show whether a record came through CNAME, as well as other useful information. But I think that should be a separate pull request.

rkd-msw:
@eleanor-merry, before I merge, are you happy with me not doing a SAS log here?

Issue #223, opened 2014-12-29
CassandraStore logs with too high severity


mirw:
The CassandraStore classes log with too high a severity when minor things go wrong.  As reported [on the mailing list](http://lists.projectclearwater.org/pipermail/clearwater/2014-December/001957.html), it raises the following errors when the only issue is that an invalid username is trying to register.

```
29-12-2014 18:39:26.005 UTC Error cassandra_store.cpp:431: Cassandra request failed: rc=2, Row 6505550903 not present in column_family impi
29-12-2014 18:39:26.033 UTC Error cassandra_store.cpp:431: Cassandra request failed: rc=2, Row 6505550903 not present in column_family impi
29-12-2014 18:39:29.249 UTC Error cassandra_store.cpp:431: Cassandra request failed: rc=2, Row 6505550903 not present in column_family impi
```

PR #227, opened 2015-01-02
[Reviewer: Ellie] Reduce the log level for Cassandra 'not found' errors

rkd-msw:
Fixes https://github.com/Metaswitch/cpp-common/issues/223. Compiles, but not otherwise tested.


eleanor-merry:
Looks good - can you fix up any UT coverage issues as well

PR #228, opened 2015-01-02
[Reviewer: Andy] Configurable memcached latency

### LINKED ISSUE #38: Memcached store times out connects after 10ms
This is probably fine for local replication, but for the remote store where RTT can be~100ms, this is too short.

We should pass local/remote down to the store so we can use good values for these numbers.

rkd-msw:
Fixes #38. I have a separate branch that increases the timeout to 100ms for Sprout's remote store.

### DIFF
--- a/include/memcachedstore.h
+++ b/include/memcachedstore.h
@@ -64,8 +64,7 @@ public:
   MemcachedStore(bool binary,
                  const std::string& config_file,
                  CommunicationMonitor* comm_monitor = NULL,
-                 Alarm* vbucket_alarm = NULL,
-                 unsigned int max_connect_latency_ms = 50);
+                 Alarm* vbucket_alarm = NULL);
 
   ~MemcachedStore();
 
@@ -76,6 +75,8 @@ public:
                 const std::vector<std::string>& new_servers);
 
   bool has_servers() { return (_servers.size() > 0); };
+
+  void set_max_connect_latency(unsigned int ms);
 
   /// Gets the data for the specified table and key.
   Store::Status get_data(const std::string& table,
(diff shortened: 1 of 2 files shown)


include/memcachedstore.h
```diff
@@ -64,7 +64,9 @@ class MemcachedStore : public Store
   MemcachedStore(bool binary,
                  const std::string& config_file,
                  CommunicationMonitor* comm_monitor = NULL,
-                 Alarm* vbucket_alarm = NULL);
+                 Alarm* vbucket_alarm = NULL,
+                 unsigned int max_connect_latency_ms = 50);
```
bossmc: Preferably this wouldn't be a constructor argument, it would be a setter on the `MemcachedStore` object (or maybe a value as part of a `MemcachedConfig` struct that's passed in to the constructor.  I realise that this is tricky as the `MemcachedStore` currently attempts to create the first view immediately so I'm happy for you to ignore this if you want to get this change in soon.

include/memcachedstore.h
```diff
@@ -165,6 +167,11 @@ class MemcachedStore : public Store
   // The list of servers in this view.
   std::vector<std::string> _servers;
 
+  // The time to wait before timing out a connection to memcached.
+  // (This is only used during normal running - at start-of-day we use
+  // a fixed 10ms time, to start up as quickly as possible).
+  unsigned int _max_connect_latency_ms;
```
bossmc: Actually, this makes my previous point much cheaper, since we use a special value on first boot, we don't need to worry about the timing and can simply add a setter to the Store that may be called at any time to change the time-out for future view creation actions.

bossmc:
One question (across two comments) and you'll need to fix up the merge.

rkd-msw:
Done and done. Will merge when I have time to update all the cpp-commons.

PR #231, opened 2015-01-16
[Reviewer: Andy] Preserve old DNS records if the server doesn't respond

rkd-msw:
We discussed earlier how the existing code probably doesn't behave how it's meant to (because records expire on schedule, so by the time we've realised that the DNS server hasn't responded, we've deleted the old records). This should improve that.

The basic change is that we add 5 minutes to the expiry time when working out when to free records, so there's a period where it has expired (and we'll re-query the DNS server) but is still in memory (and so we can re-use it if the server is down).

I also spotted that the existing code sets `ce->expires = 30` but I think it should be an epoch timestamp - I added a check to `add_to_expiry_list` to ensure this.

I haven't tested this yet (or even compiled it), but wanted to get it out for review to check the approach. In particular, this is meant to preserve existing records indefinitely if the server stays down - do you think that's right?

I'll test by:
* checking that my new warning isn't triggered during normal operation
* checking that DNS records are re-queried and updated during normal operation
* checking that if a DNS server goes down, the DNS record is preserved in the cache
* checking that if the server comes back with a different DNS record, that's picked up
* checking that NXDOMAIN responses don't preserve the existing cached record
* checking that if a record doesn't exist at start-of-day, everything still works

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -116,7 +116,7 @@ DnsResult::~DnsResult()
 }
 
 DnsCachedResolver::DnsCachedResolver(const std::string& dns_server) :
-  _cache_lock(PTHREAD_MUTEX_INITIALIZER),
+  _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
   _cache()
 {
   LOG_STATUS("Creating Cached Resolver using server %s", dns_server.c_str());
@@ -739,13 +739,18 @@ DnsCachedResolver::DnsTsx::~DnsTsx()
 
 void DnsCachedResolver::DnsTsx::execute()
 {
+  // Note that in error cases, ares_query can call the callback
+  // synchronously on the same thread. _cache_lock has to be recursive
+  // to account for this (and it's slightly cleaner to increment
+  // pending_queries first, to stop it going negative).
+  ++_channel->pending_queries;
+  
   ares_query(_channel->channel,
              _domain.c_str(),
              ns_c_in,
              _dnstype,
              DnsTsx::ares_callback,
              this);
-  ++_channel->pending_queries;
 }
 
 void DnsCachedResolver::DnsTsx::ares_callback(void* arg,
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -205,6 +205,7 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
        ++i)
   {
     LOG_VERBOSE("Check cache for %s type %d", (*i).c_str(), dnstype);
+    bool do_query = false;
     if (get_cache_entry(*i, dnstype) == NULL)
     {
       LOG_DEBUG("No entry found in cache");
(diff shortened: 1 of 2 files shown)


src/dnscachedresolver.cpp
```diff
@@ -545,8 +562,18 @@ DnsCachedResolver::DnsCacheEntryPtr DnsCachedResolver::create_cache_entry(const
...
 void DnsCachedResolver::add_to_expiry_list(DnsCacheEntryPtr ce)
 {
-  LOG_DEBUG("Adding %s to cache expiry list with expiry time of %d", ce->domain.c_str(), ce->expires);
-  _cache_expiry_list.insert(std::make_pair(ce->expires, std::make_pair(ce->dnstype, ce->domain)));
+  int sensible_minimum = 1420070400;  // 1st January 2015
+  if ((ce->expires != 0) && (ce->expires < sensible_minimum))
+  {
+    LOG_WARNING("Cache expiry time is %d - expecting either 0 or an epoch timestamp (> %d)",
+                ce->expires,
+                sensible_minimum);
```
bossmc: Can we hit this due to bogus external inputs?  If not, it's a pure code bug and (I reckon) we shouldn't try to recover... it's better to explode and drop a core for diagnosis/debugging.

bossmc:
That looks like the right approach, your test plan sounds sensible too.

src/dnscachedresolver.cpp
rkd-msw: Sensible in principle, but that's not the approach we follow elsewhere (e.g. https://github.com/Metaswitch/sprout/blob/dev/sprout/stack.cpp#L358) and might go against our current trend of trying to crash less (because crashing wipes out all other transactions).

Worth polling the team, though - if everyone agrees that's the right approach, I agree that this is one of the places where we should do that (but we should check all the LOG_WARNING and LOG_ERRORs too).

rkd-msw:
- [ ] checking that my new warning isn't triggered during normal operation
- [ ] checking that DNS records are re-queried and updated during normal operation
- [ ] checking that if a DNS server goes down, the DNS record is preserved in the cache
- [ ] checking that if the server comes back with a different DNS record, that's picked up
- [ ] checking that NXDOMAIN responses don't preserve the existing cached record
- [ ] checking that if a record doesn't exist at start-of-day, everything still works

rkd-msw:
Tested and fixed - can you re-review? Changes since the initial version are:
* special handling of NXDOMAIN
* moving the code into a common function, because it was duplicated

bossmc:
I've added some comments to https://github.com/Metaswitch/cpp-common/commit/748cc83a4a0bb00b8ac7008d4d502279f78f9b41, but otherwise. looks good.

rkd-msw:
OK - I've addressed your comment about asymmetry in the way we discussed (the single-domain dns_query now calls into the multiple-domain dns_query, so I don't need a common_query_logic function to avoid duplication between the two).

I haven't addressed your comment about it being a safe order - that's existing code, and orthogonal to this enhancement (but potentially worth digging into later).

I've also made _cache_lock a recursive mutex - ares_query can call the callback synchronously on the same thread in some error cases (including the ones this change is designed to cope with), which causes non-recursive mutexes to deadlock.

I've tested by making a call, stopping dnsmasq on Sprout, waiting for the TTL (60 secs), making another call, and checking that it succeeds.

PR #234, opened 2015-02-06
Use ipc instead of tcp for 0MQ interface

dtheobald:


AlexHockey:
Looks good. I will merge once we close on the other multiple networks fixes you have submitted.

PR #235, opened 2015-02-09
Fixes for port to Ubuntu 14.04's tools chain (gcc 4.8.2)

dtheobald:
I'm guessing that it was just luck that the code generated by Ubuntu 12.04's tools chain (gcc 4.6.3) got away without having a copy constructor for DnsResult, since the shallow copy performed by the default copy constructor for DnsResult causes a double delete of the pointers stored in the _records vector. Regardless, the double delete causes a crash (sig 11) on loads built with Ubuntu 14.0.4's gcc tool chain. This error prevents an Ubuntu 14.04 built sprout from ever completing its initialization causing an endless crash loop.

Also, I had to add support in the test_iterposer for glibc 2.4.0.

### DIFF
--- a/include/dnscachedresolver.h
+++ b/include/dnscachedresolver.h
@@ -56,6 +56,7 @@ class DnsResult
 public:
   DnsResult(const std::string& domain, int dnstype, const std::vector<DnsRRecord*>& records, int ttl);
   DnsResult(const std::string& domain, int dnstype, int ttl);
+  DnsResult(const DnsResult &obj);
   ~DnsResult();
 
   const std::string& domain() const { return _domain; }
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -55,13 +55,36 @@ DnsResult::DnsResult(const std::string& domain,
   _records(),
   _ttl(ttl)
 {
+  LOG_DEBUG("%x::DnsResult - records.size()=%d", this, records.size());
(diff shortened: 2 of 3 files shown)


AlexHockey:
Ha! Yeah not sure how we got away with that for so long. Good spot. 

I took a look at the existing code and it looks like the only times we copy the `DnsResult` is when we are about to destroy it. I am slightly worried that we are going to unecessarily copy a whole load of DNS entries, and that may affect performance. Could we should use a [move constructor](http://en.cppreference.com/w/cpp/language/move_constructor) instead?

dtheobald:
Yes, I too was concerned about all the copying, but I was in "get it to work" mode and figured I'd leave it up to PC to decide how to fix it with an eye toward better performance.

AlexHockey:
Don't you care about performance too? I appreciate you made this fix when trying to ship a release, but I would expect you to revisit it after the release has shipped. 

Having said that, I would prefer slightly reduced performance to a potential crash, so I will accept this change. However I'm going to raise an issue to cover migrating to a move constructor.

rkd-msw:
I started hitting the double-free issues reliably after some refactoring in dnscachedresolver.cpp, so I've pulled this in to https://github.com/Metaswitch/cpp-common/pull/231 and added a move constructor (https://github.com/Metaswitch/cpp-common/commit/b486f4dbdc0c0a84556ff573ae7010aaff63e731).

PR #237, opened 2015-02-11
[Reviewer: Alex] Close race condition in CapturingTestLogger

rkd-msw:
Some UTs using CapturingTestLogger can crash, if they hit the following race condition:

* one thread starts destroying the CapturingTestLogger
* all the CapturingTestLogger-specific bits are destroyed
* another thread writes to it, corrupting memory
* the base PrintingTestLogger destructor is called, at which point the half-destroyed CapturingTestLogger stops being the global logger

I've fixed quite simply, by adding an alternative "change the logger" method that only works if you're the current logger. This means that it can get called twice in the CapturingTestLogger destructor (once at the beginning, again when the PrintingTestLogger destructor runs) without causing havoc.

Should be safe for non-UT usage - none of the code changed is called outside of UTs.

### DIFF
--- a/include/fakelogger.h
+++ b/include/fakelogger.h
@@ -44,18 +44,18 @@
 #include "log.h"
 #include "logger.h"
 
-/// Logger that prints logged items to stdout.
-/// PrintingTestLogger::DEFAULT should be the only instance needed.
-class PrintingTestLogger : public Logger
+const int DEFAULT_LOGGING_LEVEL = 4;
+
+class BaseTestLogger : public Logger
 {
 public:
-  /// Get a logger with the default behaviour.
-  PrintingTestLogger();
+BaseTestLogger():
+  _last_logger(NULL),
+  _last_logging_level(DEFAULT_LOGGING_LEVEL)
+ {};
 
-  virtual ~PrintingTestLogger()
+  virtual ~BaseTestLogger()
   {
-    Log::swapLogger(this, _last_logger);
-    Log::setLoggingLevel(_last_logging_level);
   }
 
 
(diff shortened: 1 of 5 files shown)


AlexHockey:
Two high level thoughts:

* I don't like that `CapturingTestLogger` subclasses `PrintingTestLogger` (and this is the root cause of the bug). It would be better to have a `TestLogger` class that contains the variables and methods to manage all the swapping shenanigans (but that doesn't call them in its *structor), and have both the actual loggers subclass this. 
* What happens if I do this: create logger1, create logger2, delete logger1, delete logger2. Doesn't logger2 set the (now deleted) logger1 as the logger? Should we actually be implementing this as a logger list? 

Low level comments:

* It feels like `take_over` should have a mirror method that relinquishes control. This could use `SwapLogger` under the covers.
* Would `compareAndSwapLogger` be a better name (given it also does a comparison)?
* You should add some comments explaining that certain functions *must* be called first in the *lowest* class int he hierarchy, otherwise things break.

rkd-msw:
I'd argue that CapturingTestLogger should subclass PrintingTestLogger - it needs to be able to print as well as capture. (In other words, there's nothing in PrintingTestLogger which wouldn't be in your proposed TestLogger.)

Yes `create logger1, create logger2, delete logger1, delete logger2` is an invalid sequence. The old code [explicitly guarded against this](https://github.com/Metaswitch/cpp-common/pull/237/files#diff-e22613ad6a4593a90c6ff13a3d0d3e08L61), but I couldn't see how to make that work. In practice, though, I don't see a need for objects other than PrintingTestLogger and CapturingTestLogger, or a usage pattern other than "global PrintingTestLogger, sometimes a CapturingTestLogger on the stack of specific tests", so think we're unlikely to hit that.

Happy with your low-level comments, will apply.

AlexHockey:
But the current hierarchy has a single object that tries to unset itself twice. That feels like inheritance gone wrong. Happy to move the printing functionality into a superclass, but it feels like we should only ever call `relinquish_control` (or whatever it is called) once per object.

If you move to a model where you only call `relinquish_control` once can't you trivially fix the `create logger1, create logger2, delete logger1, delete logger2` problem? Just assert that the logger being destroyed is the one that's currently in control.

rkd-msw:
Oh. Yes, that is a nicer model - I'll do that. (I think I then don't need the compare-and-swap functionality, because that only exists to guard against the same object trying to unset itself twice.)

PR #239, opened 2015-02-18
Reviewer Alex - Call the new *_vb functions when setting/CASing keys

bossmc:
Alex, if you've not got time to look at these this sprint, please push back.

This simply passed the vbucket ID through on SET-like requests in `MemcachedStore`.  Requires the changes from https://github.com/Metaswitch/libmemcached-upstream/pull/2.


src/memcachedstore.cpp
```diff
@@ -94,8 +94,8 @@ MemcachedStore::MemcachedStore(bool binary,
   // timeout because libmemcached tries to connect to all servers sequentially
   // during start-up, and if any are not up we don't want to wait for any
   // significant length of time.
-  _options = "--CONNECT-TIMEOUT=10 --SUPPORT-CAS";
-  _options += (binary) ? " --BINARY_PROTOCOL" : "";
+  _options = "--SUPPORT-CAS --POLL-TIMEOUT=250";
```
rkd-msw: Why the timeout changes?

src/memcachedstore.cpp
```diff
@@ -612,26 +612,28 @@ Store::Status MemcachedStore::set_data(const std::string& table,
...
       // New record, so attempt to add.  This will fail if someone else
       // gets there first.
-      rc = memcached_add(replicas[replica_idx],
-                         key_ptr,
-                         key_len,
-                         data.data(),
-                         data.length(),
-                         memcached_expiration,
-                         exptime);
+      rc = memcached_add_vb(replicas[replica_idx],
```
AlexHockey: Doesn't this just fail if we use the ASCII protocol? Do we need to zero out the vbucket if we're using the ASCII protocol? Or should we just bin support for ASCII?

PR #241, opened 2015-02-19
[Reviewer: Ellie] Initial health-checker code

rkd-msw:
This is basically the pseudocode from my design with comments added, so should be pretty uncontroversial.

### DIFF
--- a/include/health_checker.h
+++ b/include/health_checker.h
@@ -38,6 +38,7 @@
 #define HEALTH_CHECKER_H
 
 #include <atomic>
+#include <pthread.h>
 
 // Health-checking object which:
 //  - is notified when "healthy behaviour" happens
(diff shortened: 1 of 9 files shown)


include/health_checker.h
```diff
@@ -0,0 +1,77 @@
...
+#include <pthread.h>
+
+// Health-checking object which:
+//  - is notified when "healthy behaviour" happens
+//  - is notified when an exception is hit
+//  - checks every 60 seconds to see if an exception has been hit and
+//    no healthy behaviour has been seen since the last check, and
+//    aborts the process if so.
+//
+// Healthy behaviour is defined as:
```
eleanor-merry: This doesn't feel like the right place for this information - I think this should be split out between the components

src/health_checker.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+#include <cassert>
+#include "time.h"
+
+#include "health_checker.h"
+#include "log.h"
+
+HealthChecker::HealthChecker()
+{
+  _recent_passes = 0;
```
eleanor-merry: Can these go in the initializer list?

src/health_checker.cpp
```diff
@@ -0,0 +1,142 @@
...
+// the process if we have hit an exception and health_check_passed()
+// has not been called since this was last run.
+void HealthChecker::do_check()
+{
+  if (_hit_exception.load() && (_recent_passes.load() == 0))
+  {
+    // LCOV_EXCL_START - only covered in "death tests"
+
+    // LOG_ERROR("Check for overall system health failed - aborting");
+    assert(!"Health check failed");
```
eleanor-merry: Can you use exit() rather than assert?

eleanor-merry:
Reviewed

PR #242, opened 2015-02-24
[Reviewer: Rob] Exception handling

eleanor-merry:
Rob, can you please review these changes to add the exception handler class, and plumb it into the HTTP Stack. (So far) I've live tested exceptions in Sprout in the SIP worker threads

### DIFF
--- a/src/handle_exception.cpp
+++ b/src/handle_exception.cpp
@@ -42,7 +42,7 @@
 
 #include "handle_exception.h"
 
-pthread_key_t _jmp_buf_except;
+pthread_key_t _jmp_buf;
 
 HandleException::HandleException(int ttl, 
                                  bool attempt_quiesce) :
@@ -58,14 +58,14 @@ HandleException::~HandleException()
 void HandleException::handle_exception()
 {
   // Check if there's a stored jmp_buf on the thread and handle if there is
-  jmp_buf* env = (jmp_buf*)pthread_getspecific(_jmp_buf_except);
+  jmp_buf* env = (jmp_buf*)pthread_getspecific(_jmp_buf);
 
   if (env != NULL)
   {
     // Create a child thread, then abort it to create a core file.
     if (!fork())
     {
-      // Set thread local env to null
+      // TODO Set thread local env to null
       abort();
     }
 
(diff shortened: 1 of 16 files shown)


src/handle_exception.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+  // Wait for a random time up to the _ttl. This thread was detached when it 
+  // was created, so we can safely call sleep
+  int sleep_time = rand() % ((HandleException*)det)->_ttl;
+  sleep(sleep_time);
+
+  // Raise a SIGQUIT if needed. 
+  if (((HandleException*)det)->_attempt_quiesce)
+  {
+    raise(SIGQUIT);
```
rkd-msw: I'm not hugely fond of this - I'd rather we passed a QuiescingManager object in here and then triggered it. I can see that this might be a bit painful to refactor, though.

src/httpstack.cpp
```diff
@@ -257,7 +261,18 @@ void HttpStack::handler_callback(evhtp_request_t* req,
...
                 req->uri->path->full,
                 req->uri->query_raw);
-    handler->process_request(request, trail);
+
+    CW_TRY
+      handler->process_request(request, trail);
+    CW_EXCEPT(_handle_exception)
+      send_reply_internal(request, 500, trail); 
+
+      if (_num_threads == 1)
```
rkd-msw: Is this necessary, and is it complete? Surely if our only thread dies, monit will kill us quickly. If it is necessary, should we be decrementing _num_threads, so that if we hit 50 exceptions in quick succession we don't hang?

rkd-msw: We discussed, and I think you're planning to use pthread_exit instead.

src/handle_exception.cpp
```diff
@@ -0,0 +1,103 @@
...
+    sleep(5);
```
rkd-msw: Is 5 seconds long enough for a quiesce?

src/httpstack.cpp
```diff
@@ -257,7 +261,18 @@ void HttpStack::handler_callback(evhtp_request_t* req,
     LOG_VERBOSE("Process request for URL %s, args %s",
                 req->uri->path->full,
                 req->uri->query_raw);
-    handler->process_request(request, trail);
+
+    CW_TRY
```
mirw: Is there any reason you can't write `CW_TRY {` and the `} CW_EXCEPT(...) {`.  The curly braces might not be necessary, but they should be valid, and they make the code far clearer and easier to navigate (e.g. jumping between curly braces is common in most text editors).

rkd-msw:
Reviewed.

PR #243, opened 2015-02-25
[Reviewer: Ellie] Multiple DNS servers

rkd-msw:
Counterpart to https://github.com/Metaswitch/sprout/pull/955

### DIFF
--- a/include/dnscachedresolver.h
+++ b/include/dnscachedresolver.h
@@ -177,7 +177,7 @@ private:
   void wait_for_replies(DnsChannel* channel);
   static void destroy_dns_channel(DnsChannel* channel);
 
-  struct in_addr _dns_server;
+  struct in_addr _dns_servers[3];
 
   // The thread-local store - used for storing DnsChannels.
   pthread_key_t _thread_local;
--- a/include/dnscachedresolver.h
+++ b/include/dnscachedresolver.h
@@ -74,6 +74,7 @@ class DnsCachedResolver
 {
 public:
   DnsCachedResolver(const std::string& dns_server);
+  DnsCachedResolver(const std::vector<std::string>& dns_server);
   ~DnsCachedResolver();
 
   /// Queries a single DNS record.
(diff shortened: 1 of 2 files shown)


include/dnscachedresolver.h
```diff
@@ -177,7 +178,8 @@ class DnsCachedResolver
   void wait_for_replies(DnsChannel* channel);
   static void destroy_dns_channel(DnsChannel* channel);
 
-  struct in_addr _dns_server;
+  struct in_addr _dns_servers[3];
```
eleanor-merry: Can max_size be defined and used here and in the constructor

src/dnscachedresolver.cpp
```diff
@@ -83,6 +83,34 @@ DnsResult::~DnsResult()
...
+  _dns_servers_count = std::min(max_servers, dns_servers.size());
+
+  LOG_STATUS("Creating Cached Resolver using servers:");
+  for (size_t i = 0; i < _dns_servers_count; i++)
+  {
+    LOG_STATUS("    %s", dns_servers[i].c_str());
+    // Parse the DNS server's IP address.
+    if (!inet_aton(dns_servers[i].c_str(), &(_dns_servers[i])))
+    {
+      LOG_ERROR("Failed to parse '%s' as IP address - defaulting to 127.0.0.1", dns_servers[i].c_str());
```
eleanor-merry: Will there be any issues if multiple DNS servers are set to 127.0.0.1?

eleanor-merry:
Looks good

include/dnscachedresolver.h
rkd-msw: Not nicely - if I define DNSResolver::MAX_SERVERS that can't be used as an array size, and I don't really want to `#define` it.

src/dnscachedresolver.cpp
rkd-msw: Probably not, but I'll check.

PR #244, opened 2015-02-26
[Reviewer: Andy] Memcached tombstones

AlexHockey:
Andy, please can you review my enhancement to `MemcachedStore` to write tombstones instead of deleting records. This should match the design you reviewed last week. 

Tested as follows:

* Scripted tests (using the FV framework I've been developing). Unless otherwise stated these all work regardless of whether the system is configured to use tombstones. 
  * `set_data`. `get_data` returns the record. `delete_data`. `get_data` returns `NOT_FOUND`
  * `set_data`. `get_data` to get the CAS. `set_data` with CAS. `get_data` returns the second record.
  * `set_data`. `get_data` to get the CAS. `set_data` with the wrong CAS. `get_data` returns the first record.
  * `set_data` with CAS=0. `set_data` with CAS=0, returns `DATA_CONTENTION`. 
  * `set_data`. `delete_data`. `set_data`. `get_data` returns second record.
  * (this one only works when tombstones are configured) `set_data`, `delete_data`, an add via memcached fails with `MEMCACHED_NOTSTORED`.

* Live tests:
  * Live tests run cleanly
  * I can sucessfully retrieve a call list from Memento using curl. 
  * I set up Ralf and a CDF to request an interim interval of 20s. Made a call, and checked that interims were delivered at the correct rate. 

As discussed in scrum, no docs updates are needed.


src/memcachedstore.cpp
```diff
@@ -165,6 +173,82 @@ void MemcachedStore::new_view(const std::vector<std::string>& servers,
...
+          }
+        }
+      }
+    }
+    f.close();
+
+    if (servers.size() > 0)
+    {
+      LOG_DEBUG("Update memcached store");
+      this->new_view(servers, new_servers);
```
bossmc: Why the `this`?

src/memcachedstore.cpp
```diff
@@ -165,6 +173,82 @@ void MemcachedStore::new_view(const std::vector<std::string>& servers,
...
+        else if (tokens[0] == "tombstone_lifetime")
+        {
+          // Read the tombstone lifetime from the config file. Check it is
+          // actually a valid integer before committing to the member variable
+          // (atoi stops when it reaches non-numeric characters).
+          int lifetime = atoi(tokens[1].c_str());
+
+          if (std::to_string(lifetime) == tokens[1])
+          {
+            _tombstone_lifetime = lifetime;
```
bossmc: Does this need locking?  Or should it be atomic?

src/memcachedstore.cpp
```diff
@@ -165,6 +173,82 @@ void MemcachedStore::new_view(const std::vector<std::string>& servers,
...
+          }
+          else
+          {
+            LOG_ERROR("'%s' does not contain a valid tombstone lifetime - keeping previous setting",
```
bossmc: Log the bogus value?  And the old setting that will be used?

src/memcachedstore.cpp
```diff
@@ -165,6 +173,82 @@ void MemcachedStore::new_view(const std::vector<std::string>& servers,
...
+    f.close();
+
+    if (servers.size() > 0)
+    {
+      LOG_DEBUG("Update memcached store");
+      this->new_view(servers, new_servers);
+    }
+    else
+    {
+      LOG_ERROR("'%s' does not contain a valid set of servers - keeping previous settings", _config_file.c_str());
```
bossmc: Ditto re: logs.

src/memcachedstore.cpp
```diff
@@ -401,40 +485,21 @@ Store::Status MemcachedStore::get_data(const std::string& table,
...
-        // char*, so this is safe.
-        memcached_result_free(&result);
-        break;
-      }
-      else
-      {
-        // Free the result and continue the read loop.
-        memcached_result_free(&result);
-      }
+    if (rc == MEMCACHED_SUCCESS)
```
bossmc: Should this be `memcached_success(rc)`?  I think yes (as `MEMCACHED_SOME_ERRORS` is a successful return code).

src/memcachedstore.cpp
```diff
@@ -453,29 +518,56 @@ Store::Status MemcachedStore::get_data(const std::string& table,
...
+      {
+        SAS::Event got_tombstone(trail, SASEvent::MEMCACHED_GET_TOMBSTONE, 0);
+        got_tombstone.add_var_param(fqkey);
+        got_tombstone.add_static_param(cas);
+        SAS::report_event(got_tombstone);
+      }
+
+      // We have read a tombstone. Return NOT_FOUND to the caller, and also
+      // zero out the CAS (returning a non-zero CAS makes the interface
+      // cleaner).
```
bossmc: Weirdly contradictory comment, please clarify.

src/memcachedstore.cpp
AlexHockey: Good spot. It probably should be atomic.

src/memcachedstore.cpp
```diff
@@ -603,22 +696,24 @@ Store::Status MemcachedStore::set_data(const std::string& table,
...
+      // New record, so attempt to add (but overwrite any tombstones we
+      // encounter).  This will fail if someone else got there first and some
+      // data already exists in memcached for this key.
+      rc = add_overwriting_tombstone(replicas[replica_idx],
+                                     key_ptr,
+                                     key_len,
+                                     data,
+                                     memcached_expiration,
+                                     exptime,
+                                     trail);
```
bossmc: Why no equivalent log to https://github.com/Metaswitch/cpp-common/pull/244/files#diff-30c129b561cf0ea4cdd2d115fa5011baR731-737?

src/memcachedstore.cpp
AlexHockey: Not my code (I just moved it), but I'll fix.

src/memcachedstore.cpp
AlexHockey: I'm not sure. Perhaps the code was templated at some point? I'll remove.

src/memcachedstore.cpp
AlexHockey: Sure thing.

src/memcachedstore.cpp
AlexHockey: Yes.

src/memcachedstore.cpp
AlexHockey: "non-zero" should read "zero" :-/

src/memcachedstore.cpp
AlexHockey: `add_overwriting_tombstone` does the necessary logging. 

p.s. that link didn't work at all!

bossmc:
Ok, looks pretty good.

PR #247, opened 2015-02-27
[Reviewer: Rob] Add in wrapper around thread pool

eleanor-merry:
Rob, can you review this change to the thread pool as well.

### DIFF
--- a/include/cassandra_store.h
+++ b/include/cassandra_store.h
@@ -177,22 +177,25 @@ public:
 
   /// Configure the store.
   ///
-  /// @param cass_hostname   - The hostname for the cassandra database.
-  /// @param cass_port       - The port to connect to cassandra on.
-  /// @param num_threads     - The number of worker threads to use for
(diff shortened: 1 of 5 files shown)


include/cassandra_store.h
```diff
@@ -257,13 +260,22 @@ class Store
   /// @return                - The current time (in micro-seconds).
   static int64_t generate_timestamp();
 
+  static void exception_callback(std::pair<Operation*, Transaction*> work)
+  {
+    // No recovery behaviour
```
rkd-msw: Worth being more verbose as to why this isn't needed?

rkd-msw:
Looks good. I'm not very clear on what kind of error would trigger this, though - can you note what you've done to test this in this pull request, for future reference?

eleanor-merry:
Sure - it's used by the Cassandra store and the memento HTTP stack. I've tested by adding segfaults to the relevant code paths and checking that the exception is caught and the correct recovery behaviour is run.

PR #249, opened 2015-03-02
[Reviewer: Alex] Drop special handling of CURLE_COULDNT_CONNECT

### LINKED ISSUE #200: HttpConnection does not retry if a target server does not accept a connection.
I set up a deployment with two homestead nodes. I stopped the homestead process on one of them and tried to register. Sprout tried to contact the stopped node, then failed the register. 

I think this is due to the code in https://github.com/Metaswitch/cpp-common/blob/master/src/httpconnection.cpp#L638, which bails out of the retry loop if curl returns `CURLE_COULDNT_CONNECT`. 

Looking at some old code there is a comment about `CURLE_COULDNT_CONNECT` meaning that curl has already retried (https://github.com/Metaswitch/cpp-common/blob/d8277c8e96681fe810f7fa193e1caf9b9f51466c/src/httpconnection.cpp#L521). This is no longer true, as we now pass curl a URL containing an IP address.

rkd-msw:
Fixes #200 . I'm pretty confident that this is the fix, but I'll test by spinning up two Homestead nodes, stopping one, and ensuring that the live tests still pass.

### DIFF
--- a/src/httpconnection.cpp
+++ b/src/httpconnection.cpp
@@ -632,8 +632,7 @@ HTTPCode HttpConnection::send_request(const std::string& path,       //< Absolut
       // Decide whether to keep trying.
       if ((num_http_503_responses + num_timeouts_or_io_errors >= 2) ||
           (num_http_504_responses >= 1) ||
-          fatal_http_error ||
-          (rc == CURLE_COULDNT_CONNECT))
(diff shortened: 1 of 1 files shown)


AlexHockey:
Looks good, as does your test plan.

PR #250, opened 2015-03-02
[Reviewer: Matt] Handle the case where we erase our peer and end up with an empty list

### LINKED ISSUE #107: Log spam (and bug) when Diameter connection fails
I'd misconfigured my CDF such that it rejected ralf's connection.

In the logs, I see:

```
22-05-2014 20:10:04.693 Warning diameterstack.cpp:263: Failed to connect to 10.151.82.163
22-05-2014 20:10:04.693 Error realmmanager.cpp:81: Failed to connect to peer 10.151.82.163
22-05-2014 20:10:04.693 Error diameterstack.cpp:276: Unexpected host on callback (type 10) from freeDiameter: 10.151.82.163
22-05-2014 20:10:04.693 Error freeDiameter: ERROR: in '(fd_p_ce_msgrcv(&msg, (hdr->msg_flags & 0x80), peer))' : Invalid argument
```

This is quite a few warning/error logs for a single failure.  In particular, the last 2 just seem wrong, and suggest there might be a bug here.

rkd-msw:
Fixes one of the two unhelpful logs you spotted in #107. I don't think the second log is worth fixing (per my analysis in #107) - do you agree that we can close #107 after this fix?

I'll test by configuring Homestead with `hss_hostname=1.2.3.4` or something equally spurious, and checking the logs.

### DIFF
--- a/src/diameterstack.cpp
+++ b/src/diameterstack.cpp
@@ -267,12 +267,14 @@ void Stack::fd_peer_hook_cb(enum fd_hook_type type,
     DiamId_t realm = peer->info.runtime.pir_realm;
     pthread_mutex_lock(&_peers_lock);
     std::vector<Peer*>::iterator ii;
+    bool peer_found = false;
     for (ii = _peers.begin();
          ii != _peers.end();
          ii++)
     {
       if ((*ii)->host().compare(host) == 0)
       {
+        peer_found = true;
         if ((*ii)->listener())
         {
           if (type == HOOK_PEER_CONNECT_SUCCESS)
@@ -304,9 +306,8 @@ void Stack::fd_peer_hook_cb(enum fd_hook_type type,
       }
     }
 
-    if (ii == _peers.end())
+    if (!peer_found)
     {
-      // Peer not found.
       LOG_ERROR("Unexpected host on callback (type %d) from freeDiameter: %s", type, host);
     }
     pthread_mutex_unlock(&_peers_lock);


mirw:
The fix looks good.

By "the second log" are you referring to the `fd_p_ce_msgrcv` error?  What condition is triggering this `EINVAL` and how often do we get errors from freeDiameter actually dropping out here?  It may well be that we can leave this log in, but I don't think I follow the logic yet.

rkd-msw:
Yes, I'm referring to the fd_p_ce_msgrcv error. It returns EINVAL whenever it gets a return code that isn't "success", "too busy" or "election lost", and returning a non-0 value triggers the cleanup code in https://github.com/Metaswitch/freeDiameter/blob/954a518fbf4cd836f3fc9937d7a8c8b79dca3d7f/libfdcore/p_psm.c#L630 to run.

This function is specific to CEAs, so we'd only see errors here at start-of-day on a misconfiguration. We'd see the same log if the CEA's Result-Code was any other value (e.g. 5012 UNABLE_TO_COMPLY), so changing it to ECONNREFUSED would potentially be misleading.

So:
* it's not safe to return 0
* it's potentially misleading to return things other than EINVAL
* you'd hit this particular log very rarely, only at start-of-day with a misconfiguration, where plenty of other diagnostics (other error logs, ENT logs) are available to help pinpoint the problem

mirw:
Thanks - that's much clearer.  I'm happy with this.  Feel free to merge and close the issue.

PR #251, opened 2015-03-02
[Reviewer: Andy] Interface for MemcachedConfig code

AlexHockey:
This change defines the interface to the memcached config reading code I am writing. Please could you take a look at it and let me know if this works for you? 

Note that I have defined separate classed for `MemcachedConfigReader` (the interface) and `MemcachedConfigFileReader` (the implementation). I think this is cleaner and encourages looser coupling between the classes that use the config. For the same reason, I have also taken your suggestion of changing `MemcachedStoreView::update` to take a config object.

### DIFF
--- a/include/memcached_config.h
+++ b/include/memcached_config.h
@@ -38,6 +38,9 @@
 #ifndef MEMCACHED_CONFIG_H_
 #define MEMCACHED_CONFIG_H_
 
+#include <vector>
+#include <string>
+
 
 /// Structure holding memcached configuration parameters.
 struct MemcachedConfig
@@ -77,12 +80,12 @@ public:
   virtual bool read_config(MemcachedConfig& config) = 0;
 
   /// Virtual destructor.
-  virtual ~MemcachedConfigReader();
+  virtual ~MemcachedConfigReader() {};
 };
 
 
 /// Class that reads the memcached config from a file on disk.
-class MemcachedConfigFileReader
+class MemcachedConfigFileReader : public MemcachedConfigReader
 {
 public:
   /// Constructor
--- a/include/memcached_config.h
+++ b/include/memcached_config.h
@@ -1,0 +1,103 @@
+/**
+ * @file memcached_config.h Classes to handle reading and parsing memcached
+ * configuration.
+ *
+ * Project Clearwater - IMS in the cloud.
+ * Copyright (C) 2015  Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version, along with the "Special Exception" for use of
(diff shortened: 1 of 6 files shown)


AlexHockey:
BTW, please don't merge this :-)

include/memcached_config.h
```diff
@@ -0,0 +1,103 @@
...
+  std::vector<std::string> new_servers;
+
+  /// The expiry time (in seconds) of tombstone records that are written when
+  /// data is deleted from memcached.
+  ///
+  /// 0 means that tombstones are not used. Deleted data is simply removed.
+  int tombstone_lifetime;
+
+  /// Default constructor.
+  MemcachedConfig();
```
bossmc: I don't think you need a default constructor unless the default value for one of the variables is not that variable's default value.  In this case I assume the defaults are empty lists and 0 which would be the automatic defaults.

bossmc:
Looks good.  Happy for you to go ahead with implementation.

include/memcached_config.h
AlexHockey: Not quite. The default is currently 0 but will be changing to a non-zero value in the next Project Clearwater release. (Also, I *think* that POD types are never default-initialized anyway, but I could be wrong about that).

bossmc: Ok, sounds reasonable.

AlexHockey:
@bossmc, I've added the implementation. Please could you review this as well?

include/memcachedstore.h
```diff
@@ -49,30 +49,35 @@ extern "C" {
...
+//
+// This file contains the definition of a memcached-based store. This is split
+// across two classes:
+// -  BaseMemcachedStore which contains the majority of the code.
+// -  MemcachedStore. This is the class that users of the store should
+//    create.
+//
+// MemcacheStore is separate class as this allows different constructors to be
+// defined which all delegate to the BaseMemcachedStore class. This is required
+// as BaseMemcachedStore contains constant memebers, and our version of GCC
```
bossmc: sp: "memebers"

include/memcachedstore.h
```diff
@@ -49,30 +49,35 @@ extern "C" {
...
+// does not support delegating to constructors in the same class.
```
bossmc: You could use the old `init` function trick to commonalize construction, but what you've done is fine (shame we don't have proper C++11 support).

bossmc:
Looks fine.  One spelling mistake.  I'll fix it when pull into my integration branch unless you want to rush this into trunk?

include/memcachedstore.h
AlexHockey: Actually I can't. the old `BaseMemcachedStore` has some constant members that have to be initialized in the initializer list.

AlexHockey:
Cheers. I have no urgent desire to get this into trunk on it's own (and there is no point both of us updating the submodule refs), so I'm closing the PR.

PR #252, opened 2015-03-05
[Reviewer: Andy] Fix deadlock seen when two threads look up the same nam...

rkd-msw:
...e and take _cache_lock and ce->lock in different orders

There's a deadlock in DnsCacheResolver, introduced/worsened by https://github.com/Metaswitch/cpp-common/pull/231/files, which is reasonably easy to reproduce when two threads are trying to look up the same name at the same time (e.g. if a DNS server is being slow).

I’ve fixed this by removing the per-DnsCacheEntryPtr lock and just relying on the cache lock. To ensure this is safe, I've:
* tried to make sure that `get_cache_entry` and `create_cache_entry` are only called when the cache lock is held
* tried to make sure that any DnsCacheEntryPtrs I get while holding the lock are not used after releasing that lock
 
(as a corollary of those two, all modifications to a DnsCacheEntryPtr happen while the cache lock is held)

There was one bit where we released the cache lock, and relied on being able to take the DnsCacheEntryPtr lock as a way to check that another thread had finished doing work. I’ve replaced that with a condition variable (and recognised that waiting on a condition variable implicitly releases and re-takes the lock, so treated any previously-owned DnsCacheEntryPtrs as invalid and looked them up again).

I've tested by pointing a Sprout node at a nonexistent DNS server (so that it always times out) and sending 50 simultaneous REGISTERs into Sprout (so that 50 threads try and look up hs.homedomain at once). That reliably deadlocks on master, but doesn't do so with this fix.

(I also spotted that I changed one constructor to initialize a recursive mutex but not the other - I've corrected that while in the area.)

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -117,7 +117,6 @@ DnsResult::~DnsResult()
 
 DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers) :
   _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
-  _got_reply_cond(PTHREAD_COND_INITIALIZER),
   _cache()
 {
   // Initialize the ares library.  This might have already been done by curl
@@ -142,11 +141,16 @@ DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers
   // We store a DNSResolver in thread-local data, so create the thread-local
   // store.
   pthread_key_create(&_thread_local, (void(*)(void*))&destroy_dns_channel);  
+
+  pthread_condattr_t cond_attr;
+  pthread_condattr_init(&cond_attr);
+  pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC);
+  pthread_cond_init(&_got_reply_cond, &cond_attr);
+  pthread_condattr_destroy(&cond_attr);
 }
 
 DnsCachedResolver::DnsCachedResolver(const std::string& dns_server) :
   _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
-  _got_reply_cond(PTHREAD_COND_INITIALIZER),
   _cache()
 {
   LOG_STATUS("Creating Cached Resolver using server %s", dns_server.c_str());
@@ -167,6 +171,12 @@ DnsCachedResolver::DnsCachedResolver(const std::string& dns_server) :
   // We store a DNSResolver in thread-local data, so create the thread-local
   // store.
   pthread_key_create(&_thread_local, (void(*)(void*))&destroy_dns_channel);
+
+  pthread_condattr_t cond_attr;
+  pthread_condattr_init(&cond_attr);
+  pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC);
+  pthread_cond_init(&_got_reply_cond, &cond_attr);
+  pthread_condattr_destroy(&cond_attr);
 }
 
 DnsCachedResolver::~DnsCachedResolver()
@@ -561,7 +571,7 @@ void DnsCachedResolver::dns_response(const std::string& domain,
 
   // Another thread may be waiting for our query to finish, so
   // broadcast a signal to wake it up.
-  pthread_cond_signal(&_got_reply_cond);
+  pthread_cond_broadcast(&_got_reply_cond);
 
   pthread_mutex_unlock(&_cache_lock);
 }
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -116,7 +116,8 @@ DnsResult::~DnsResult()
 }
 
 DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers) :
-  _cache_lock(PTHREAD_MUTEX_INITIALIZER),
+  _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
+  _got_reply_cond(PTHREAD_COND_INITIALIZER),
   _cache()
 {
   // Initialize the ares library.  This might have already been done by curl
@@ -145,6 +146,7 @@ DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers
 
 DnsCachedResolver::DnsCachedResolver(const std::string& dns_server) :
   _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
+  _got_reply_cond(PTHREAD_COND_INITIALIZER),
   _cache()
 {
   LOG_STATUS("Creating Cached Resolver using server %s", dns_server.c_str());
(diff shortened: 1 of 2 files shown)


mirw:
I just wanted to check that we weren't increasing contention on the global cache lock.  From a very quick code skim, I don't think we are - can you confirm?

rkd-msw:
Not in the general case. There's once case where we are, a bit - if threads 1-5 are waiting on a DNS query for A, threads 6-10 are waiting on a DNS query for B, and threads 11-15 are waiting on a DNS query for C, then I think threads 2-5, 7-10 and 12-15 will all wait at https://github.com/Metaswitch/cpp-common/pull/252/files#diff-b76e44fc3db821f376c4d58277322bc6R280, and when the result for A comes in, threads 2-5 will wake up and carry on, but threads 7-10 and 12-15 will wake up, take the lock again, see that their result hasn't actually come in yet, and release the lock and start waiting again.

I can't see how to avoid that, though, and I think it's very much an edge case - both because a lot of the time responses will be served from cache, and because processing is pretty linear. If a server is being slow returning a result for hs.example.com, all the threads will probably be blocked waiting for that result - it won't be the case that some threads have got past the Homestead processing and are now blocked on a Homer or Ralf DNS query.

src/dnscachedresolver.cpp
```diff
@@ -116,7 +116,8 @@ DnsResult::~DnsResult()
 }
 
 DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers) :
-  _cache_lock(PTHREAD_MUTEX_INITIALIZER),
+  _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
+  _got_reply_cond(PTHREAD_COND_INITIALIZER),
```
bossmc: You probably want to configure this `condvar` to use `CLOCK_MONOTONIC`.  Even if we don't currently use `timedwait` on this, we might in the future and this is a trap waiting to happen.

src/dnscachedresolver.cpp
```diff
@@ -145,6 +146,7 @@ DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers
 
 DnsCachedResolver::DnsCachedResolver(const std::string& dns_server) :
   _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
+  _got_reply_cond(PTHREAD_COND_INITIALIZER),
```
bossmc: Ditto.

src/dnscachedresolver.cpp
```diff
@@ -821,7 +820,7 @@ void DnsCachedResolver::DnsTsx::ares_callback(void* arg,
 
 void DnsCachedResolver::DnsTsx::ares_callback(int status, int timeouts, unsigned char* abuf, int alen)
 {
-  --_channel->pending_queries;
   _channel->resolver->dns_response(_domain, _dnstype, status, abuf, alen);
+  --_channel->pending_queries;
```
bossmc: Why this change?  (Not that it looks wrong, just unexpected).

src/dnscachedresolver.cpp
```diff
@@ -561,7 +558,10 @@ void DnsCachedResolver::dns_response(const std::string& domain,
   // Flag that the cache entry is no longer pending a query, and release
   // the lock on the cache entry.
   ce->pending_query = false;
-  pthread_mutex_unlock(&ce->lock);
+
+  // Another thread may be waiting for our query to finish, so
+  // broadcast a signal to wake it up.
+  pthread_cond_signal(&_got_reply_cond);
```
bossmc: Signal or broadcast?  If we have multiple people waiting on DNS responses, we can't guarantee that the right one will be woken by this signal (and if it's the wrong one, it will go back to sleep and real one will never recover).

Using broadcast will increase the thrash on the cache lock (as everyone will wake up, check the `request_pending` flag and go back to sleep.  Maybe the requesting threads should provide a condvar to signal, associated with their request (or one associated with the `ce` that we broadcast, waking up everyone who cares about that entry... this seems the nicest)?  That way we'll always wake the correct thread(s) (and no others).

bossmc:
One big comment, I think you've swapped a deadly embrace for a starvation race.

src/dnscachedresolver.cpp
rkd-msw: Yes, should be broadcast. I don't think we can have a condvar associated with the `ce`, can we? That would risk the condvar disappearing from under the holder's feet if the `ce` expired while it was waiting for the signal.

Per https://github.com/Metaswitch/cpp-common/pull/252#issuecomment-77345713, I think you'd only hit that if you had multiple threads waiting on different DNS responses, and I'm not sure that's going to be very common.

src/dnscachedresolver.cpp
rkd-msw: I don't think it makes a lot of difference, but if we have things looping `while (pending_queries > 0)`, presumably pending_queries shouldn't hit 0 until the query has actually been processed and the last `ce->pending_query` has been set to false.

bossmc: Seems reasonable.

src/dnscachedresolver.cpp
bossmc: Hmmm... I feel we should be able to do better but `broadcast` at least gives working behaviour.  Might be worth raising a low-priority performance issue?

rkd-msw: I'd like to have some idea of how to do better before raising an issue - otherwise it will just be a generic "maybe we could do better" issue with no clear resolution. The basic problem is how to have a condvar such that, when thread 1 looks up "hs.example.com", it signals that condvar, and when thread 2 sees that a lookup for "hs.example.com" is pending, it waits on that condvar.

The only possibilities I can see are:
* a global condvar, so everything waits on the same one
* a condvar in DnsCacheEntry, but that might be freed while still in use if the TTL is 0 and the DnsCacheEntry gets freed immediately
* a global std::map of domain names to condvars, but I can't immediately see how you'd know that it was safe to free the condvars

graemerobertson:
I've just seen this causing problems on memento and raised an issue for it. https://github.com/Metaswitch/cpp-common/issues/253.

src/dnscachedresolver.cpp
bossmc: Or:

* Attaching a condvar to a DNS query, provided by the waiting thread.
* Preventing the `ce` being destroyed while there are requests pending on it (smart_ptr/reference counting)

rkd-msw: smart_ptr/reference-counting sounds good to me - I'll raise an issue to implement this which would reduce lock contention.

rkd-msw: (I'm not sure I understand "Attaching a condvar to a DNS query, provided by the waiting thread" - surely subsequent threads that need to wait don't have access to the original DNS query?)

bossmc: We don't elide multiple requests for the same domain name (I don't think) so each DNS client would kick off its own request, attach a `condvar` and sleep on it.

rkd-msw: Surely we do? Look at https://github.com/Metaswitch/cpp-common/blob/dns_fix_deadlock/src/dnscachedresolver.cpp#L203 - the first thread will call get_cache_entry, get NULL, call create_cache_entry, send the query, and unlock the cache lock while it waits for replies. Second thread calls get_cache_entry, gets a non-NULL value (because create_cache_entry created it), so it doesn't send a new query and skips the processing until it hits https://github.com/Metaswitch/cpp-common/blob/dns_fix_deadlock/src/dnscachedresolver.cpp#L276, where it sees that ce->pending_query is true and waits for the original request for that domain name to complete.

PR #256, opened 2015-03-09
Reviewer AMC:  No dns query if outstanding upstream

steelyvoid:
This fixes a problem in DNS queries whereby, if a query is outstanding at the point that another query comes in for the same domain, the second query generates and sends a request to the DNS server rather than waiting for the first query to complete (as the rest of the code intends, in order to avoid a "DNS storm" when a record for a local resource - such as the HSS - hits its TTL).  

Tested by firing 100 simultaneous REGISTERS repeatedly (10 - 20 seconds between each burst) at a simplex system whose DNS record for the homestead server is hosted by a DNS server that takes at least 50 ms to respond and sets a TTL of only 8 seconds (thereby ensuring that we regularly get many simultaneous requests for the same expired record.

Used debug logging to confirm that the pending query condition was being hit repeatedly and gave rise to a single real DNS request with all pended threads picking up the response from that single query.


src/dnscachedresolver.cpp
```diff
@@ -287,8 +296,10 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
     {
       // We must release the global lock and let the other thread finish
       // the query.
+      LOG_DEBUG("Pending query");
```
bossmc: I'm a little concerned that this log doesn't give enough context to be useful in the general case.  I think it would be better if it was clear what query we were waiting for and the fact that we're blocked on a real DNS query, not a cached one, something like:

    LOG_DEBUG("Waiting for (non-cached) DNS query for %s", i->c_str());

src/dnscachedresolver.cpp
```diff
@@ -287,8 +296,10 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
...
       pthread_cond_wait(&_got_reply_cond, &_cache_lock);
       ce = get_cache_entry(*i, dnstype);
+      LOG_DEBUG("Reawoken from wait for %s type %d", i->c_str(), dnstype);
```
bossmc: It would be nicer if we could drop a different log in the case where we're prematurely woken from the case where we're woken because the `ce` has been updated.  Then we can make the proper wake-up log more descriptive (similar to my previous comment).

Can you have a think about whether there's a sensible way to do this?  I think the existence of your logs are useful, but the content doesn't give enough context to tie them together properly.

src/dnscachedresolver.cpp
```diff
@@ -238,9 +238,18 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
     else if (ce->expires < now)
     {
       LOG_DEBUG("Expired entry found in cache");
-      do_query = true;
+      // Only query if we don't have another thread already doing this query
+      // for us
+      if (ce->pending_query)
+      {
+        LOG_DEBUG("Query already in progress on another thread");
```
bossmc: "Non-cached query..."?  You're already in a query... admittedly the terminology is confusing here as we're a DNS server and client at the same time.  Maybe "down-stream"?

bossmc:
Logging suggestions in-line.  The fix itself looks fine.

PR #260, opened 2015-03-12
[Reviewer: Andy] Update resource bundle version number

AlexHockey:
Because you can't update SAS events without a really boring cpp-common update...


bossmc:
Yawn

PR #265, opened 2015-03-22
[Reviewer: Alex] Fix HttpConnection segfault (and other potential ones)

rkd-msw:
Fixes https://github.com/Metaswitch/sprout/issues/621. See my comment in there for why.

I've grepped cpp-common, sprout, chronos, ralf and homestead for pthread_key_create calls - only cpp-common and sprout had any (sprout just had one in enumservice.cpp, which I'll fix up without review once this is OKed).

### DIFF
--- a/src/cassandra_store.cpp
+++ b/src/cassandra_store.cpp
@@ -261,6 +261,8 @@ Store::~Store()
   // (as the pool stores a pointer to the store). Make sure this is the case.
   stop();
   wait_stopped();
+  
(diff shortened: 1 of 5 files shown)


rkd-msw:
I've tested by running `build/bin/sprout_test --gtest_filter="HssC*" --gtest_repeat=2000` which now doesn't segfault.

AlexHockey:
Looks good. Nice job tracking this down!

PR #269, opened 2015-03-26
[Reviewer: Rob] Add in new get/delete options

eleanor-merry:
Rob, can you review these changes to support new uses of the send_get/send_delete methods

### DIFF
--- a/src/httpconnection.cpp
+++ b/src/httpconnection.cpp
@@ -288,6 +288,20 @@ HTTPCode HttpConnection::send_delete(const std::string& path,
 HTTPCode HttpConnection::send_delete(const std::string& path,
                                      SAS::TrailId trail,
                                      const std::string& body,
+                                     const std::string& override_server)
+{
+  std::string unused_response;
+  std::map<std::string, std::string> unused_headers;
+  _server = override_server;
+  _host = host_from_server(override_server);
+  _port = port_from_server(override_server);
+
+  return send_delete(path, unused_headers, unused_response, trail, body);
+}
+
+HTTPCode HttpConnection::send_delete(const std::string& path,
+                                     SAS::TrailId trail,
+                                     const std::string& body,
                                      std::string& response)
 {
   std::map<std::string, std::string> unused_headers;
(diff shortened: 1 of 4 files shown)


test_utils/fakecurl.cpp
```diff
@@ -156,27 +156,27 @@ CURLcode FakeCurl::easy_perform(FakeCurl* curl)
   {
     // Call the debug callback with some dummy HTTP messages (to exercise SAS
     // logging code).
-    char* text;
+    std::string text;
```
rkd-msw: Why have you made this change (just to cast it back to a char*)?

src/httpconnection.cpp
```diff
@@ -417,22 +461,26 @@ HTTPCode HttpConnection::send_get(const std::string& path,                     /
...
                                   const std::string& username,                 //< Username to assert (if assertUser was true, else ignored)
+                                  std::vector<std::string> headers_to_add,     //< Extra headers to add to the request
                                   SAS::TrailId trail)                          //< SAS trail
 {
   CURL *curl = get_curl_handle();
 
   curl_easy_setopt(curl, CURLOPT_HTTPGET, 1);
 
-  return send_request(path, "", response, username, trail, "GET", curl);
+  std::vector<std::string> unused_req_headers;
```
rkd-msw: Isn't this unnecessary?

src/httpconnection.cpp
```diff
@@ -288,6 +288,20 @@ HTTPCode HttpConnection::send_delete(const std::string& path,
 HTTPCode HttpConnection::send_delete(const std::string& path,
                                      SAS::TrailId trail,
                                      const std::string& body,
+                                     const std::string& override_server)
```
rkd-msw: What does override_server do?

test_utils/fakecurl.cpp
eleanor-merry: To stop Chronos whining

src/httpconnection.cpp
rkd-msw: (Presumably it changes the underlying server used by this connection - should that be a separate method that makes its side effect very clear? `change_server`?)

test_utils/fakecurl.cpp
```diff
@@ -156,27 +156,27 @@ CURLcode FakeCurl::easy_perform(FakeCurl* curl)
...
 
     text = "GET / HTTP/1.1\r\n\r\n";
     _debug_callback((CURL*)this,
                     CURLINFO_HEADER_OUT,
-                    text,
-                    strlen(text),
+                    (char*)text.c_str(),
+                    strlen(text.c_str()),
```
rkd-msw: `text.size()`?

test_utils/fakecurl.cpp
rkd-msw: You could change `debug_callback_t` to take a `const char*`, or `strdup()` the const char* and free it at the end of the function.

rkd-msw:
Looks fine. Some slightly perfectionist comments.

test_utils/fakecurl.cpp
bossmc: `text.length()`?

(jk - they're defined to be equal).

PR #270, opened 2015-03-27
[Reviewer: Andy] Set flags field to millisecond-accuracy rolling timestamp

mirw:
Andy,

Please can you review this change to use a millisecond-accuracy rolling timestamp for the flags field?

Thanks,

Matt


bossmc:
Looks good.

PR #275, opened 2015-04-08
[Reviewer: Alex] Add in different signal handlers

eleanor-merry:
Alex, these are the changes to make the signal handler passed to the updater configurable

### DIFF
--- a/include/memcachedstore.h
+++ b/include/memcachedstore.h
@@ -190,7 +190,7 @@ private:
                                                SAS::TrailId trail);
 
   // Stores a pointer to an updater object
-  Updater<void, BaseMemcachedStore>* _updater;
+  Updater<void, BaseMemcachedStore, SignalHandler<SIGHUP>>* _updater;
 
   // Used to store a connection structure for each worker thread.
   pthread_key_t _thread_local;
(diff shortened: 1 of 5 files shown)


AlexHockey:
Thanks for putting this PR together :-). 

It's a bit nasty that we have to expose a new template parameter to the `Updater` just to cope with the fact that the signal handler is itself templated. I think it would also be nice to have the `Updater` default to listening for `SIGHUP` rather than forcing each user to specify `_sighup_handler`

The only method that the updater calls on the signal handler is `wait_for_signal`. You could refactor this in to a purely virtual (non-templated) superclass (called, say, `SignalWaiter`), and have `SignalHandler` derive from that class and implement `wait_for_signal`. This means the `Updater` just has a pointer to `SignalWaiter` which is the same type no matter what signal handler is passed. 

Thoughts?

AlexHockey:
I've made the changes I suggested. @bossmc is going to review as part of the rest of my Astaire work.

bossmc:
I reviewed this as part of #276 right?

AlexHockey:
Ah, yes, I forgot that merging this branch into my sto1162 branch would make this change appear in #276.

PR #277, opened 2015-04-09
[Reviewer: Matt] Sto1170

eleanor-merry:
Matt, can you please review these changes to add the new Chronos alarm and update the stats script. 

Please ignore the changes to signalhandler and updater


src/alarmdefinition.cpp
```diff
@@ -331,6 +331,22 @@ namespace AlarmDef {
       }
     },
 
+    {CHRONOS_SCALE_IN_PROGRESS, DATABASE_INCONSISTENCY,
+      {
+        {CLEARED,
+          "Chronos: Chronos timers are synchronized",
+          "The local Chronos process is synchronized with the rest of the cluster. If a scale-up or "
+          "scale-down operation is in progress, it can now be completed."
```
mirw: Actually, you need to wait for both this and Astaire to complete.  We should probably update this and the Astaire alarm to make this clear.

scripts/stats-c/cw_stat.cpp
```diff
@@ -276,18 +276,20 @@ int main(int argc, char** argv)
...
-        (msgs[0] == "P_rejected_overload_0"))
+    if ((msgs[0] == "client_count")                 ||
+        (msgs[0] == "incoming_requests")            ||
+        (msgs[0] == "rejected_overload")            ||
+        (msgs[0] == "H_incoming_requests")          ||
+        (msgs[0] == "H_rejected_overload")          ||
+        (msgs[0] == "P_incoming_requests_0")        ||
+        (msgs[0] == "P_rejected_overload_0")        ||
+        (msgs[0] == "chronos_scale_nodes_to_query") ||
+        (msgs[0] == "chronos_scale_timers_processed"))
```
mirw: I know we've historically exposed these statistics at a very fine granularity, but I think this might have been a mistake.  Should we be combining these?  (Probably not now but I think we're going to need rework statistics generally in future.)

mirw:
Looks good.  A couple of comments.

src/alarmdefinition.cpp
rkd-msw: Although the text should be slightly flexible, to accommodate node types which might have Chronos but not memcached (or vice-versa).

PR #278, opened 2015-04-09
[Reviewer: Rob] Iss273

### LINKED ISSUE #273: Whitespace in cluster_settings breaks memcached replication
Memcached tries to connect to " address" which doesn't exist. Fix is to strip whitespace from file

eleanor-merry:
Rob, can you please review this fix to strip whitespace from the memcached config file. 

Fixes #273 

Tested live

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -128,7 +128,9 @@ namespace Utils
   {
     return ltrim(rtrim(s));
   }
-  
+
+  // Strip all whitespace from the string (using the same pattern as the trim
+  // functions above - note that this modifies the passed in string)
   inline std::string& strip_whitespace(std::string& s)
   {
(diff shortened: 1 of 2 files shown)


rkd-msw:
Looks good. How easy would it be to unit-test this? If it's not difficult, maybe we should do it.

include/utils.h
```diff
@@ -124,10 +124,16 @@ namespace Utils
   }
 
   // trim from both ends
-  inline std::string& trim(std::string &s)
+  inline std::string& trim(std::string& s)
   {
     return ltrim(rtrim(s));
   }
+  
+  inline std::string& strip_whitespace(std::string& s)
```
rkd-msw: Should this modify its argument, rather than returning a new string? That feels like it could potentially bite us in future.

PR #279, opened 2015-04-10
[Reviewer: Ellie] Fix log to use correct variables

rkd-msw:
Spotted some logs like:

`09-04-2015 16:54:05.748 UTC Debug zmq_lvc.cpp:285: Replaying cache for entry 0x1ded608 (length: 31341008)`

I realised that the very high length was because we were passing a pointer rather than the vector size. This fixes that.


eleanor-merry:
Looks good

PR #281, opened 2015-04-20
[Reviewer: Alex] Enforce a 2-second gap between every change to the token bucket rate

rkd-msw:
On the hangout call a while ago, someone (you? Matt?) said it might be a good idea to modify the load monitor to only change the token bucket rate once every N seconds. This changes it to only change the token bucket rate after it's seen at least 20 requests and at least 2 seconds have passed.

I've upgraded one of the logs from LOG_DEBUG to LOG_STATUS - this is something that @steelyvoid  was interested in, and it should be safe (the log is 200 bytes at most - even if the rate changes every 2 seconds, that's only 352KB of extra logs an hour, and I'd be very surprised if latency fluctuates so wildly that the rate does change every two seconds).

I've used CLOCK_MONOTONIC_COARSE as a premature optimisation, to help make sure that checking the time more frequently doesn't have a performance impact.

I'll live test by running the live tests (which should generate >20 requests over >2 seconds) and checking the logs to make sure they make sense.


AlexHockey:
The code looks good.

PR #283, opened 2015-04-22
[Reviewer: Matt] Add stats

eleanor-merry:
Matt, can you review these changes to add the new diameter routing stats to cpp-common

### DIFF
--- a/src/diameterstack.cpp
+++ b/src/diameterstack.cpp
@@ -232,12 +232,12 @@ void Stack::fd_error_hook_cb(enum fd_hook_type type,
 
   // Increment routing error stats if they're supported
   if ((_realm_counter != NULL) && 
-      (strcmp((char*)other, "Message for another realm") == 0))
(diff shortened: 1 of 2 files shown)


src/diameterstack.cpp
```diff
@@ -221,19 +223,25 @@ void Stack::fd_error_hook_cb(enum fd_hook_type type,
...
+  LOG_INFO("Routing error: '%s' for message with "
+           "Command-Code %d, Destination-Host %s and Destination-Realm %s",
+           (char *)other,
+           msg2.command_code(),
+           dest_host.c_str(),
+           dest_realm.c_str());
 
+  // Increment routing error stats if they're supported
+  if ((_realm_counter != NULL) && 
+      (strcmp((char*)other, "Message for another realm") == 0))
```
mirw: This (comparing strings) feels pretty fragile.  What can we do to make it better?

mirw:
One comment.  Otherwise fine.

PR #284, opened 2015-04-24
[Reviewer: Alex] Fixes to the load monitor for UTability

rkd-msw:
Follow-on to https://github.com/Metaswitch/cpp-common/pull/281.

Trivial changes - allowed CLOCK_MONOTONIC_COARSE to be controlled by the UT time controls, clarified some logs.

More interesting change - smoothed_latency started at 0. This meant that if your first 20 requests are all dead-on your target latency of 100ms, the smoothed latency is actually about 92ms, and your bucket rate increases. I think that's wrong, so I've defaulted it to target_latency.


AlexHockey:
Looks good. Your change to the latency initial conditions sounds sensible as well.

PR #287, opened 2015-04-30
[Reviewer: Andy] Make DNS blacklist duration configurable

mirw:
Andy,

Please can you review this fix to make the DNS blacklist duration configurable?  There's no UT for this, but I've tested it live (and it's actually just refactoring).

Matt

### DIFF
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -92,6 +92,7 @@ public:
   BaseResolver(DnsCachedResolver* dns_client);
   virtual ~BaseResolver();
 
+  void blacklist(const AddrInfo& ai) { blacklist(ai, _default_blacklist_duration); }
   void blacklist(const AddrInfo& ai, int ttl);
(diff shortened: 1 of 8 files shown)


bossmc:
Code looks fine, does make me wonder why the blacklist is implemented in both HTTPResolver and SIPResolver rather than being generically in BaseResolver...

mirw:
Sorry, I'm not sure I understand.  Can you elaborate?  Are you referring to the fact that BaseResolver requires create_blacklist to be called rather than creating the blacklist implicitly in the constructor?

bossmc:
Ah, I got confused because `create_blacklist` and `destroy_blacklist` do nothing of the sort in `BaseResolver` which I assumed meant that the actual implementation was in `SIPResolver...`.  Turns out they're just odd.

mirw:
Thanks - that makes sense.

PR #290, opened 2015-05-01
[Reviewer: Ellie] sto1264 changes

AlexHockey:
This PR contains two main changes:

* Add SAS correlation to the diameter stack based on diameter session ID. 
* Split out the datestamp from the resource bundle name (to allow us to define other resource bundles in future.

### DIFF
--- a/src/diameterstack.cpp
+++ b/src/diameterstack.cpp
@@ -828,9 +828,9 @@ void Stack::fd_sas_log_diameter_message(enum fd_hook_type type,
 
   // Now construct a correlating marker based on the diameter session ID.
   struct session* sess;
-  int is_new;
+  int dummy_is_new;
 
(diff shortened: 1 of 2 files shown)


src/diameterstack.cpp
```diff
@@ -825,6 +825,29 @@ void Stack::fd_sas_log_diameter_message(enum fd_hook_type type,
   event.add_compressed_param(data->length, data->buffer);
 
   SAS::report_event(event);
+
+  // Now construct a correlating marker based on the diameter session ID.
+  struct session* sess;
+  int is_new;
```
eleanor-merry: What's is_new do? Can this have a different name to make it clear we're not using it?

eleanor-merry:
Looks good, one comment

PR #289, opened 2015-05-01
[Reviewer: Rob] Add in etcd chronos alarms

eleanor-merry:
Rob, can you review the new alarm text for chronos etcd clustering

PR #291, opened 2015-05-03
[Reviewer: Ellie] Add new memcached/etcd alarms

rkd-msw:
It probably makes sense for the MEMCACHED_NOT_YET_CLUSTERED text and the CHRONOS_NOT_YET_CLUSTERED text to be consistent, so I'll update this if you apply markups to https://github.com/Metaswitch/cpp-common/pull/289.


src/alarmdefinition.cpp
```diff
@@ -576,6 +594,24 @@ namespace AlarmDef {
         }
       }
     },
+
+    {ETCD_CLUSTERING_TAKING_TOO_LONG, DATABASE_INCONSISTENCY,
+      {
+        {CLEARED,
+          "etcd:Scaling process has finished",
```
eleanor-merry: Space between etcd and Scaling

src/alarmdefinition.cpp
```diff
@@ -378,6 +378,24 @@ namespace AlarmDef {
       }
     },
 
+    {MEMCACHED_NOT_YET_CLUSTERED, DATABASE_INCONSISTENCY,
+      {
+        {CLEARED,
+          "Memcached: The local Memcached is part of the Chronos cluster",
+          "The local Memcached process is synchronized with the rest of the cluster. Changes made by"
+          " this node (e.g. to subscriber registrations) will be correctly replicated."
```
eleanor-merry: Not so happy with the spacing here

eleanor-merry:
Looks good, some tedious spacing comments

src/alarmdefinition.cpp
graemerobertson: If I had to make one change to this code I'd change "Memcached: The local Memcached is part of the **Chronos** cluster" rather than this spacing issue

PR #295, opened 2015-05-13
[Reviewer: Ellie] Don't crash if we have an empty /etc/clearwater/remote_cluster_settings

rkd-msw:
Other part of https://github.com/Metaswitch/sprout/pull/1035.

### DIFF
--- a/src/memcached_config.cpp
+++ b/src/memcached_config.cpp
@@ -84,6 +84,12 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
                             0, 
                             true);
 
+        if ((tokens[0] == "servers") && tokens.size() == 1)
+        {
+          LOG_DEBUG("Accepting empty servers= line");
+          return true;
+        }
+
         if (tokens.size() != 2)
         {
           LOG_ERROR("Malformed config file (got bad line: '%s')",
--- a/src/memcachedstoreview.cpp
+++ b/src/memcachedstoreview.cpp
@@ -261,12 +261,14 @@ std::string MemcachedStoreView::view_to_string()
 std::string MemcachedStoreView::replicas_to_string(const std::vector<std::string>& replicas)
 {
   std::string s;
-  for (size_t ii = 0; ii < replicas.size()-1; ++ii)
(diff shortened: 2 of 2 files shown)


src/memcached_config.cpp
```diff
@@ -84,6 +84,12 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
                             0, 
                             true);
 
+        if ((tokens[0] == "servers") && tokens.size() == 1)
```
eleanor-merry: This will accept "servers=" in /etc/clearwater/cluster_settings as well won't it?

src/memcachedstoreview.cpp
```diff
@@ -261,12 +261,14 @@ std::string MemcachedStoreView::view_to_string()
...
 {
   std::string s;
-  for (size_t ii = 0; ii < replicas.size()-1; ++ii)
+  if (!replicas.empty())
   {
-    s += replicas[ii] + "/";
+    for (size_t ii = 0; ii < replicas.size()-1; ++ii)
+    {
+      s += replicas[ii] + "/";
+    }
```
eleanor-merry: Can you test this at all in clearwater-fv-test?

eleanor-merry:
Looks good, but I'd like the new function to only apply to the remote_cluster_settings file

src/memcached_config.cpp
rkd-msw: Yes, but there's code in sprout's main.cpp to error out if the main store has no servers, but (after my change above) allow it for the remote store. I think that's probably the right approach - this config file parser allows "servers=" as a valid file, but the applications can treat that as bad confguration for the primary store.

src/memcachedstoreview.cpp
rkd-msw: This particular code just logs out what servers map to what buckets (at start=-of-day or on reload) - so I don't think I can, and I'm not sure it's worth it (we don't usually test logs).

I think the only clearwater-fv-test test that makes sense here is "does a store configured with `servers=` do nothing?", which doesn't feel like a hugely valuable test.

PR #296, opened 2015-05-15
Reviewer AMC: Don't pass localhost straight through to evhtp_bind_socket

graemerobertson:
Tested live with default bind address (localhost, resolving to ipv6 address), bind address set to localhost (ipv6), bind address set to localhost (ipv4), bind address set to the local ip (ipv4), bind address set to 127.0.0.1 and bind address set to ::1. Anything else worth testing?

Fixes https://github.com/Metaswitch/chronos/issues/128.


bossmc:
Get excited!

PR #298, opened 2015-05-20
[Reviewer: Ellie] Memcached fixes

rkd-msw:
This suppresses one instance of log spam we discussed (128 lines of empty memcached view) and fixes up cluster_settings parsing so that:

```
servers=
new_servers=10.0.0.1:11211
```

is treated properly. I've tested live.

### DIFF
--- a/src/memcached_config.cpp
+++ b/src/memcached_config.cpp
@@ -57,6 +57,7 @@ MemcachedConfigFileReader::~MemcachedConfigFileReader() {}
 
 bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
 {
+  bool seen_servers = false;
   config.servers.clear();
   config.new_servers.clear();
   config.tombstone_lifetime = DEFAULT_TOMBSTONE_LIFETIME;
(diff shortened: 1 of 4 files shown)


src/memcached_config.cpp
```diff
@@ -84,45 +84,57 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
...
 
-          if (std::to_string(config.tombstone_lifetime) != tokens[1])
+          if (std::to_string(config.tombstone_lifetime) != value)
           {
             LOG_ERROR("Config contained an invalid tombstone_lifetime line:\n%s",
                       line.c_str());
             return false;
           }
         }
+         else
```
eleanor-merry: Indentation

src/memcached_config.cpp
```diff
@@ -132,11 +144,5 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
     return false;
   }
 
-  if (config.servers.size() == 0)
```
eleanor-merry: Should this check that at least one of servers/new_servers has entries instead?

rkd-msw: No, because `servers=` is a valid remote_cluster_settings file. The application can do that checking by calling MemcachedStore::has_servers().

eleanor-merry:
Looks good

Issue #299, opened 2015-05-21
diameterstack generates rogue error log when it's being used client side


graemerobertson:
When the diameter stack is being used on the client side, it often generates the following error log:

```
21-05-2015 10:36:57.312 UTC Error diameterstack.cpp:321: Unexpected host on callback (type 12) from freeDiameter: gdr-2.obuenf.mvoip.aaa.datcon.co.uk`
```

This is because the diameterstack has registered for a hook callback on a peer connection success (or failure). This is required by the realmmanager. When the diameterstack is being used as a client we don't use the realmmanager and we're not interested in this hook. The realmmanager should be responsible for registering for this hook.

I've got a fix for this, but it breaks the UTs at the moment.

graemerobertson:
The fix for this is on cpp-common branch iss299. It currently breaks the RealmmanagerTest.CreateDestroy homestead UT. I think the fix is probably just to mock out diameter stack's new register_peer_hook_hdlr method in mockdiameterstack.hpp.

PR #300, opened 2015-05-21
Reviewer AMC: RealmManager should register peer hooks rather than the Diameter stack

graemerobertson:
Hi Andy,

Fixes https://github.com/Metaswitch/cpp-common/issues/299.

Only register for the peer connected/failed hook when we've actually got a realm manager.

I haven't tested this, it doesn't seem that controversial to me and I don't have time to test it at the moment - what do you think?

Thanks,
Graeme


src/diameterstack.cpp
```diff
@@ -121,13 +115,27 @@ void Stack::initialize()
...
 }
 
+void Stack::register_peer_hook_hdlr()
+{
+  int rc = fd_hook_register(HOOK_MASK(HOOK_PEER_CONNECT_SUCCESS,
+                                      HOOK_PEER_CONNECT_FAILED),
+                            fd_peer_hook_cb,
+                            this,
+                            NULL,
+                            &_peer_cb_hdlr);
```
bossmc: Shouldn't the callback be provided by the `RealmManager` too?

bossmc: I know this gets cleaned up in `~Diameter::Stack()` but it should probably also be possible for the `RealmManager` to do it's own unregister for symmetry.

graemerobertson: I can't really envisage what you're suggesting, and I doubt you're going to convince me to implement it - it sounds like a lot of work for virtually no gain. Do you think it's required for this PR?

I'm happy to implement your second comment.

bossmc:
Ok fine.

PR #302, opened 2015-06-02
[Reviewer: Alex] Support comments

eleanor-merry:
This supports comments in the memcached configuration files (part of the fix for https://github.com/Metaswitch/clearwater-etcd/issues/34)

### DIFF
--- a/src/memcached_config.cpp
+++ b/src/memcached_config.cpp
@@ -76,7 +76,7 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
 
       LOG_DEBUG("Got line: %s", line.c_str());
 
-      if (line.length() > 0 && line[0] != '#')
+      if ((line.length() > 0) && (line[0] != '#'))
       {
         // Read a non-blank line.
         std::vector<std::string> tokens;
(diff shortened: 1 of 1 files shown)


src/memcached_config.cpp
```diff
@@ -72,14 +72,15 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
     {
       std::string line;
       getline(f, line);
+      line = Utils::strip_whitespace(line);
```
AlexHockey: An `ltrim` or `trim` be better here - no need to strip all internal space from the line as well

src/memcached_config.cpp
```diff
@@ -72,14 +72,15 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
...
 
       LOG_DEBUG("Got line: %s", line.c_str());
 
-      if (line.length() > 0)
+      if (line.length() > 0 && line[0] != '#')
```
AlexHockey: Would normally have brackets around sub-expressions

AlexHockey:
Reviewed - a couple of markups.

src/memcached_config.cpp
eleanor-merry: We need to strip all internal whitespace - see #273

AlexHockey: Ah OK.

PR #305, opened 2015-06-10
[Reviewer: Andy] Supress 100 Continue responses

### LINKED ISSUE #304: 100 Continue messages from HTTP stack confuse SAS
Our HTTP stack (based on libevhtp) supports sending 100 Continue messages. Curl always adds a `Expect: 100-continue` header`, which means that (for example) if Sprout contacts Homestead the flow is: 

* Request (sprout -> homestead)
* 100 Continue (homestead -> sprout)
* Final response (homestead -> sprout)

Our SAS logging on the client (sprout) sees the 100 Continue but the logging on the server does not. This means the response flows looks different on the two ends, which throws SAS message matching for a loop.

AlexHockey:
Andy, can you review this PR which stops cURL from asking for 100 Continue responses in some situations (which fixes #304). 

Tested live. Here is a before shot:

image

And here is an after:

image

### DIFF
--- a/src/httpconnection.cpp
+++ b/src/httpconnection.cpp
@@ -442,11 +442,11 @@ HTTPCode HttpConnection::send_get(const std::string& path,
 }
 
 /// Get data; return a HTTP return code
-HTTPCode HttpConnection::send_get(const std::string& path,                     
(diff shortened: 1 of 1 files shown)


bossmc:
Looks fine.

PR #308, opened 2015-06-11
[Reviewer: Matt] Don't count ping handler responses

### LINKED ISSUE #306: Ping handler responses are included in Homestead's average latency
These are super quick, so they suppress the average latency for Homestead. This means that the load monitor doesn't throttle Homestead as quickly as it should, and it can mean that Homestead starts with a stupidly high token rate if the first couple of calculations are done only on the ping handler times.

eleanor-merry:
This stops us from including any response from the Ping Handler in our latencies
Fixes #306

### DIFF
--- a/include/httpstack.h
+++ b/include/httpstack.h
@@ -148,7 +148,7 @@ public:
       }
     }
 
-    void send_reply(int rc, SAS::TrailId trail);
+    void send_reply(int rc, SAS::TrailId trail, bool use_latency=true);
     inline evhtp_request_t* req() { return _req; }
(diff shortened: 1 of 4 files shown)


mirw:
As discussed, adding a `use_latency` flag to a `send_reply` method (which says nothing else about latency) feels a bit like we're overloading it.  You're going to have another think.

eleanor-merry:
The request itself now owns whether it should track the latency - can you please re-review

mirw:
Looks good.

PR #311, opened 2015-06-19
[Reviewer: Matt] SNMP tooling

rkd-msw:
Matt,

Can you review my code to add SNMP helpers so that Sprout can become an AgentX subagent?

There's
- snmp_agent.{h, cpp} which sets up the AgentX thread
- snmp_scalar.h, snmp_table.h and snmp_table.cpp which set up the general infrastructure
- snmp_time_period_table.h, snmp_counter_table.{h, cpp}, snmp_accumulator_table.{h, cpp} and snmp_ip_count_table.h which define specific table structures
- a change to HTTPConnection so that it can be created with a SNMP::IPCountTable* instead of the current 0MQ classes


include/httpconnection.h
```diff
@@ -184,12 +193,16 @@ class HttpConnection
     ~PoolEntry();
 
     void set_remote_ip(const std::string& value);
+    void update_zmq_ip_counts(const std::string& value);
```
mirw: Why are we still referring to zmq?

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+
+#ifndef SNMP_ACCUMULATOR_TABLE_H
+#define SNMP_ACCUMULATOR_TABLE_H
+
+// This file contains infrastructure for tables which:
+//   - are indexed by time period
+//   - accumulate data samples over time
+//   - report a count of samples, mean sample value, variance, high-water-mark and low-water-mark
+//
+//   The thing sampled doesn't matter - it could be latency, size of a queue, etc.
```
mirw: Indentation?

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+
+namespace SNMP
+{
+
+// Storage for the underlying data
+struct Statistics
```
mirw: A generic name for a not-very-generic set of statistics?

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
+/**
+ * @file snmp_accumulator_table.h
```
mirw: This file seems to define everything related to an "accumulator", not just the table.  Should we rename?

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+                   int oidlen) :
+    ManagedTable<AccumulatorRow, int>(name, tbl_oid, oidlen),
+    five_second(5),
+    five_minute(300)
+  {
+    _tbl.add_index(ASN_INTEGER);
+    _tbl.set_visible_columns(2, 6);
+
+    // We have a fixed number of rows, so create them in the constructor.
+    add_row(0);
```
mirw: Significance of 0, 1, 2?  Should we have an enum?

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+public:
+  AccumulatorTable(std::string name,
+                   oid* tbl_oid,
+                   int oidlen) :
+    ManagedTable<AccumulatorRow, int>(name, tbl_oid, oidlen),
+    five_second(5),
+    five_minute(300)
+  {
+    _tbl.add_index(ASN_INTEGER);
+    _tbl.set_visible_columns(2, 6);
```
mirw: Magic numbers everywhere!

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+
+    // We have a fixed number of rows, so create them in the constructor.
+    add_row(0);
+    add_row(1);
+    add_row(2);
+  }
+
+  AccumulatorRow* new_row(int index)
```
mirw: Presumably there's a comment somewhere about how add_row and new_row interact?

mirw: Does this method need to be public, or should it be protected?

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+
+  // Accumulate a sample into the underlying statistics.
+  void accumulate(uint32_t sample)
+  {
+    // Pass samples through to the underlying data structures
+    accumulate_internal(five_second, sample);
+    accumulate_internal(five_minute, sample);
+  }
+
+  void accumulate_internal(AccumulatorRow::CurrentAndPrevious& data, uint32_t sample);
```
mirw: This should presumably be private.

include/snmp_accumulator_table.h
```diff
@@ -0,0 +1,136 @@
...
+      case 2:
+        // Five-minute row
+        view = new AccumulatorRow::PreviousView(&five_minute);
+        break;
+    }
+    return new AccumulatorRow(index, view);
+  }
+
+  // Accumulate a sample into the underlying statistics.
+  void accumulate(uint32_t sample)
```
mirw: How do I know how to use this class?  After reviewing the code, I think I construct an instance and then just call accumulate.  Is that correct?  ...and where is this documented?

include/snmp_agent.h
```diff
@@ -0,0 +1,49 @@
...
+ * "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
+ * Project and licensed under the OpenSSL Licenses, or a work based on such
+ * software and licensed under the OpenSSL Licenses.
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+///
+///
```
mirw: ?

include/snmp_agent.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef CW_SNMP_AGENT_H
+#define CW_SNMP_AGENT_H
+
+// Starts the SNMP agent thread
+int snmp_setup(const char* name);
```
mirw: What's the significance of `name`?

include/snmp_counter_table.h
```diff
@@ -0,0 +1,128 @@
...
+    _tbl.set_visible_columns(2, 3);
+
+    // Fixed number of rows, so set them up now.
+    add_row(0);
+    add_row(1);
+    add_row(2);
+  }
+  
+  // Map row indexes to the view of the underlying data they should expose
+  CounterRow* new_row(int index)
```
mirw: Protected?

include/snmp_counter_table.h
```diff
@@ -0,0 +1,128 @@
...
+  void increment()
+  {
+    // Increment each underlying set of data.
+    five_second.update_time();
+    five_second.current->count++;
+    five_minute.update_time();
+    five_minute.current->count++;
+  }
+
+  CounterRow::CurrentAndPrevious five_second;
```
mirw: Private?

include/snmp_ip_count_table.h
```diff
@@ -0,0 +1,131 @@
...
+  {
+    // Convert the string into a type (IPv4 or IPv6) and a sequence of bytes
+    if (inet_pton(AF_INET, addrStr.c_str(), &_addr) == 1) {
+      _addr_type = 1; // IPv4
+      _addr_len = sizeof(struct in_addr);
+    } else if (inet_pton(AF_INET6, addrStr.c_str(), &_addr) == 1) {
+      _addr_type = 2; // IPv6
+      _addr_len = sizeof(struct in6_addr);
+    } else {
+      _addr_type = 0; // unknown
```
mirw: Is it right to construct a 0/0 OID in this case?  Is there not a chance we'll end up having two rows with the same index?  What happens then?  I think we should at least report an error.

include/snmp_ip_count_table.h
```diff
@@ -0,0 +1,131 @@
...
+class IPCountRow : public Row
+{
+public:
+  IPCountRow(std::string addrStr) :
+    Row(),
+    _count(0)
+  {
+    // Convert the string into a type (IPv4 or IPv6) and a sequence of bytes
+    if (inet_pton(AF_INET, addrStr.c_str(), &_addr) == 1) {
+      _addr_type = 1; // IPv4
```
mirw: Presumably defined by RFC 2851?  Worth a comment?

include/snmp_scalar.h
```diff
@@ -0,0 +1,78 @@
...
+    netsnmp_register_read_only_ulong_instance(name.c_str(),
+                                              _oid,
+                                              _oidlen,
+                                              &value,
+                                              NULL);
+  }
+
+  ~U32Scalar()
+  {
+    unregister_mib(_oid, _oidlen);
```
mirw: Where's this defined?

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+
+#include "snmp_includes.h"
+#include "log.h"
+
+#ifndef SNMP_TABLE_H
+#define SNMP_TABLE_H
+
+// netsnmp handler function (of type Netsnmp_Node_Handler). Called for each SNMP request on a table,
+// and maps the row and column to a value.
+int netsnmp_table_handler_fn(netsnmp_mib_handler *handler,
```
mirw: This is declared as a public function but is only used privately within this library.  Shouldn't we hide it away (e.g. as a "static" function or in a namespace)?

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  virtual ~Row()
+  {
+    netsnmp_tdata_delete_row(_row);
+  }
+
+  virtual ColumnData get_columns() = 0;
+
+  netsnmp_tdata_row* get_netsnmp_row() { return _row; };
```
mirw: Does this need to be public?  Can't it be friend Table?

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+protected:
+  netsnmp_tdata_row* _row;
+};
+
+// Generic SNMPTable class wrapping a netsnmp_tdata and netsnmp_table_registration_info and exposes
```
mirw: Typo: "wrapping ... and exposes"

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+// an API for manipulating them easily. Doesn't need subclassing, but should usually be wrapped in a
+// ManagedTable subclass for convenience.
+template<class T> class Table
+{
+public:
+  Table(std::string name,
+        oid* tbl_oid,
+        int oidlen):
+    _name(name),
```
mirw: What's the significant of the name?

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+      netsnmp_unregister_handler(_handler_reg);
+    }
+    snmp_free_varbind(_table->indexes_template);
+    snmp_free_varbind(_table_info->indexes);
+    netsnmp_tdata_delete_table(_table);
+    SNMP_FREE(_table_info);
+  }
+
+  // Helper method to set a column index on both underlying objects.
+  void add_index(int type)
```
mirw: Unused variable.  Do you see compilation warnings?

mirw: ...and the variable should be used below.

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+  {
+    netsnmp_tdata_add_index(_table, ASN_INTEGER);
+    netsnmp_table_helper_add_index(_table_info, ASN_INTEGER);
+  }
+
+  // Helper method to change the visibility of columns in this table.
+  void set_visible_columns(int min, int max)
```
mirw: Why aren't these constructor parameters?  Should they ever change?

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+    _table_info->min_column = min;
+    _table_info->max_column = max;
+  }
+
+  // Add a Row into the underlying table.
+  void add(T* row)
+  {
+    if (_handler_reg == NULL)
+    {
+      register_tbl();
```
mirw: Why don't we register the table until we create the first row?  Shouldn't we do this on construction?  This is at least worth a comment.

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+  };
+
+  // Remove a Row from the underlying table.
+  void remove(T* row)
+  {
+    netsnmp_tdata_remove_row(_table, row->get_netsnmp_row());
+  };
+
+protected:
+  // Registers an SNMP handler for this table. Called when the first row is added.
```
mirw: True, but *why*?

include/snmp_table.h
mirw: Or should we pass the list of indices into the table on construction?

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+
+  std::string _name;
+  oid* _tbl_oid;
+  int _oidlen;
+  netsnmp_handler_registration* _handler_reg;
+  netsnmp_table_registration_info* _table_info;
+  netsnmp_tdata* _table;
+};
+
+// Base ManagedTable, wrapping an SNMPTable and managing the ownership of rows.
```
mirw: Why does this wrap SNMPTable rather than subclassing it?

mirw: The class is SNMP::Table, not SNMPTable.

include/snmp_table.h
```diff
@@ -0,0 +1,301 @@
...
+    if (_map.find(key) == _map.end())
+    {
+      add_row(key);
+    }
+
+    return _map.at(key);
+  }
+
+  // Deletes the row keyed off `key`.
+  void remove(TRowKey key)
```
mirw: Why is it `add_row` but `remove` (rather than `remove_row`)?

include/snmp_time_period_table.h
```diff
@@ -0,0 +1,171 @@
...
+#include "snmp_includes.h"
+#include "logger.h"
+
+#ifndef SNMP_TIME_PERIOD_TABLE_H
+#define SNMP_TIME_PERIOD_TABLE_H
+
+namespace SNMP
+{
+template <class T> class TimeBasedRow : public Row
+  {
```
mirw: Indentation.

include/snmp_time_period_table.h
```diff
@@ -0,0 +1,171 @@
...
+  public:
+
+    // For the data type T over the period _interval, store a current and previous accumulation of
```
mirw: "Accumulation"?  Isn't this used for things other than the accumulator?

include/snmp_time_period_table.h
```diff
@@ -0,0 +1,171 @@
...
+      {
+        struct timespec now;
+        clock_gettime(CLOCK_MONOTONIC_COARSE, &now);
+
+        // Count of how many _interval periods have passed since the epoch
+        uint32_t new_tick = (now.tv_sec / _interval);
+
+        // Count of how many _interval periods have passed since the last change
+        uint32_t tick_difference = new_tick - _tick;
+        _tick = new_tick;
```
mirw: We have no locking or atomics here.  This means that we're liable to get corruption even during normal operation (not just leaking of statistics between periods).

include/snmp_time_period_table.h
```diff
@@ -0,0 +1,171 @@
...
+
+        if (tick_difference == 1)
+        {
+          T* tmp;
+          tmp = previous;
```
mirw: This might even result in segmentation faults (due to reading an invalid pointer).  (Unlikely, because the compiler will probably 8-byte align the pointer, but I think we should avoid it.)

src/httpconnection.cpp
```diff
@@ -841,6 +884,39 @@ void HttpConnection::PoolEntry::set_remote_ip(const std::string& value)  //< Rem
...
+  }
+
+  _remote_ip = value;
+}
+
+void HttpConnection::PoolEntry::update_snmp_ip_counts(const std::string& value)  //< Remote IP, or "" if no connection.
+{
+  if (!_remote_ip.empty())
+  {
+      if (_parent->_stat_table->get(remote_ip)->decrement() == 0)
```
mirw: Indentation.

src/httpconnection.cpp
```diff
@@ -841,6 +884,39 @@ void HttpConnection::PoolEntry::set_remote_ip(const std::string& value)  //< Rem
...
+      {
+        _parent->_stat_table->remove(remote_ip);
+      }
+  }
+
+  if (!value.empty())
+  {
+      _stat_table->get(remote_ip)->increment();
```
mirw: Ditto.

src/snmp_accumulator_table.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+ColumnData AccumulatorRow::get_columns()
+{
+  Statistics accumulated = *(_view->get_data());
+  uint32_t sum = accumulated.sum;
+  uint32_t sumsq = accumulated.sqsum;
+  uint32_t count = accumulated.count;
+
+  // Calculate the average and the variance from the stored sum and sum-of-squares.
+  uint32_t avg = sum/std::max(count, 1u);
```
mirw: Why std::max?  (I know, but it's worth a comment.)

src/snmp_accumulator_table.cpp
```diff
@@ -0,0 +1,87 @@
...
+  data.current->sum += latency;
+  data.current->sqsum += (latency * latency);
+
+  // Change the high-water-mark and low-water-mark if this data is the highest/lowest we've seen.
+  if (latency > data.current->hwm)
+  {
+    data.current->hwm = latency;
+  }
+
+  if ((latency < data.current->lwm) || (data.current->lwm == 0))
```
mirw: Is this right?  It means 0 will be ignored as a possible least value.

src/snmp_agent.cpp
```diff
@@ -0,0 +1,86 @@
...
+  netsnmp_container_init_list();
+  int rc = init_agent(name);
+  if (rc != 0)
+  {
+    TRC_WARNING("SNMP AgentX initialization failed");
+    return -1;
+  }
+  else
+  {
+    TRCSTATUS("AgentX agent initialised");
```
mirw: Why is it `TRC_WARNING` with an underscore in but `TRCSTATUS` without?

src/snmp_agent.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "snmp_agent.h"
+#include "snmp_latency_table.h"
+#include "log.h"
+
+static pthread_t snmp_thread_var;
+
+void* snmp_thread(void* data)
+{
+  while (1) {
```
mirw: Use hadouken braces.

mirw:
Looks good.  A few (mainly commenting/encapsulation) comments inline.  Only significant one is around atomicity/locking where I think we're much less robust than we used to be.

include/snmp_accumulator_table.h
rkd-msw: I think this name is OK - I don't think we expose any other statistics (everything is just the mean/variance/HWM/LWM). If in future we add other statistics that don't follow this pattern, we should rename it then (and we'll have a better idea of what distinguishes the statistics and be able to pick better names).

include/snmp_accumulator_table.h
rkd-msw: I think the only reason you'd include this is to use the table - the other stuff is just there to support that, so I think something like `snmp_accumulator_utils.h` would be more technically accurate but less clear.

include/snmp_accumulator_table.h
rkd-msw: That's correct - I'll add a file comment to document that.

include/snmp_scalar.h
rkd-msw: It's a netsnmp API call - I'll comment to that effect because it doesn't have the netsnmp prefix.

include/snmp_ip_count_table.h
rkd-msw: I'm going to move this logic out of the constructor into `IPCountTable::new_row` - if the address doesn't parse I'll report an error and return NULL.

src/snmp_accumulator_table.cpp
rkd-msw: Yes, I think this is the best solution:
 - 0 microseconds is a pretty unlikely latency, and the next smallest value will also be pretty small - reporting 3 us instead of 0 us doesn't feel like a big deal
 - the alternative is setting this to UINT_MAX or something, and just reporting 0 if the count is 0, but I'd be nervous about that - what if we get a torn read and a count of 1 but a LWM of UINT_MAX? That would screw up any graphs much more than reporting 0 us.

mirw: On what basis is 0 microseconds unlikely?  ...and why is the next smallest value also going to be pretty small?

The alternative (of setting UINT_MAX) is exactly what we used to do.  I agree this would be wrong, but you're right it would be noticeably wrong - much better to be able to detect this than not to!  If you're worried about the torn-read case, why not fix it up when we push these values into the SNMP row?

rkd-msw:
All review markups addressed; merging.

PR #314, opened 2015-06-26
[Reviewer: Matt] Fix variance, HWM and LWM

rkd-msw:
In my new SNMP stats code, I miscalculated variance, and put the HWM and LWM in the opposite columns. This is a trivial fix to fix that.

Tests in a separate PR.


mirw:
Looks good.

PR #319, opened 2015-07-02
[Reviewer: Matt] Hide netsnmp behind interfaces

rkd-msw:
I've tested in clearwater-fv-test, by forcing a compile-time error if READ is defined after including the client-code-facing header files that should be interface-only.


include/snmp_table.h
```diff
@@ -38,6 +38,7 @@
 #include <map>
 #include <string>
 
+#include "snmp_row.h"
 #include "snmp_includes.h"
```
mirw: Worth a comment in the header that this is not intended to be included by core code and why?

rkd-msw: I wonder if it's actually worth shoving all these headers into a `include/snmp_internals` subfolder.

mirw: Yes, that sounds good.  Although it's still worth commenting why.

mirw:
Looks good.  I eye-balled the code moves, but didn't check them line-by-line - are you confident about those moves?  (I guess so, but wanted to check.)

rkd-msw:
Yes - I've confirmed that the clearwater-fv-test tests still compile (with the change to call create()), and that READ isn't defined after including the 'public' headers.

PR #320, opened 2015-07-03
Reviewer AMC: IP46Address mustn't use difference of s_addr values in compare functi…

steelyvoid:
…on as this overruns the integer return value.

Verified live at a deployment whose blacklisted IP addresses were previously causing this overflow regularly.


bossmc:
Fixes #321

bossmc:
Fine.

PR #323, opened 2015-07-06
Reviewer AMC: Extra sas log upstream

Tom-Rees:


src/httpconnection.cpp
```diff
@@ -737,6 +737,12 @@ HTTPCode HttpConnection::send_request(const std::string& path,                 /
           (num_http_504_responses >= 1) ||
           fatal_http_error)
       {
+        // Make a SAS log so that its clear that we have stopped retrying
+        // deliberately.
+        std::string reason = fatal_http_error
+             ? "Permanent error response received"
+             : "Temporary error response threshold exceeded";
```
mirw: Generally, I think the "SAS way" is not to pass this as a string - it increases SAS occupancy.  Instead, either pass as an enum or just create different SAS events for the different situations.  @AlexHockey should be able to confirm/recommend.

mirw:
@AlexHockey, can you confirm on the question above?

src/httpconnection.cpp
AlexHockey: Yes. Not only does in increase occupancy, it makes our SAS logs harder to localize in future. In this case I would recommend logging this as static parameter (with value 0 or 1) and define an enum in the resource bundle that renders to "Permanent error response received" or "Temporary error response threshold exceeded" (see for example the `ENABLED_DISABLED` enum in [the resource bundle](https://github.com/Metaswitch/clearwater-sas-resources/blob/master/clearwater_sas_resource_bundle.yaml#L1658))

bossmc:
@AlexHockey Can you please review as SAS-expert?

AlexHockey:
Generally looks good :-). One comment in this PR, plus a few in https://github.com/Metaswitch/clearwater-sas-resources/pull/70

Tom-Rees:
Hi Alex, could you please review these new changes? They should correspond with the bundle in [Metaswitch/clearwater-sas-resources/pull/70](https://github.com/Metaswitch/clearwater-sas-resources/pull/70)

src/httpconnection.cpp
```diff
@@ -695,6 +695,10 @@ HTTPCode HttpConnection::send_request(const std::string& path,                 /
           (num_http_504_responses >= 1) ||
           fatal_http_error)
       {
+        // Make a SAS log so that its clear that we have stopped retrying
+        // deliberately. 1 is Permanent error, 0 is Temporary error.
+        uint32_t reason = fatal_http_error ? 1 : 0;
```
AlexHockey: We try to avoid "hard coding" constants as literal values in the middle of a bunch of other code.  Also, the reason parameter isn't *really* an integer as it can only take certain predefined values. 

I think it would be better to define the constants in an enum (defined in the private section of HttpConnection). This code, and `sas_log_http_abort` would refer to this enum. This prevents accidentally passing a bad value into `sas_log_http_abort`.

The enum is also clearer when comparing he code to the resource bundle as there are enums in both places. You should probably also comment the enum in the cpp code to prompt anyone who changes it to also update the resource bundle. :-)

AlexHockey:
Hi Tom. Looks good, but I think it would be better to use an enum here (like you did in the resource bundle).

Tom-Rees:
HI Alex, could you take a look at this enum change please?

AlexHockey:
Looks good, I'm happy for you to merge. 

As this is a cpp-common change you will also need to update all our other repos that include cpp-common as a submodule. @steelyvoid will be able to help you with this.

PR #324, opened 2015-07-07
[Reviewer: Ellie] Switch time to REALTIME

rkd-msw:
Basically as discussed.

### DIFF
--- a/include/snmp_internal/snmp_time_period_table.h
+++ b/include/snmp_internal/snmp_time_period_table.h
@@ -85,7 +85,7 @@ public:
     void update_time()
     {
       struct timespec now;
-      clock_gettime(CLOCK_MONOTONIC_COARSE, &now);
+      clock_gettime(CLOCK_REALTIME_COARSE, &now);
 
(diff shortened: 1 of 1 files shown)


eleanor-merry:
Looks good. What testing have you done?

rkd-msw:
Good question - I've had a think about what testing would be useful here. I think:

* with this change, the 5-second row count will reset at x:y:00, x:y:05, x:y:10 etc. as documented
* before this change, the 5-second row count would reset at x:y:n, x:y:n+5, x:y:n+10 etc. - it would still update every 5 elapsed seconds, but not at a fixed offset (because the CLOCK_MONOTONIC epoch is arbitrary and not 1/1/1970)
* both before and after, if I set the clock back 3 hours, the count should still reset after 5 seconds - it shouldn't stay fixed for the next 3 hours and 5 seconds

I'm going to confirm that all of the above are true (by running snmpget and sleep in a loop).

rkd-msw:
I've confirmed all three.

PR #325, opened 2015-07-07
[Reviewer: Ellie] Prevent double free on SNMP shutdown

rkd-msw:
We shouldn't be calling `shutdown_agent()` ourselves (and don't in clearwater-fv-test). Tested live, fixes a crash I was seeing.


eleanor-merry:
Looks good

PR #326, opened 2015-07-09
[Reviewer: Ellie] Log net-snmp messages to the Clearwater log

rkd-msw:
Hopefully reasonably self-explanatory - netsnmp lets you provide a callback for logging, so I've done that and written a callback that logs to our normal log. (Previously, it went to syslog.)

### DIFF
--- a/src/snmp_agent.cpp
+++ b/src/snmp_agent.cpp
@@ -49,11 +49,56 @@ void* snmp_thread(void* data)
   return NULL;
 };
 
+int logging_callback(int majorID, int minorID, void *serverarg, void *clientarg)
+{
+  snmp_log_message *log_message = (snmp_log_message *) serverarg;
+  int snmp_prio = log_message->priority;
+  int clearwater_prio = Log::STATUS_LEVEL;
+
+  switch (snmp_prio) {
+    case LOG_EMERG:
+    case LOG_ALERT:
+    case LOG_CRIT:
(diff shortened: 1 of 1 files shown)


src/snmp_agent.cpp
```diff
@@ -49,12 +49,57 @@ void* snmp_thread(void* data)
   return NULL;
 };
 
+int logging_callback(int majorID, int minorID, void *serverarg, void *clientarg)
+{
+  snmp_log_message *log_message = (snmp_log_message *) serverarg;
```
eleanor-merry: spacing looks funny here

src/snmp_agent.cpp
```diff
@@ -49,12 +49,57 @@ void* snmp_thread(void* data)
...
+  int snmp_prio = log_message->priority;
```
eleanor-merry: Can you not just use priority as the name here?

eleanor-merry:
Looks good

src/snmp_agent.cpp
rkd-msw: I'd like to be clear that this is switching between the netsnmp priority levels and the Clearwater ones - hence the naming choices.

rkd-msw: Oh, you mean calling them snmp_priority and clearwater_priority. Yes.

PR #327, opened 2015-07-09
Added successFailCountTable class [reviewer: Rob]

alango:


include/snmp_success_fail_count_table.h
```diff
@@ -0,0 +1,80 @@
...
+struct RegistrationStatsTables
+{
+  SuccessFailCountTable* init_reg_tbl;
+  SuccessFailCountTable* re_reg_tbl;
+  SuccessFailCountTable* de_reg_tbl;
+};
+
+struct AuthenticationStatsTables
+{
+  SuccessFailCountTable* SIP_digest_auth_tbl;
```
rkd-msw: We usually make variable names all-lowercase - even for acronyms like SIP, IMS and AKA which would normally be capitalised.

src/snmp_success_fail_count_table.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "snmp_success_fail_count_table.h"
+#include "logger.h"
+
+namespace SNMP
+{
+
+// Storage for the underlying data
+struct SuccessFailCount
+{
+  uint64_t attempts;
```
rkd-msw: It would be better for this to be `std::atomic_uint_fast64_t` (like https://github.com/Metaswitch/cpp-common/blob/master/src/snmp_accumulator_table.cpp#L47) to avoid torn reads/writes (see http://preshing.com/20130618/atomic-vs-non-atomic-operations/ for more context).

rkd-msw:
Looks good - one comment about variable naming, one about atomic types.

src/snmp_success_fail_count_table.cpp
rkd-msw: @bossmc points out that data races are the risk here, not torn reads.

bossmc: Technically we won't see a torn read/write (that's "impossible" on a 64-bit integer [assuming this is an x86_64 {and the integer is aligned...}]), o.t.o.h. we will see data races.

PR #329, opened 2015-07-14
[Reviewer: Rob] Split the alarm code up to make it easy to subclass

eleanor-merry:
Rob, can you review this change to add a base class to the alarms

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -54,7 +54,7 @@ class AlarmState
 {
 public:
   AlarmState(const std::string& issuer,
-             AlarmDef::Index index,
+             const int index,
              AlarmDef::Severity severity);
 
   /// Queue request to update the alarm identified by index, to the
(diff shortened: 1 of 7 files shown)


src/alarm.cpp
```diff
@@ -211,7 +211,9 @@ bool AlarmReqAgent::zmq_init_sck()
     return false;
   }
 
-  std::string addr = "tcp://127.0.0.1:" + std::to_string(ZMQ_PORT);
+  std::string addr = "ipc:///var/run/clearwater/alarms";
+  TRC_DEBUG("AlarmRegAgent: addr=%s", addr.c_str());
```
rkd-msw: Is this a helpful log? (Also, Reg -> Req)

include/alarm.h
```diff
@@ -104,14 +100,31 @@ class Alarm
...
-  AlarmDef::Index _index;
+  const int _index;
   AlarmState _clear_state;
   AlarmState _set_state;
 
   std::atomic<bool> _alarmed;
+
+protected:
+  // Constructor. This is protected to prevent the BaseAlarm from being
+  // instantiated directly.
```
rkd-msw: Worth a comment to say that subclasses should just have a different constructor using an enum for the index, so that non-Clearwater components can use this infrastructure while defining a different set of alarms? It's not immediately clear to the reader what the rationale for this split is otherwise.

include/alarm.h
```diff
@@ -54,7 +54,7 @@ class AlarmState
 {
 public:
   AlarmState(const std::string& issuer,
```
rkd-msw: AlarmState is never created by anything other than BaseAlarm, so:
* does it need to be exposed in the header file (apart from just a forward declaration)?
* shoul everything be protected and `friend class BaseAlarm`?

rkd-msw:
Looks good.

PR #331, opened 2015-07-16
[Reviewer: Ellie] Attempt to shrink the per-node stats boilerplate

rkd-msw:
The big change here is the addition of an `add` method to SNMP::ManagedTable where you can provide the row directly, instead of having it constructed by new_row. This makes more sense for tables where all the rows are constructed up front.

I've also used NodeTypes a bit more widely - I'm not sure if this is sensible (I'm happy to back it out and use ints and a cast).


eleanor-merry:
Looks good.

Issue #246, opened 2015-02-27
IPv6 DNS servers aren't supported


rkd-msw:
Although Sprout's dnsresolver.cpp has code specifically to handle IPv6 servers, dnscachedresolver.cpp doesn't. That was fine when we relied on dnsmasq and always hardcoded 127.0.0.1, but now that we have multiple namespaces, signaling_dns_server needs to be able to support IPv6 addresses.

PR #340, opened 2015-07-19
[Reviewer: Andy] Allow IPv6 DNS servers

rkd-msw:
This fixes https://github.com/Metaswitch/cpp-common/issues/246, mostly by porting code from Sprout's ENUM code.

I've given the resolver three constructors - one with a single IP address as a string, one with multiple IP address as strings, and one taking IP46Addresses directly (so that the ENUM resolution code can plug into this).

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -115,28 +115,14 @@ DnsResult::~DnsResult()
   }
 }
 
-DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers) :
+DnsCachedResolver::DnsCachedResolver(const std::vector<IP46Address>& dns_servers) :
+  _dns_servers(dns_servers),
   _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
   _cache()
 {
   // Initialize the ares library.  This might have already been done by curl
   // but it's safe to do it twice.
   ares_library_init(ARES_LIB_INIT_ALL);
-
-  size_t max_servers = 3;
-  _dns_servers_count = std::min(max_servers, dns_servers.size());
-
-  TRC_STATUS("Creating Cached Resolver using servers:");
-  for (size_t i = 0; i < _dns_servers_count; i++)
-  {
-    TRC_STATUS("    %s", dns_servers[i].c_str());
-    // Parse the DNS server's IP address.
-    if (!inet_aton(dns_servers[i].c_str(), &(_dns_servers[i])))
-    {
-      TRC_ERROR("Failed to parse '%s' as IP address - defaulting to 127.0.0.1", dns_servers[i].c_str());
-      (void)inet_aton("127.0.0.1", &(_dns_servers[i]));
-    }
-  }
 
   // We store a DNSResolver in thread-local data, so create the thread-local
   // store.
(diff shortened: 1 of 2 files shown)


src/dnscachedresolver.cpp
```diff
@@ -149,34 +135,39 @@ DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers
...
+    }
+    else if (inet_pton(AF_INET6, dns_servers[i].c_str(), &(addr.addr.ipv6)))
+    {
+      addr.af = AF_INET6;
+    }
+    else
+    {
+      TRC_ERROR("Failed to parse '%s' as IP address - defaulting to 127.0.0.1", dns_servers[i].c_str());
+      addr.af = AF_INET;
+      (void)inet_aton("127.0.0.1", &(addr.addr.ipv4));
```
bossmc: Does it matter if this appears multiple times in the final list?

src/dnscachedresolver.cpp
```diff
@@ -149,34 +135,39 @@ DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers
   pthread_condattr_destroy(&cond_attr);
 }
 
-DnsCachedResolver::DnsCachedResolver(const std::string& dns_server) :
-  _cache_lock(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP),
-  _cache()
+DnsCachedResolver* DnsCachedResolver::from_server_ips(const std::vector<std::string>& dns_servers)
 {
-  TRC_STATUS("Creating Cached Resolver using server %s", dns_server.c_str());
+  std::vector<IP46Address> dns_server_ips;
```
bossmc: You can `reserve` this to `dns_servers.size()` so you might as well.

src/dnscachedresolver.cpp
```diff
@@ -767,8 +758,9 @@ DnsCachedResolver::DnsChannel* DnsCachedResolver::get_dns_channel()
   // Get the channel from the thread-local data, or create a new one if none
   // found.
   DnsChannel* channel = (DnsChannel*)pthread_getspecific(_thread_local);
+  size_t server_count = std::min((size_t)3u, _dns_servers.size());
```
bossmc: Worth a log if we're dropping server entries?

src/dnscachedresolver.cpp
```diff
@@ -791,6 +784,35 @@ DnsCachedResolver::DnsChannel* DnsCachedResolver::get_dns_channel()
...
+      {
+        memcpy(&ares_addr->addr.addr4, &server.addr.ipv4, sizeof(ares_addr->addr.addr4));
+      }
+      else
+      {
+        memcpy(&ares_addr->addr.addr6, &server.addr.ipv6, sizeof(ares_addr->addr.addr6));
+      }
+    }
+    
+    ares_set_servers(channel->channel, &(_ares_addrs[0]));
```
bossmc: Why `&_ares_addrs[0]` that's the same as `_ares_addrs` (or probably `&_ares_addrs`... C++ be crazy, yo).

bossmc:
Reviewed.

src/dnscachedresolver.cpp
rkd-msw: As discussed:

- it's a std::vector
- this is valid - pointers are only invalidated on reallocation (http://stackoverflow.com/a/8737812)

src/dnscachedresolver.cpp
rkd-msw: Sure.

src/dnscachedresolver.cpp
rkd-msw: Nope - and this preserves existing behaviour. (I tested with two invalid addresses when I wrote the "three DNS servers" code.)

src/dnscachedresolver.cpp
bossmc: Yep you're right, I was assuming that `_ares_addrs` was an array.  You can use [data()](http://www.cplusplus.com/reference/vector/vector/data/) rather that depending on the dereference/reference dance.

rkd-msw:
OK, I've made your suggested markups, and also decided to have three constructors (IP46Address list, std::string list, std::string) rather than factory methods, just to reduce the amount of updating I have to do.

PR #342, opened 2015-07-21
[Reviewer: Matt] Support closing Diameter connections

eleanor-merry:
Matt, can you please review this change to
* Register a handler for peer connections; this rejects CERs if the Diameter stack isn't accepting requests
* Add plumbing to allow users of the stack to shut down all the Diameter connections. 

Tested live


mirw:
As I think I commented on the design, `CommunicationMonitor` describes itself as being for the purpose of alarms.  We're not using it for alarms here, so I think we need to update the comments around `CommunicationMonitor`.

include/communicationmonitor.h
```diff
@@ -78,8 +78,8 @@ class CommunicationMonitor
   /// the caller it should be passed to avoid duplicate work.
   virtual void inform_failure(unsigned long now_ms = 0);
 
-private:
-  void update_alarm_state(unsigned long now_ms);
+protected:
+  virtual void update_alarm_state(unsigned long now_ms);
```
mirw: Why does `CedarCommunicationMonitor` need to override `update_alarm_state`?  Isn't it for closing TCP connections?

src/diameterstack.cpp
```diff
@@ -558,6 +570,12 @@ void Stack::wait_stopped()
   }
 }
 
+void Stack::close_connections()
```
mirw: Is there a reason why `close_connections` and `set_allow_connections` are independent?  Are there any situations in which you'd want to close connections but not block further connections (or vice-versa)?

PR #343, opened 2015-07-24
[reviewer: Rob] Success fail count by request type table

alango:


include/snmp_internal/snmp_counts_by_other_type_table.h
```diff
@@ -34,19 +34,20 @@
...
 
-template <class T> class CountsByNodeTypeTableImpl: public ManagedTable<T, int>
+template <class T> class CountsByOtherTypeTableImpl: public ManagedTable<T, int>
 {
 public:
-  CountsByNodeTypeTableImpl(std::string name,
-                            std::string tbl_oid):
+  CountsByOtherTypeTableImpl(std::string name,
+                             std::string tbl_oid,
+                             std::vector<int> other_types):
```
rkd-msw: You can just call this 'types' - you don't need the intermediate variable on line 58.

include/snmp_internal/snmp_counts_by_other_type_table.h
```diff
@@ -54,43 +55,44 @@ template <class T> class CountsByNodeTypeTableImpl: public ManagedTable<T, int>
                          { ASN_INTEGER , ASN_INTEGER }) // Types of the index columns
   {
     int n = 0;
-    std::vector<NodeTypes> nodes = { NodeTypes::SCSCF, NodeTypes::ICSCF, NodeTypes::BGCF };
+    std::vector<int> types = other_types;
+    // printf("first: %d\n", types.front());
```
rkd-msw: Debugging printf - you should take this back out.

src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "snmp_internal/snmp_includes.h"
+#include "snmp_internal/snmp_counts_by_other_type_table.h"
+#include "snmp_sip_request_types.h"
+#include "logger.h"
+
+namespace SNMP
+{
+
+// Storage for the underlying data
+struct SuccessFailCount
```
rkd-msw: Should these be atomic, as at https://github.com/Metaswitch/cpp-common/blob/master/src/snmp_success_fail_count_table.cpp#L46? (Actually, should you just pull SuccessFailCount  in a separate header file for common use?)

src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -0,0 +1,137 @@
...
+    // Construct and return a ColumnData with the appropriate values
+    ColumnData ret;
+    ret[1] = Value::integer(this->_index);
+    ret[2] = Value::integer(this->_type_index);
+    ret[3] = Value::uint(counts.attempts);
+    ret[4] = Value::uint(counts.successes);
+    ret[5] = Value::uint(counts.failures);
+    return ret;
+  }
+  static int get_count_size() { return 2; }
```
rkd-msw: It's a bit unintuitive that this is 2 rather than 3 (the number of columns) - worth tweaking the code that calls this, so that this actually represents the number of columns?

rkd-msw:
Looks good - a couple of comments.

PR #344, opened 2015-07-24
Reviewer RKD: Add new node types

graemerobertson:


include/snmp_node_types.h
```diff
@@ -62,6 +62,12 @@ enum NodeTypes
   EPDG = 17
 };
 
+static std::vector<int> node_types = { NodeTypes::SCSCF,
```
rkd-msw: Suggest you move this into `snmp_single_count_by_node_type_table.h` and have it as a static member of `SingleCountByNodeTypeTable` - `node_types` is quite a generic name so it should be namespaced.

rkd-msw:
One comment, then good.

PR #346, opened 2015-07-27
[Reviewer: Graeme] Allow the supported node types to be defined when creating a table

rkd-msw:
As discussed on Friday, this lets you define what node type rows should be in a table when you create it, rather than hardcoding a list in cpp-common.


graemerobertson:
Looks fine, but I think there is a corresponding change in clearwater-fv-test too? Anywhere else?

PR #348, opened 2015-07-27
[Reviewer: Andy] Fix memory leak failures in UT

rkd-msw:
Two fixes:

* DnsCachedResolver previously ignored 0.0.0.0 as an IP. I've restored this behaviour.
* We never delete DnsTsx objects (and didn't notice previously, because 0.0.0.0 was our IP address in UT...). They now self-destruct at the end of their useful life.

I'm going to throw together a test in clearwater-fv-test to query Google Public DNS and ensure that there are no other DnsCachedResolver leaks.

I might also fake out ares_query() so that the Sprout UTs can hit all of DnsCachedResolver.

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -138,11 +138,16 @@ void DnsCachedResolver::init(const std::vector<IP46Address>& dns_servers)
 void DnsCachedResolver::init_from_server_ips(const std::vector<std::string>& dns_servers)
 {
   std::vector<IP46Address> dns_server_ips;
-  dns_server_ips.reserve(dns_servers.size());
 
   TRC_STATUS("Creating Cached Resolver using servers:");
   for (size_t i = 0; i < dns_servers.size(); i++)
   {
+    if (dns_servers[i] == "0.0.0.0")
+    {
+      // Skip this DNS server
+      continue;
+    }
+
     IP46Address addr;
     TRC_STATUS("    %s", dns_servers[i].c_str());
     // Parse the DNS server's IP address.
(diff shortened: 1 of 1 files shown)


src/dnscachedresolver.cpp
```diff
@@ -138,11 +138,16 @@ void DnsCachedResolver::init(const std::vector<IP46Address>& dns_servers)
...
 
   TRC_STATUS("Creating Cached Resolver using servers:");
   for (size_t i = 0; i < dns_servers.size(); i++)
   {
+    if (dns_servers[i] == "0.0.0.0")
+    {
+      // Skip this DNS server
+      continue;
+    }
+
```
bossmc: Should we extend this a little, we have code later that says "if it's not 0.0.0.0 but it doesn't parse, use 127.0.0.1".  I think it's nicer to say "if it's 0.0.0.0 or it doesn't parse, skip it" then "if there's no IPs in the list, use 127.0.0.1".  Removes the inconsistencies.

bossmc:
One comment for sanity, otherwise looks sensible.

src/dnscachedresolver.cpp
rkd-msw: I think that leads to exactly the behaviour we don't want - a DnsCachedResolver initialised with 0.0.0.0 would actually use 127.0.0.1 and make real DNS queries in UT.

bossmc: Ah true, fine, leave it as is.

rkd-msw:
Fixes #350 .

PR #349, opened 2015-07-28
[Reviewer: Andy] Consistency changes upstream

Tom-Rees:
Hi Andy,

Would you mind reviewing these changes please?

It has associated changes in homestead (and others). I'll link the pull requests together once the others are created.

Cheers,
Tom

### DIFF
--- a/src/cassandra_store.cpp
+++ b/src/cassandra_store.cpp
@@ -661,32 +661,45 @@ put_columns(const std::vector<RowColumns>& to_put,
 // -  The arguments for the underlying get method.
 //
 // It works as follows:
-// -  Call the underlying method with a consistency level of ONE.
-// -  If this raises a NotFoundException, try again with a consistency level of
-//    QUORUM.
-// -  If this fails again with either NotFoundException or UnavailableException
-//    (meaning the necessary servers are not currently available), re-throw the
-//    original exception.
+// -  Call the underlying method with a consistency level of LOCAL_QUORUM.
+//    If successful, this indicates that we've managed to find multiple nodes
+//    with the same data, and so we can trust the response (if we're wrong, it
+//    will be because several local nodes failed or were down simultaneously
+//    which we can consider a non-mainline failure case for which some level of
+//    service impact is acceptable)
+// -  If this raises an UnavailableException (in other words, we couldn't
+//    contact a quorum of servers on the local node), try again with a
+//    consistency level of QUORUM.  This is going to attempt to get a quorum of
+//    responses from BOTH sites in a GR system, so is inevitably slower than
+//    LOCAL_QUORUM.  However, we cannot just drop straight through to the ONE
+//    level as this is the expected behaviour in a 2+2 GR system when we are
+//    restarting one of the homesteads (e.g. thanks to an upgrade) and we can't
+//    run the risk that the ONE read might hit another recently restarted
+//    homestead that is still out of date.
+// -  If this *also* fails with UnavailableException, perform a ONE read.  In
+//    this case at least half of the servers in the cluster are down, and so
+//    we are already in error recovery mode in which some service impact is not
+//    unexpected, so the risk that we might return out of date data is
+//    acceptable.
+//
 #define HA(METHOD, ...)                                                      \
         try                                                                  \
         {                                                                    \
-          METHOD(__VA_ARGS__, ConsistencyLevel::ONE);                        \
+          METHOD(__VA_ARGS__, ConsistencyLevel::LOCAL_QUORUM);               \
         }                                                                    \
-        catch(NotFoundException& nfe)                                        \
+        catch(UnavailableException)                                          \
         {                                                                    \
           TRC_DEBUG("Failed ONE read for %s. Try QUORUM", #METHOD);          \
+          // SASLOG "LOCAL_QUORUM didn't work"                               \
                                                                              \
           try                                                                \
           {                                                                  \
             METHOD(__VA_ARGS__, ConsistencyLevel::QUORUM);                   \
           }                                                                  \
-          catch(NotFoundException)                                           \
-          {                                                                  \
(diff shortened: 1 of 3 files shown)


src/cassandra_store.cpp
```diff
@@ -653,6 +654,14 @@ put_columns(const std::vector<RowColumns>& to_put,
...
 }
 
+// A map from quorum consistency levels to their corresponding SAS enum value.
+// Any changes made to this must be consistently applied to the sas resource
+// bundle.
+std::map<cass::ConsistencyLevel::type, uint32_t> quorum_consistency_map =
+  {
+    {cass::ConsistencyLevel::LOCAL_QUORUM, 0},
+    {cass::ConsistencyLevel::QUORUM,       1},
+  };
```
bossmc: Indenting is off here.

src/cassandra_store.cpp
```diff
@@ -661,32 +670,54 @@ put_columns(const std::vector<RowColumns>& to_put,
...
-//    original exception.
-#define HA(METHOD, ...)                                                      \
+// -  Call the underlying method with a consistency level of LOCAL_QUORUM.
+//    If successful, this indicates that we've managed to find multiple nodes
+//    with the same data, and so we can trust the response (if we're wrong, it
+//    will be because several local nodes failed or were down simultaneously
+//    which we can consider a non-mainline failure case for which some level of
+//    service impact is acceptable)
+// -  If this raises an UnavailableException (in other words, we couldn't
+//    contact a quorum of servers on the local node), try again with a
```
bossmc: "in the local datacenter"

src/cassandra_store.cpp
```diff
@@ -661,32 +670,54 @@ put_columns(const std::vector<RowColumns>& to_put,
...
+//    consistency level of QUORUM.  This is going to attempt to get a quorum of
+//    responses from BOTH sites in a GR system, so is inevitably slower than
+//    LOCAL_QUORUM.  However, we cannot just drop straight through to the ONE
+//    level as this is the expected behaviour in a 2+2 GR system when we are
+//    restarting one of the homesteads (e.g. thanks to an upgrade) and we can't
```
bossmc: This function is generic to all users of CassandraStore, not just homestead.

src/cassandra_store.cpp
bossmc: Actually, using a map here has runtime costs (and we never need to translate from a dynamically learned key to a value so it's not gaining us anything).  Should we just use an `enum struct` instead?

src/cassandra_store.cpp
```diff
@@ -661,32 +670,54 @@ put_columns(const std::vector<RowColumns>& to_put,
...
         {                                                                    \
-          METHOD(__VA_ARGS__, ConsistencyLevel::ONE);                        \
+          METHOD(__VA_ARGS__, ConsistencyLevel::LOCAL_QUORUM);               \
         }                                                                    \
-        catch(NotFoundException& nfe)                                        \
+        catch(UnavailableException& ue)                                      \
         {                                                                    \
           TRC_DEBUG("Failed ONE read for %s. Try QUORUM", #METHOD);          \
-                                                                             \
+          /* SASLOG "LOCAL_QUORUM didn't work" */                            \
```
bossmc: Remove this comment.

src/cassandra_store.cpp
```diff
@@ -661,32 +670,54 @@ put_columns(const std::vector<RowColumns>& to_put,
...
           }                                                                  \
-          catch(NotFoundException)                                           \
-          {                                                                  \
-            throw nfe;                                                       \
-          }                                                                  \
-          catch(UnavailableException)                                        \
+          catch(UnavailableException& ue)                                    \
           {                                                                  \
-            throw nfe;                                                       \
+            /* SASLOG "QUORUM didn't work either" */                         \
```
bossmc: Ditto

bossmc:
Looks mostly fine, some comments inline.

Good explanatory comment above the macro BTW, thanks!

src/cassandra_store.cpp
Tom-Rees: I could use something like:
```c++
enum class Quorum_Consistency_Levels
{
  LOCAL_QUORUM = 0,
  QUORUM = 1
};
```
And then use this enum as the argument to `add_static_param`. But I don't like the way that this is decoupled from the cassandra enums. One way to re-couple would be with an enum constructor of type `cass::enum -> Quorum_Consistency_Levels`, or similar function, but I don't know how this compares to a map in performance terms.
Or maybe I have misunderstood something completely?

bossmc: I was suggesting the former.  Currently there is no coupling between the `::cass` enum and yours at runtime (your calls to `quorum_consistency_map[...]` hard-code their parameters).

Can you write 

    enum class Quorum_Consistency_Levels
    {
      LOCAL_QUORUM = ::cass::ConsistencyLevel::LOCAL_QUORUM,
      QUORUM = ::cass::ConsistencyLevel::QUORUM
    };

which would give us the best of both worlds.

Tom-Rees: I can, providing that we know the values of the two cassandra enums and we also know that they won't change. Otherwise the values here may become out of sync with those in the sas bundle.

bossmc: Ah yes... that makes sense.  Cassandra enums will only change when we change cassandra version (as that's the only time they're allowed to change the API) so I'd be reasonably confident of making this change and us spotting the issue if we ever upgrade Cassandra again (we've only done it once in 3 years...)

Tom-Rees:
Hi Andy, 
Would you mind checking my feedback changes please?
I'm also trying to track down all files affected by this interface change. So far I have fixed files in memento and homestead. Do you know of any that i'm missing and the best way to check for others?

bossmc:
Markups look fine.

PR #352, opened 2015-07-29
[reviewer: Rob] SIP request type string to request type enum method

alango:


rkd-msw:
Looks good. No comments.

PR #357, opened 2015-07-31
[reviewer: Rob] split out init_snmp function

alango:


src/snmp_agent.cpp
```diff
@@ -111,7 +111,12 @@ int snmp_setup(const char* name)
   {
     TRC_STATUS("AgentX agent initialised");
   }
+  return 0;
```
rkd-msw: Should we just return `rc` directly (and remove the `return 1`above)?

rkd-msw:
Looks good, one minor comment.

PR #360, opened 2015-08-10
[Reviewer: Matt] - Adding overload statistics

TygerBurning:
Add new class ContinuousAccumulatorTable and rename the old AccumulatorTable to EventAccumulatorTable.

The reset() function on a SNMP table now expects two parameters, the previous table and the period start time of this period.


include/snmp_internal/snmp_time_period_table.h
```diff
@@ -72,13 +72,17 @@ template <class T> class TimeBasedRow : public Row
     CurrentAndPrevious(int interval):
       current(&a),
       previous(&b),
-      _interval(interval),
-      _tick(0),
+      _interval(interval), // In s.
```
mirw: Might be worth spelling out "seconds" - I know it's an SI unit, but there's not much context.

include/snmp_internal/snmp_time_period_table.h
```diff
@@ -72,13 +72,17 @@ template <class T> class TimeBasedRow : public Row
...
     {
-      a.reset();
-      b.reset();
+      struct timespec now;
+      clock_gettime(CLOCK_REALTIME_COARSE, &now);
+      uint64_t time_now_ms = (now.tv_sec * 1000) + (now.tv_nsec / 1000000);
+
+      _tick = (now.tv_sec / _interval);
+      a.reset(time_now_ms, NULL);
+      b.reset(time_now_ms - (interval*1000), NULL);
```
mirw: Is this second reset correct?  Why does the period start one interval's duration before the current time?

include/snmp_internal/snmp_time_period_table.h
```diff
@@ -129,6 +134,8 @@ template <class T> class TimeBasedRow : public Row
     View(CurrentAndPrevious* data): _data(data) {};
     virtual ~View() {};
     virtual T* get_data() = 0;
+    // Return interval in ms
+    uint32_t get_interval_ms() { return (this->_data->get_interval()) * 1000; }
```
mirw: Why are we inconsistent about seconds vs milliseconds?

src/load_monitor.cpp
```diff
@@ -81,7 +83,12 @@ void TokenBucket::replenish_bucket()
 }
 
 LoadMonitor::LoadMonitor(int init_target_latency, int max_bucket_size,
-                         float init_token_rate, float init_min_token_rate)
+                         float init_token_rate, float init_min_token_rate,
+                         SNMP::ContinuousAccumulatorTable* token_rate_tbl,
```
mirw: Convention is to call the parameter the same thing as the member, except with the `_` stripped off.  I'd also suggest using the `member_name(parameter_name)` notation rather than setting them explicitly in the constructor - it's more concise and more restrictive (so you know you don't need to read it in as much detail, as it can't be doing clever stuff).

src/load_monitor.cpp
```diff
@@ -115,6 +122,15 @@ LoadMonitor::LoadMonitor(int init_target_latency, int max_bucket_size,
...
   clock_gettime(CLOCK_MONOTONIC_COARSE, &last_adjustment_time);
   min_token_rate = init_min_token_rate;
+  _token_rate_table = token_rate_tbl;
+  _smoothed_latency_scalar = smoothed_latency_sclr;
+  _target_latency_scalar = target_latency_sclr;
+  _penalties_scalar = penalties_sclr;
+  _token_rate_scalar = token_rate_sclr;
+
+  // As this statistics reporting is continuous, we should
+  // publish the statistics when initialised.
```
mirw: Good comment!

src/load_monitor.cpp
```diff
@@ -195,9 +211,11 @@ void LoadMonitor::request_complete(int latency)
 
       // latency is above where we want it to be, or we are getting overload responses from
       // Homer/Homestead, so adjust the rate downwards by a multiplicative factor
+
+      float new_rate = bucket.rate;
```
mirw: Why is this change necessary?

src/load_monitor.cpp
```diff
@@ -237,3 +257,27 @@ void LoadMonitor::request_complete(int latency)
...
+  {
+    _penalties_scalar->value = penalties;
+  }
+  if (_token_rate_table != NULL)
+  {
+    _token_rate_table->accumulate(bucket.rate);
+  }
+  if (_token_rate_scalar != NULL)
+  {
+  _token_rate_scalar->value = bucket.rate;
```
mirw: Indentation.

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+namespace SNMP
+{
+
+// Storage for the underlying data
+struct ContinuousStatistics
+{
+  std::atomic_uint_fast64_t count;
+  std::atomic_uint_fast64_t current_value;
+  std::atomic_uint_fast64_t time_lastupdate_ms;    // In ms.
```
mirw: Good to have `_ms` in the variable name.  I'm not sure the comment is required.

mirw: Why the odd underscore spacing?  Why isn't this `time_last_update_ms`?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+    current_data->count++;
+
+    // Compute the updated sum and sqsum based on the previous values, dependant on
+    // how long since an update happened. Additionally update the sum of squares as a
+    // rolling total, and update the time of the last update. Also maintain a
+    // current value held, that can be used if the period ends.
+    struct timespec now;
+    clock_gettime(CLOCK_REALTIME_COARSE, &now);
+
+    uint64_t time_sincelastupdate = ((now.tv_sec * 1000) + (now.tv_nsec / 1000000))
```
mirw: Same comment as above regarding underscore use?  Why not `time_since_last_update`?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  void accumulate_internal(ContinuousAccumulatorRow::CurrentAndPrevious& data, uint32_t sample)
+  {
+    ContinuousStatistics* current_data = data.get_current();
+
+    TRC_DEBUG("Calling accumulate with value: (%uui)", sample);
+
+    current_data->count++;
+
+    // Compute the updated sum and sqsum based on the previous values, dependant on
```
mirw: (Massive pedantry.)  I think the adjective is depend*e*nt.

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+    // how long since an update happened. Additionally update the sum of squares as a
+    // rolling total, and update the time of the last update. Also maintain a
+    // current value held, that can be used if the period ends.
```
mirw: Good comment.

include/snmp_internal/snmp_time_period_table.h
```diff
@@ -72,13 +72,17 @@ template <class T> class TimeBasedRow : public Row
...
-      _interval(interval),
-      _tick(0),
+      _interval(interval), // In s.
       a(),
       b()
     {
-      a.reset();
-      b.reset();
+      struct timespec now;
+      clock_gettime(CLOCK_REALTIME_COARSE, &now);
```
mirw: Just for my education, why do we go for `CLOCK_REALTIME_COARSE`?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+        break;
+    }
+    return new ContinuousAccumulatorRow(index, view);
+  }
+
+  void accumulate_internal(ContinuousAccumulatorRow::CurrentAndPrevious& data, uint32_t sample)
+  {
+    ContinuousStatistics* current_data = data.get_current();
+
+    TRC_DEBUG("Calling accumulate with value: (%uui)", sample);
```
mirw: Maybe better to focus on what the code is doing in the trace statement, e.g. `"Accumulating sample (%uui) into statistic"`?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  count.store(0);
+  sum.store(0);
+  sqsum.store(0);
+
+  if (previous != NULL)
+  {
+    current_value.store(previous->current_value.load());
+    lwm.store(previous->current_value.load());
+    hwm.store(previous->current_value.load());
```
mirw: Why do we update LWM and HWM from the previous period?  (I think we should initialize them both to the current value.)

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+  }
+  else
+  {
+    current_value.store(0);
+    lwm.store(0);
```
mirw: Doesn't this mean that we can never get a new LWM - how will anything be less than 0?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+};
+
+void ContinuousStatistics::reset(uint64_t periodstart_ms, ContinuousStatistics* previous)
+{
+  struct timespec now;
+  clock_gettime(CLOCK_REALTIME_COARSE, &now);
+
+  TRC_DEBUG("New period created at time: (%uui)", periodstart_ms);
+
+  count.store(0);
```
mirw: Can we have a high-level comment about what we're doing here?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+  uint_fast64_t sum = accumulated->sum.load();
+  uint_fast64_t sqsum = accumulated->sqsum.load();
+
+  TRC_DEBUG("Avg: (%uui)", avg);
+  TRC_DEBUG("time_lastupdate: (%uui)", time_lastupdate_ms);
+  TRC_DEBUG("time_periodstart: (%uui)", time_periodstart_ms);
+  TRC_DEBUG("sum: (%uui)", sum);
+  TRC_DEBUG("sqsum: (%uui)", sqsum);
+
+  if (count > 0)
```
mirw: Is this the right test?  Can't count be 0 but the average still be non-zero?  Actually, I see this does work, but it could definitely do with a comment upfront.

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+  {
+    time_lastupdate_ms = accumulated->time_lastupdate_ms.load();
+    time_periodstart_ms = accumulated->time_periodstart_ms.load();
+    sum = accumulated->sum.load();
+    sqsum = accumulated->sqsum.load();
```
mirw: Didn't you do these just above?  Can we remove the duplication?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+    uint64_t time_sincelastupdate_ms = time_comesfirst_ms - time_lastupdate_ms;
+    accumulated->time_lastupdate_ms.store(time_comesfirst_ms);
+
+    // Calculate the average and the variance from the stored average/time of last upadted
+    // and sum-of-squares.
+    sum += time_sincelastupdate_ms * current_value;
+    accumulated->sum.store(sum);
+    sqsum += time_sincelastupdate_ms * current_value * current_value;
+    accumulated->sqsum.store(sqsum);
+    avg = sum/(time_comesfirst_ms - time_periodstart_ms);
```
mirw: What if `time_comesfirst_ms - time_periodstart_ms` is 0?

src/snmp_continuous_accumulator_table.cpp
```diff
@@ -0,0 +1,269 @@
...
+  if (count > 0)
+  {
+    time_lastupdate_ms = accumulated->time_lastupdate_ms.load();
+    time_periodstart_ms = accumulated->time_periodstart_ms.load();
+    sum = accumulated->sum.load();
+    sqsum = accumulated->sqsum.load();
+
+    // Distinguish between periods and only take the relevant value for
+    // time_sincelastupdate as once a period is finished, we should not
+    // recalculate the values.
```
mirw: I think I know what you're getting at in this comment, but maybe you could be a bit clearer?

mirw:
Looks good.  Not many comments - and most of them are minor ones about commenting.

include/snmp_internal/snmp_time_period_table.h
TygerBurning: I agree. I've made the switch over that all statistics should be reported in ms. Should I document this change (note relevant variable name change from `interval` to `inverval_ms` which should indicate change)

include/snmp_internal/snmp_time_period_table.h
TygerBurning: Because it's more intuitive. As we don't implement a way for the *end* of the period to not be boundary aligned, we would like the previous period to certainly finish before the current period - and without taking the interval gap away, we cannot ensure that two distinct time period will not finish at the same point. Without this, it may also result in a problem in get_columns(), as it will consider it to still be current.

src/snmp_continuous_accumulator_table.cpp
TygerBurning: This is only true while the previous table was NULL (i.e. in the first case). This means the first table will always have a LWM of 0, but subsequent tables will be more accurate. I've updated this to reflect that the LWM should take the first value it sees, but old tables should still print 0 (rather than 463XXXX)

include/snmp_internal/snmp_time_period_table.h
TygerBurning: This is a valid point. I use this to remain consistent with current methods of measuring clock, but it feels like MONOTONIC may be a better measure (as we are generally comparing two times)

mirw: If it's the same as everywhere else, fine.

mirw:
Fine.  Feel free to merge.

PR #362, opened 2015-08-12
[Reviewer: MIRW] Add code to retrieve socket in another namespace

rkd-msw:


src/namespace_hop.cpp
```diff
@@ -0,0 +1,137 @@
...
+  message.msg_namelen = 0;
+  message.msg_control = ctrl_buf;
+  message.msg_controllen = CMSG_SPACE(sizeof(int));
+  message.msg_iov = &iov;
+  message.msg_iovlen = 1;
+
+  int res = recvmsg(socket, &message, 0);
+  if (res <= 0)
+  {
+    fprintf(stderr, "recvmsg returned %d (%d %s)\n", res, errno, strerror(errno));
```
mirw: Is it right to write to stderr?  Normally, we'd TRC_WARN (or whatever) - will this get picked up in normal log files?

mirw:
Code looks code.

src/namespace_hop.cpp
rkd-msw: The main user of this code for now is a Python process - where the TRC_WARN logs wouldn't get caught but stderr logs will.

Long-term we should refactor the cpp-common logging system, so you can pass in a callback that calls the Python logger. I'm happy to raise an issue to track that deficiency.

mirw: Yes, please raise an issue.  Are there any C++ users of this code?  (I don't know whether the sas-client stuff uses it, as I haven't reviewed that.)  If there are, will stderr output appear in those logs?

rkd-msw: I've had a think and decided to switch to TRC_DEBUG-style calls:

- in C++ code this works perfectly
- in Python code, we'll use the default logger which will print to stdout, and will be caught just the same

I'll test that both assertions are true. I'll also raise an issue to cover better long-term Python integration.

mirw: Sounds great - thanks!

PR #365, opened 2015-08-13
[Reviewer: Matt] - Made the update time function in time period table more consistent

TygerBurning:
Previously the update time function could be called and it would have a (slightly) different time to when the original function was called. 

It is now required to pass a timspec value into update time, which will generally be when the function will be called, to ensure consistency

### DIFF
--- a/include/snmp_internal/snmp_time_period_table.h
+++ b/include/snmp_internal/snmp_time_period_table.h
@@ -86,11 +86,8 @@ public:
     }
 
     // Rolls the current period over into the previous period if necessary.
-    void update_time()
+    void update_time(timespec now)
     {
-      struct timespec now;
-      clock_gettime(CLOCK_REALTIME_COARSE, &now);
-
       // Count of how many _interval periods have passed since the epoch
       uint64_t new_tick = (now.tv_sec / (_interval_ms / 1000));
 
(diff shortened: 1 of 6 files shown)


include/snmp_internal/snmp_time_period_table.h
```diff
@@ -86,11 +86,8 @@ template <class T> class TimeBasedRow : public Row
     }
 
     // Rolls the current period over into the previous period if necessary.
-    void update_time()
+    void update_time(timespec now)
```
mirw: Doesn't this need to be **struct** timespec?

src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -97,24 +100,33 @@ class SuccessFailCountByRequestTypeTableImpl: public CountsByOtherTypeTableImpl<
...
-  {} 
-  
+  {}
+
   void increment_attempts(SIPRequestTypes type)
   {
-    five_second[type]->get_current()->attempts++;
-    five_minute[type]->get_current()->attempts++;
+    struct timespec now;
+    clock_gettime(CLOCK_REALTIME_COARSE, &now);
```
mirw: This feels quite verbose (in this case, where we don't care about using the same time everywhere).  Maybe we should define a `get_current()` that does the `clock_gettime` call and then calls into `get_current(struct timespec*)`?

mirw:
Looks good.  A couple of comments inline.

TygerBurning:
Updated - I agree on overloading get_current() method

mirw:
Looks good.  Was the `struct` necessary, and if so, how did you not catch this earlier?

TygerBurning:
No discernible difference - at least none caught by the compiler or any unit tests. I changed to `struct` to remain consistent with previous code - although maybe the important difference is I'm not going to be writing to this `struct`, only reading from it - so maybe it's less important.

PR #366, opened 2015-08-14
[Reviewer: Matt] Updated get_data() method to take parameter

TygerBurning:
Updated code to match current implementation of get_data()


mirw:
Looks good.

PR #367, opened 2015-08-17
[Reviewer: Matt] Conditioning on the wrong variable fix

TygerBurning:
Resulting in the variance being 0 for a lot of the time. Also catches a divides by 0 error (that the initial conditional was meant to fix!)


mirw:
Looks good.  Please can you confirm you've rerun UT and FV test?

TygerBurning:
Yes, all tests pass as expected

PR #368, opened 2015-08-17
Reviewer: Andy] - Added a reverse time function

TygerBurning:
Used this function to test overflow problems in Chronos

### DIFF
--- a/test_utils/test_interposer.cpp
+++ b/test_utils/test_interposer.cpp
@@ -125,6 +125,15 @@ void cwtest_advance_time_ms(long delta_ms)  ///< Delta to add to the current off
   pthread_mutex_unlock(&time_lock);
 }
 
+/// Go back a way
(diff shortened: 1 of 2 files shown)


test_utils/test_interposer.cpp
```diff
@@ -125,6 +125,15 @@ void cwtest_advance_time_ms(long delta_ms)  ///< Delta to add to the current off
   pthread_mutex_unlock(&time_lock);
 }
 
+/// Go back a way
+void cwtest_reverse_time_ms(long delta_ms) ///< Delta to *remove* to the current offset applied to returned times (in ms)
```
bossmc: In English we generally remove "from" something...

bossmc: Also, the stars around remove are only meaningful since this is being added in contrast to `cwtest_advance_time_ms` once they both the stars make less sense.

PR #371, opened 2015-09-04
[reviewer: Rob] Cx stats table

alango:


src/snmp_cx_counter_table.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include "snmp_internal/snmp_includes.h"
+#include "snmp_internal/snmp_time_period_and_double_indexed_table.h"
+#include "snmp_cx_counter_table.h"
+#include <vector>
+#include "logger.h"
+
+namespace SNMP
+{
+
+static std::vector<int> base_result_codes =
```
rkd-msw: Worth referencing the relevant IMS/IETF specs in the comments?

src/snmp_cx_counter_table.cpp
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+};
+
+class CxCounterTableImpl: public ManagedTable<CxCounterRow, int>, public CxCounterTable
+{
+public:
+  CxCounterTableImpl(std::string name,
+                     std::string tbl_oid,
+                     std::vector<int> base_codes = base_result_codes,
```
rkd-msw: Is it ever going to be useful to set these to something other than the default? If not, I think we should use `base_result_codes` and `_3gpp_result_codes ` directly.

rkd-msw:
Looks good.

include/snmp_cx_counter_table.h
```diff
@@ -0,0 +1,72 @@
...
+//   - increment a single counter over time
+//   - report a single column for each time period with that count
+//
+// This is defined as an interface in order not to pollute the codebase with netsnmp include files
+// (which indiscriminately #define things like READ and WRITE).
+
+namespace SNMP
+{
+
+enum AppId
```
rkd-msw: DiameterAppId might be a better name - 'AppId' is a bit too generic.

PR #375, opened 2015-09-09
Reviewer AMC: Sas logs for 'no diameter peer' conditions

steelyvoid:
Adds support for a couple of SAS logs that add context in "HSS connection failed" SAS traces

-  A log that says "No diameter peers.  DNS records probably absent for the diameter realm"
-  A log that says "No connected diameter peers.  Check connectivity and for realm mismatch"

Tested by configuring as follows and checking that REGISTER SAS trace is as expected

-  an hss_realm with correctly configured DNS records.  SAS trace shows success and the new logs are not made
-  an hss_realm with DNS records for an HSS configured with a different realm.  SAS trace shows failure and the second SAS log above.
-  an hss_realm with no DNS records.  SAS trace shows failure and the first SAS log above

Also tested hss_hostname configurations to check for unexpected behaviour (the second log is NOT made if the hss_hostname is invalid - this is a known limitation of the fix, stemming from the fact that DiameterStack does not track the connectivity status of the hss_hostname peer).

The logs are made before submitting the requests to freeDiameter (as opposed to following failure response) as some messages are not explicitly responded to via fd hooks, so notification of failure would not necessarily arrive at DiameterStack.

PRs for clearwater-sas-resources to follow


src/diameterstack.cpp
```diff
@@ -258,6 +259,58 @@ void Stack::fd_error_hook_cb(enum fd_hook_type type,
...
+  //
+  // We don't bother looking at peers for whom connected() returns "false" as
+  // these are peers that have recently been started, and for which
+  // initial connectivity is still ongoing (so they should be considered "good"
+  // peers until we determine otherwise).
+  //
+  // The SAS logs won't appear if hss_hostname has been configured (rather than
+  // hss_realm) as we have no visibility of the connectedness of the single
+  // peer in this case (there is always exactly one peer and freeDiameter
+  // automatically retries connection failures on our behalf).
```
bossmc: Bit picky, but the bulk of this comment really belongs at the other end of this function call.  `Stack` doesn't care how the caller determines the number of peers, only that it provides a number here.  I'd split the "what counts as a peer" bits out and move them to the `RealmManager`.

bossmc:
One tedious comment, otherwise fine.

src/diameterstack.cpp
bossmc: Hmmm, on a re-read, I think I agree with you that this comment belongs here.  We're really discussing the difference between "peers we know about" and "peers we're connected to" which are non-equivalent sets.

bossmc:
I've taken back my comment, fix looks good.

PR #378, opened 2015-09-11
[Reviewer: Rob] Allow all in the httpstack

eleanor-merry:
Rob, can you review this fix to only listen on the local address if we haven't already tried to listen on all. 

Tested live

### DIFF
--- a/src/httpstack.cpp
+++ b/src/httpstack.cpp
@@ -213,7 +213,11 @@ void HttpStack::start(evhtp_thread_init_cb init_cb)
     throw Exception("evhtp_bind_socket", rc); // LCOV_EXCL_LINE
   }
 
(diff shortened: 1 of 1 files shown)


src/httpstack.cpp
```diff
@@ -213,7 +213,11 @@ void HttpStack::start(evhtp_thread_init_cb init_cb)
     throw Exception("evhtp_bind_socket", rc); // LCOV_EXCL_LINE
   }
 
-  if (local_bind_address != full_bind_address) {
+  if ((local_bind_address != full_bind_address) &&
+      ((full_bind_address != "0.0.0.0") || (full_bind_address != "::")))
```
rkd-msw: What happens if it's `[::]`?

rkd-msw:
Looks good.

PR #379, opened 2015-09-16
[Reviewer: Alex] Hard code tombstone lifetime

eleanor-merry:
Alex, can you review this change to hard code the tombstone lifetime?

Fixes https://github.com/Metaswitch/clearwater-etcd/issues/107

### DIFF
--- a/src/memcached_config.cpp
+++ b/src/memcached_config.cpp
@@ -42,7 +42,7 @@
 #include "memcached_config.h"
 
 /// The default lifetime (in seconds) of tombstones written to memcached.
-static const int DEFAULT_TOMBSTONE_LIFETIME = 0;
+static const int DEFAULT_TOMBSTONE_LIFETIME = 30 * 60;
 
 MemcachedConfig::MemcachedConfig() :
   servers(), new_servers(), tombstone_lifetime(DEFAULT_TOMBSTONE_LIFETIME)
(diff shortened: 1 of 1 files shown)


src/memcached_config.cpp
```diff
@@ -121,17 +121,10 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
...
-          config.tombstone_lifetime = atoi(value.c_str());
-
-          if (std::to_string(config.tombstone_lifetime) != value)
-          {
-            TRC_ERROR("Config contained an invalid tombstone_lifetime line:\n%s",
-                      line.c_str());
-            return false;
-          }
+          // We no longer support this key being configurable.
+          TRC_ERROR("Config file had key 'tombstone_lifetime', which is no " \
```
AlexHockey: Error seems a bit strong - what if someone had set this in their config? Warning or status feel better.

Also you don't need the trailing `\`

AlexHockey:
A couple of minor comments.

src/memcached_config.cpp
```diff
@@ -121,17 +121,10 @@ bool MemcachedConfigFileReader::read_config(MemcachedConfig& config)
...
+                    "longer supported");
+          return false;
```
AlexHockey: Is it right to fail the parse if the option is set?

PR #380, opened 2015-09-17
[Reviewer: Rob] Only increment statistics for response codes we know about

mirw:
Rob,

Please can you review this safety fix to https://github.com/Metaswitch/homestead/issues/255?  I think there's a question as to whether we should also have an "other" counter, to count any other response codes we receive - thoughts?

Matt

### DIFF
--- a/src/snmp_cx_counter_table.cpp
+++ b/src/snmp_cx_counter_table.cpp
@@ -161,21 +161,31 @@ public:
                                     new CxCounterRow::PreviousView((*five_minute)[code])));
   }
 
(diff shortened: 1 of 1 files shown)


mirw:
Note that I haven't reproduced this problem (I'm waiting for Doug to confirm what triggers it), but I have run UT and live test cleanly with my fix in.

rkd-msw:
Looks good, and having an unknown row would be sensible (implemented the same way as the timeout count?).

rkd-msw:
(Didn't know std::map::count existed, that'll be useful.)

PR #385, opened 2015-10-05
Reviewer AMC: SRV priority and Diameter routing

graemerobertson:
Hi Andy,

Please can you review these cpp-common changes to allow us to take into account SRV priority when routing Diameter messages?

The bulk of the work here is to refactor the existing code to make it more usable. Specifically, I've made RealmManager the sole manager of Diameter peers (removing the Diameter Stack's list of them), and I've tried to keep freeDiameter out of the RealmManager (instead using the Diameter Stack as the interface to freeDiameter), and I've made RealmManager an application of the Diameter Stack, so that the Stack doesn't know anything about RealmManager.

In addition to this, RealmManager now keeps track of the peers as a map rather than a vector, so that we can easily look up the Peer object when we get a call back. This is important because we now need to do this on the call path (i.e. when taking into account the SRV priority on message routing).

There is an accompanying Homestead PR (https://github.com/Metaswitch/homestead/pull/261) because the RealmManager is UTed in Homestead. I've still got complete code coverage on the RealmManager. I've also verified that existing function still works live, and I'm intending to run through the rest of the test plan later in the week.

Let me know if you've got any questions. It would be great if you could review by **end of Wednesday**, but let me know if that's going to be a problem.

Thanks,
Graeme

### DIFF
--- a/include/diameterstack.h
+++ b/include/diameterstack.h
@@ -55,6 +55,7 @@
 #include "snmp_counter_table.h"
 
 typedef std::function<void(bool, const std::string&, const std::string&)> PeerConnectionCB;
+typedef std::function<void(struct fd_list*)> RtOutCB;
 
 class RealmManager;
 
@@ -627,8 +628,12 @@ public:
 
   static inline Stack* get_instance() {return INSTANCE;};
   virtual void initialize();
-  virtual void register_peer_hook_hdlr(PeerConnectionCB peer_connection_cb);
-  virtual void unregister_peer_hook_hdlr();
+  virtual void register_peer_hook_hdlr(std::string listener_id,
+                                       PeerConnectionCB peer_connection_cb);
+  virtual void unregister_peer_hook_hdlr(std::string listener_id);
+  virtual void register_rt_out_cb(std::string listener_id,
+                                  RtOutCB peer_connection_cb);
+  virtual void unregister_rt_out_cb(std::string listener_id);
   virtual void configure(std::string filename,
                          ExceptionHandler* exception_handler,
                          BaseCommunicationMonitor* comm_monitor = NULL,
@@ -698,7 +703,11 @@ private:
 
   void fd_peer_hook_cb(enum fd_hook_type type, struct peer_hdr* peer);
   static void fd_peer_hook_cb(enum fd_hook_type type, struct msg* msg, struct peer_hdr* peer, void* other, struct fd_hook_permsgdata* pmd, void* stack_ptr);
-  std::vector<PeerConnectionCB*> _peer_connection_cbs;
+  std::map<std::string, PeerConnectionCB> _peer_connection_cbs;
+
+  void fd_rt_out_cb(struct fd_list* candidates);
+  static int fd_rt_out_cb(void* stack_ptr, struct msg** pmsg, struct fd_list* candidates);
+  std::map<std::string, RtOutCB> _rt_out_cbs;
 
   void fd_error_hook_cb(enum fd_hook_type type, struct msg* msg, struct peer_hdr* peer, void *other, struct fd_hook_permsgdata* pmd);
   static void fd_error_hook_cb(enum fd_hook_type type, struct msg* msg, struct peer_hdr* peer, void* other, struct fd_hook_permsgdata* pmd, void* stack_ptr);
@@ -716,6 +725,7 @@ private:
   struct disp_hdl* _callback_handler; /* Handler for requests callback */
   struct disp_hdl* _callback_fallback_handler; /* Handler for unexpected messages callback */
   struct fd_hook_hdl* _peer_cb_hdlr; /* Handler for the callback registered for connections to peers */
+  struct fd_rt_out_hdl* _rt_out_cb_hdlr; /* Handler for the callback registered for routing messages out */
   struct fd_hook_hdl* _error_cb_hdlr; /* Handler for the callback
                                        * registered for routing errors */
   struct fd_hook_hdl* _null_cb_hdlr; /* Handler for the NULL callback registered to overload the default hook handlers */
--- a/include/diameterstack.h
+++ b/include/diameterstack.h
@@ -48,19 +48,16 @@
 #include "utils.h"
 #include "sas.h"
 #include "baseresolver.h"
-#include "realmmanager.h"
 #include "communicationmonitor.h"
 #include "exception_handler.h"
 #include "counter.h"
 #include "snmp_counter_table.h"
 
+namespace Diameter
+{
 typedef std::function<void(bool, const std::string&, const std::string&)> PeerConnectionCB;
 typedef std::function<void(struct fd_list*)> RtOutCB;
 
-class RealmManager;
-
-namespace Diameter
-{
 class Stack;
 class Transaction;
 class AVP;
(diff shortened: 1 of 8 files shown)


include/diameterstack.h
```diff
@@ -48,18 +48,23 @@
...
 #include "sas.h"
 #include "baseresolver.h"
+#include "realmmanager.h"
 #include "communicationmonitor.h"
 #include "exception_handler.h"
 #include "counter.h"
 #include "snmp_counter_table.h"
 
+typedef std::function<void(bool, const std::string&, const std::string&)> PeerConnectionCB;
+typedef std::function<void(struct fd_list*)> RtOutCB;
```
bossmc: Should these be inside the `Diameter` namespace?  I think probably yes.

include/diameterstack.h
```diff
@@ -48,18 +48,23 @@
...
+
+class RealmManager;
```
bossmc: Don't need this any more?

include/diameterstack.h
```diff
@@ -48,18 +48,23 @@
 #include "utils.h"
 #include "sas.h"
 #include "baseresolver.h"
+#include "realmmanager.h"
```
bossmc: Not needed, right?

include/diameterstack.h
```diff
@@ -734,23 +736,24 @@ class Stack
...
   ExceptionHandler* _exception_handler;
   BaseCommunicationMonitor* _comm_monitor;
   SNMP::CounterTable* _realm_counter;
   SNMP::CounterTable* _host_counter;
 
-  // "Managed" peer count.  This is the number of peers as discovered by the
-  // upstream manager.  Most of the time it will be the same as the size of
-  // _peers, but will differ during management cycles as peers are added and
-  // removed and the managed value is more useful for reporting purposes.
+  // Number of peers the upstream RealmManager is currently either connected to
```
bossmc: Don't discuss `RealmManager`.

include/diameterstack.h
```diff
@@ -734,23 +736,24 @@ class Stack
...
+  // or is trying to connect to, or has recently tried to connect to, and the
+  // number of peers the RealmManager is actually connected to. Used for raising
+  // appropriate SAS logs when Diameter message routing fails and we have no
+  // connected peers.
   //
   // The constructor initialises it at -1 (not 0) to reflect the fact that the
-  // managed number of peers is unknown until the manager tells us, and in
-  // instances where there is no upstream manager, it will remain at -1
+  // these counts are unknown until the RealmManager tells us, and in
```
bossmc: "the these"

include/realmmanager.h
```diff
@@ -42,7 +42,13 @@
 #include "diameterstack.h"
 #include "diameterresolver.h"
 
-class RealmManager : public Diameter::PeerListener
+namespace Diameter
+{
+class Stack;
+class Peer;
+}
```
bossmc: Don't need the pre-declaration any more, since you've broken the cyclical dependencies.

src/diameterstack.cpp
```diff
@@ -124,28 +119,59 @@ void Stack::initialize()
...
-  int rc = fd_hook_register(HOOK_MASK(HOOK_PEER_CONNECT_SUCCESS,
-                                      HOOK_PEER_CONNECT_FAILED),
-                            fd_peer_hook_cb,
-                            this,
-                            NULL,
-                            &_peer_cb_hdlr);
-  if (rc != 0)
-  {
-    throw Exception("fd_hook_register", rc); // LCOV_EXCL_LINE
+  if (_peer_connection_cbs.empty())
```
bossmc: This needs a lock (as `std::map`s are not thread-safe).

src/diameterstack.cpp
```diff
@@ -124,28 +119,59 @@ void Stack::initialize()
...
+  if (_peer_cb_hdlr && _peer_connection_cbs.empty())
   {
     fd_hook_unregister(_peer_cb_hdlr);
   }
 }
 
+void Stack::register_rt_out_cb(std::string listener_id,
+                               RtOutCB rt_out_cb)
+{
+  if (_rt_out_cbs.empty())
```
bossmc: Ditto (needs a lock).  You can probably use a single lock for all of these register/unregister functions unless you think we'd call any of them in the golden path?

src/diameterstack.cpp
```diff
@@ -269,29 +295,19 @@ void Stack::fd_error_hook_cb(enum fd_hook_type type,
...
-  // since we take peers out of our list if they subsequently failed to connect
-  // or fail realm identification checks.
-  //
-  // We don't bother looking at peers for whom connected() returns "false" as
-  // these are peers that have recently been started, and for which
-  // initial connectivity is still ongoing (so they should be considered "good"
-  // peers until we determine otherwise).
-  //
-  // The SAS logs won't appear if hss_hostname has been configured (rather than
+  // We use counts given to us by the upstream RealmManager, if it exists. The
```
bossmc: RealmManager mention.

src/diameterstack.cpp
```diff
@@ -343,60 +356,36 @@ void Stack::fd_peer_hook_cb(enum fd_hook_type type,
...
-          }
-        }
-        break;
-      }
-    }
+    std::string host = peer->info.pi_diamid;
+    std::string realm = peer->info.runtime.pir_realm;
 
-    if (!peer_found)
+    for (std::map<std::string, PeerConnectionCB>::const_iterator cb = _peer_connection_cbs.begin();
```
bossmc: If only we could use `auto`...

src/diameterstack.cpp
```diff
@@ -583,6 +572,11 @@ void Stack::stop()
       fd_hook_unregister(_peer_cb_hdlr);
     }
 
+    if (_rt_out_cb_hdlr)
+    {
+      fd_rt_out_unregister(_rt_out_cb_hdlr, NULL);
```
bossmc: Does this leak the callback functions?  We probably don't care, since we're about to shut down, but it's not very tidy.

src/realmmanager.cpp
```diff
@@ -61,8 +62,19 @@ RealmManager::RealmManager(Diameter::Stack* stack,
...
+  _stack->register_peer_hook_hdlr("realmmanager",
+                                  std::bind(&RealmManager::peer_connection_cb,
+                                            this,
+                                            _1,
+                                            _2,
+                                            _3));
+  _stack->register_rt_out_cb("realmmanager",
+                             std::bind(&RealmManager::srv_priority_cb,
+                                       this,
+                                       _1));
```
bossmc: This code looks really smooth, we should consider this elsewhere (`Updater`?).

graemerobertson: No idea if you're a) joking and b) what 'Updater' is

src/diameterstack.cpp
graemerobertson: Say what?

src/realmmanager.cpp
```diff
@@ -254,30 +316,25 @@ void RealmManager::manage_connections(int& ttl)
...
-      if ((*jj)->host() == hostname)
-      {
-        found = true;
-        break;
-      }
+      found = true;
     }
 
     // If it isn't, add it.
     if (!found)
```
bossmc: Can't you merge this with the `if` statement above (and eliminate the `found` temporary variable)?  Or is `found` used later on?

src/diameterstack.cpp
graemerobertson: As in, we should clear the list (map) of callback functions...?

src/diameterstack.cpp
bossmc: `for (auto cb = _peer_connection_cbs.begin();`

src/diameterstack.cpp
bossmc: Or warn/assert if the maps aren't empty?

src/realmmanager.cpp
bossmc: a) no
b) https://github.com/Metaswitch/cpp-common/blob/master/include/updater.h

graemerobertson: Think I'm going to leave that for now if that's ok

bossmc:
Reviewed.

graemerobertson:
Andy, please can you review my markups (https://github.com/Metaswitch/cpp-common/commit/09979db166a595d86859d13663372fb0c3efd0f2), particularly the locking stuff.

Thanks!

PR #386, opened 2015-10-06
[Reviewer: Graeme] Catch exception if there's no keysapce set up

### LINKED ISSUE #382: If no cassandra schema is installed, cassandra_store fails to start with an unknown error
Surely there must be some way to detect what the problem is

eleanor-merry:
Graeme, can you review this change to catch an exception if the schema hasn't been set up, and put the reason in the trace, e.g.:
```
Error cassandra_store.cpp:189: Store caught InvalidRequestException: Keyspace 'homestead_cache' does not exist)
```

Fixes #382


graemerobertson:
Looks great!

PR #390, opened 2015-10-22
[Reviewer: Alex] Use TWO in reads

eleanor-merry:
Use TWO in reads rather than LOCAL_QUORUM/QUORUM. 

UT'd in Homestead (not yet tested as chef is playing up)

### DIFF
--- a/include/sasevent.h
+++ b/include/sasevent.h
@@ -139,6 +139,7 @@ namespace SASEvent {
   const int CASS_CONNECT_FAIL = COMMON_BASE + 0x0300;
 
   const int QUORUM_FAILURE = COMMON_BASE + 0x0400;
+  const int TWO_FAILURE = COMMON_BASE + 0x0401;
 
 } // namespace SASEvent
 
--- a/include/sasevent.h
+++ b/include/sasevent.h
@@ -139,7 +139,6 @@ namespace SASEvent {
   const int CASS_CONNECT_FAIL = COMMON_BASE + 0x0300;
 
   const int QUORUM_FAILURE = COMMON_BASE + 0x0400;
-  const int TWO_FAILURE = COMMON_BASE + 0x0401;
 
 } // namespace SASEvent
 
--- a/src/cassandra_store.cpp
+++ b/src/cassandra_store.cpp
@@ -676,23 +676,14 @@ enum class Quorum_Consistency_Levels
 // -  The arguments for the underlying get method.
 //
 // It works as follows:
-// -  Call the underlying method with a consistency level of LOCAL_QUORUM.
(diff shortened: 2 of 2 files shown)


src/cassandra_store.cpp
```diff
@@ -676,23 +676,14 @@ enum class Quorum_Consistency_Levels
 // -  The arguments for the underlying get method.
 //
 // It works as follows:
-// -  Call the underlying method with a consistency level of LOCAL_QUORUM.
+// -  Call the underlying method with a consistency level of TWO.
 //    If successful, this indicates that we've managed to find multiple nodes
 //    with the same data, and so we can trust the response (if we're wrong, it
 //    will be because several local nodes failed or were down simultaneously
```
AlexHockey: You could say "at least two" instead of several.

src/cassandra_store.cpp
```diff
@@ -676,23 +676,14 @@ enum class Quorum_Consistency_Levels
...
-//    LOCAL_QUORUM.  However, we cannot just drop straight through to the ONE
-//    level as this is the expected behaviour in a 2+2 GR system when we are
-//    restarting one of the nodes (e.g. thanks to an upgrade) and we can't
-//    run the risk that the ONE read might hit another recently restarted
-//    node that is still out of date.
-// -  If this *also* fails with UnavailableException, perform a ONE read.  In
-//    this case at least half of the servers in the cluster are down, and so
+//    service impact is acceptable).
+// -  If this fails with UnavailableException, perform a ONE read.  In
+//    this case at least half of the replicas in the cluster are down, and so
```
AlexHockey: Well, actually at most one of the replicas is still up.

AlexHockey:
Looks good, a couple of picky comments about commenting.

include/sasevent.h
```diff
@@ -139,6 +139,7 @@ namespace SASEvent {
   const int CASS_CONNECT_FAIL = COMMON_BASE + 0x0300;
 
   const int QUORUM_FAILURE = COMMON_BASE + 0x0400;
```
bossmc: This is no longer used (are you keeping it to block out the ID?)

AlexHockey: In fact the SAS event for this log is generic enough to use for the TWO read as well (assuming you extend the cassandra consistency level enum in clearwater-sas-resources)

eleanor-merry: I thought it was better to have a SAS log where you're not sending any extra parameters though? Happy to change it back.

AlexHockey: Static parameters are really cheap as it turns out, as are short var parameters. It's long var params that we need to be concerned about.

PR #391, opened 2015-10-26
Reviewer MIRW: Don't crash

graemerobertson:


src/snmp_infinite_timer_count_table.cpp
```diff
@@ -206,6 +197,16 @@ namespace SNMP
...
         }
 
+        new_oid_len += 2;
+        new_oid[ROOT_OID_LEN + 1 + tag.length()] = identifier.at(0);
+        new_oid[ROOT_OID_LEN + 2 + tag.length()] = identifier.at(1);
+
+        char buf1[64];
+        snprint_objid(buf1, sizeof(buf1),
+                      new_oid, new_oid_len);
+        TRC_DEBUG("Parsed SNMP request to OID %s with tag %s and identifier <%d, %d>",
```
mirw: Should we still trace something earlier?  Maybe there's not much we can trace?

mirw:
Looks good.  One comment inline.  Presumably you've tested both `snmpget` and `snmpgetnext` live?

graemerobertson:
@rkd-msw what do you think we should do with this? I'd like to merge it

rkd-msw:
Agreed.

PR #396, opened 2015-11-03
[Reviewer: Graeme] Avoid issues with high SRV priority values

rkd-msw:
Two fixes:

- sets default values for priority and weight - these will otherwise be uninitialized for A records
- stops SRV priorities from making a score negative

I haven't risked any of the multiplication stuff - I thought about it, but got nervous about possible integer overflow issues.

### DIFF
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -57,6 +57,10 @@ struct AddrInfo
   int priority;
   int weight;
 
+  AddrInfo():
+    priority(1),
+    weight(1) {};
+
   bool operator<(const AddrInfo& rhs) const
   {
     int addr_cmp = address.compare(rhs.address);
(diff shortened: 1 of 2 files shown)


include/baseresolver.h
```diff
@@ -57,6 +57,10 @@ struct AddrInfo
   int priority;
   int weight;
 
+  AddrInfo():
+    priority(1),
+    weight(1) {};
```
graemerobertson: Why not 0?

graemerobertson:
Yes, ok, looks fine.

Have you run the homestead UTs? I think you're going to break coverage (homestead hosts the realmmanager UTs).

I assume you're testing live too.

graemerobertson:
Also, this fixes https://github.com/Metaswitch/homestead/issues/274 right?

include/baseresolver.h
rkd-msw: We generally advise against SRV priority and weight 0, I think, so defaulting to 1 avoids some risk of bugs.

graemerobertson: Do we? Fine anyway, it doesn't really matter

Issue #384, opened 2015-09-23
Load Monitor keeps increasing token rate even if current rate is unproven


mirw:
Watching my (broadly idle) Sprout node, I see the token rate constantly increasing.  This means that, if I were to have a sudden spike in traffic, I would give poor behavior to a larger number of requests, while it took time for me to adjust my load back down to the correct number

The logic behind increasing the token rate is that the current latency is within bounds (and we haven't had any penalties rejected from upstream).  This is basically the algorithm given in the [Adaptive Overload Control for Busy Internet Servers](http://www.eecs.harvard.edu/~mdw/papers/control-usits03.pdf) paper.  Instead, I think we should only increase the rate if we're using some (large-ish) fraction of the current token rate - if we're not, there's no evidence that we can actually support an increased number of requests.

@eleanor-merry suggests this might be related to https://github.com/Metaswitch/sprout/issues/1090 - it's not clear, but worth bearing in mind.

mike-evans:
Sounds like a sensible solution.  I had a brief worry about start-up behaviour (ie. whether we would stabilise on a high enough token rate), but I think this should work fine.

PR #397, opened 2015-11-04
[Reviewer: Alex] Don't increase our token rate spuriously during slow periods

rkd-msw:
Fixes https://github.com/Metaswitch/cpp-common/issues/384 using the approach discussed there.

I've UTed, and I'll live test by running light stress with SIPp and checking that this doesn't cause stress to go badly wrong.

### DIFF
--- a/include/load_monitor.h
+++ b/include/load_monitor.h
@@ -100,7 +100,7 @@ class LoadMonitor
     int target_latency;
     int smoothed_latency;
     int adjust_count;
-    timespec last_adjustment_time;
+    unsigned long last_adjustment_time_ms;
     float min_token_rate;
     TokenBucket bucket;
     SNMP::ContinuousAccumulatorTable* _token_rate_table;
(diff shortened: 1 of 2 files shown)


AlexHockey:
The code looks good. 

I have one niggling worry about this. Clearwater's behavior when it's first started isn't great - if you apply load in a step-function we reject requests until the token refill rate catches up. Because of the bug you've just fixed you used to be able to work around this by letting your node run for a bit before applying load. Now that workaround doesn't work. Have we made our slow starting problem more visible?

rkd-msw:
We discussed: yes, this will make our slow starting problem more visible. The right approach here is:

- do some stress testing and figure out how many allowed requests we converge on on a single-core system
- set the initial rate value, as a Sprout command-line argument, to that figure times the number of cores (or maybe 75% of that value, to allow for perf fluctuations/different clock speeds)

I'll raise, or locate, a separate issue for that.

rkd-msw:
https://github.com/Metaswitch/sprout/issues/1251 is that issue.

rkd-msw:
I've decided against doing light stress testing before merging - the work we do to fix https://github.com/Metaswitch/sprout/issues/1251 will catch any issues here.

PR #398, opened 2015-11-06
[Reviewer: AMC] Send flush markers and suppress connection management logs

AlexHockey:
Andy, please can you review this change to:

* Send flush markers when a HTTP/Diameter task is completed. 
* Suppress SAS logging of diameter connection management flows.

I tested this by running the live tests and checking that they worked, and that my SAS lint reported no unflushed trails related to diameter/HTTP transactions.

### DIFF
--- a/include/httpstack_utils.h
+++ b/include/httpstack_utils.h
@@ -107,7 +107,13 @@ namespace HttpStackUtils
       _req(req), _trail(trail)
     {}
 
-    virtual ~Task() {}
+    virtual ~Task()
+    {
+      // Now the task is complete we should flush the trail to ensure it
+      // appears promptly in SAS.
+      SAS::Marker flush_marker(_trail, MARKED_ID_FLUSH);
+      SAS::report_marker(flush_marker);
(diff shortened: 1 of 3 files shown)


include/httpstack_utils.h
```diff
@@ -107,7 +107,13 @@ namespace HttpStackUtils
       _req(req), _trail(trail)
     {}
 
-    virtual ~Task() {}
+    virtual ~Task()
+    {
+      // Now the task is complete we should flush the trail to ensure it
+      // appears promptly in SAS.
+      SAS::Marker flush_marker(_trail, MARKED_ID_FLUSH);
```
bossmc: Is this safe even if `_trail` is 0?

bossmc:
Looks good.

include/httpstack_utils.h
AlexHockey: Yes, but the trail ID will never be zero (unless the user sets it to zero, which is a really weird thing to do).

PR #399, opened 2015-11-06
[Reviewer: Chris] Log routing errors at error level, not info

rkd-msw:
Should mean we spot issues like the one you hit a bit more quickly.


chris-elford-metaswitch:
Will this get spammy at load? Can you do anything to mitigate that? If not, I'm happy with that.

rkd-msw:
It feels like if you're in this situation (no HSS connections) at load, you've got bigger problems than log spam. We don't currently have first-and-interval logs, but if we implement them, this would be quite a good candidate.

PR #401, opened 2015-11-10
Allow testing of components that POST on an HttpConnection

dimbleby:


bossmc:
Did you see https://github.com/Metaswitch/cpp-common/blob/master/test_utils/mockhttpconnection.h?  Admittedly it seems we've not got a fully-functional test http stack, but I worry that a half fake/half mock one will be super confusing.

dimbleby:
I have an HttpConnection on which I both both GET and POST.  I guess I could instead extend the MockHttpConnection to support GETs.  Would you prefer that I do that?

dimbleby:
Actually I'm wrong, I only POST, so the Mock is fine and forget about it!

PR #402, opened 2015-11-10
Compiler fixes

dimbleby:
The second commit here claims that code oughtn't to compile, even though it obviously does.  But I think that the compiler is wrong!


rkd-msw:
We build all our UT code with `-fno-access-control` - that probably explains why it compiles unexpectedly.

bossmc:
But regardless your code looks fine.

PR #403, opened 2015-11-10
[Reviewer: Andy] Add safe_pidfile function to write a pidfile only if we have an excus…

rkd-msw:
…ive file lock

Basically https://github.com/Metaswitch/python-common/blob/dev/metaswitch/common/utils.py#L404 but in C++.

I should probably call it `lock_and_write_pidfile` for consistency, shouldn't I?


bossmc:
Yes, change the name.  Otherwise, fine.

PR #405, opened 2015-11-11
Add Utils::parse_http_url

mbt-msw:
Adds a utility function for parsing http URLs into their components (server name and path).

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -119,6 +119,8 @@ namespace Utils
 
   void hashToHex(unsigned char *hash_char, unsigned char *hex_char);
 
+  bool parse_http_url(const std::string& url, std::string& server, std::string& path);
+
   std::string url_unescape(const std::string& s);
   std::string url_escape(const std::string& s);
 
--- a/include/utils.h
+++ b/include/utils.h
@@ -119,6 +119,12 @@ namespace Utils
 
   void hashToHex(unsigned char *hash_char, unsigned char *hex_char);
 
+  /// Splits a URL of the form "http://<servername>[/<path>]" into
+  /// servername and path.
+  ///
+  /// Returns true iff the URL is in the corrects form, and sets the server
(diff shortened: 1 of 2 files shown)


src/utils.cpp
```diff
@@ -48,6 +48,35 @@
 
 #include "utils.h"
 
+bool Utils::parse_http_url(const std::string& url, std::string& server, std::string& path)
+{
+  size_t colon_pos = url.find(':');
+  if (colon_pos == std::string::npos || url.substr(0, colon_pos) != "http")
```
AlexHockey: It might only be a de facto coding standard, but we bracket sub-expressions.

src/utils.cpp
```diff
@@ -48,6 +48,35 @@
...
+  size_t slash_pos = url.find('/', slash_slash_pos + 2);
+  if (slash_pos == std::string::npos)
+  {
+    // No path.
+    server = url.substr(slash_slash_pos + 2);
+    path = "/";
+  }
+  else
+  {
+    server = url.substr(slash_slash_pos + 2, slash_pos - slash_slash_pos - 2);
```
AlexHockey: The last bit might be clearer as `- (slash_slash_pos + 2)`

include/utils.h
```diff
@@ -119,6 +119,8 @@ namespace Utils
 
   void hashToHex(unsigned char *hash_char, unsigned char *hex_char);
 
+  bool parse_http_url(const std::string& url, std::string& server, std::string& path);
```
AlexHockey: I realize that the other functions in this file aren't a model of good documentation, but could you add a comment explaining the interface. In particular that the user should check the return code is `true` before reading from the output variables, and that the path is defaulted to `/` if not present in the URL.

AlexHockey:
I have reviewed. A few stylistic comments, nothing major.

PR #407, opened 2015-11-12
[Reviewer: Andy] recv_file_descriptor return -1 on error

johadalin:
This is to fix issue #408 

This should stop it returning a 0 on 'error', which would happen, according to http://linux.die.net/man/2/recvmsg, if the peer has performed an orderly shutdown. This fixes an error where listing a sas in shared_config, but not restarting clearwater-infrastructure to add it to the whitelist, causes sprout to become unresponsive and fall into a restart loop, caused by receiving file descriptor '0' for a socket.

I am not sure what essentially copying https://github.com/Metaswitch/sas-client/blob/clearwater/source/sas.cpp#L403 across would do, as i'm not quite sure why the differences have evolved, or what ramifications changes would make, so this PR is just to fix the bug that was causing sprout to become unresponsive and be shut down by monit.

The code still builds, and a previously faulty deployment has been upgraded and now doesn't hit the restart loop, so it appears to be fixed. Let me know if you think we still need to update this function to be more like the old code.

### DIFF
--- a/src/namespace_hop.cpp
+++ b/src/namespace_hop.cpp
@@ -72,7 +72,7 @@ static int recv_file_descriptor(int socket)
   if (res <= 0)
   {
     TRC_WARNING("Failed to retrieve cross-namespace socket - recvmsg returned %d (%d %s)\n", res, errno, strerror(errno));
-    return res;
+    return -1;
   }
 
   // Iterate through control message to find if there is a SCM_RIGHTS entry containing file descriptors
--- a/src/namespace_hop.cpp
+++ b/src/namespace_hop.cpp
@@ -53,7 +53,7 @@ static int recv_file_descriptor(int socket)
   // Buffer to hold the dummy "actual data" we receive
   char data[1] = {0};
 
-  struct iovec iov[1];
+  struct iovec iov[1] = {0};
   iov[0].iov_base = data;
   iov[0].iov_len = sizeof(data);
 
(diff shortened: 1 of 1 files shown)


bossmc:
Hmmmm... Beyond the better commenting of the `namespace_hop` version, I prefer almost everything about the `sas-client` implementation, so it would be good to move the code over (and merge in the comments).

Your specific fix looks fine.

johadalin:
Do you want any more than this pulled across?

src/namespace_hop.cpp
```diff
@@ -52,11 +52,12 @@ static int recv_file_descriptor(int socket)
 {
   // Buffer to hold the dummy "actual data" we receive
   char data[1] = {0};
-  struct iovec iov = {.iov_base = data,
-                      .iov_len = sizeof(data)};
+
+  struct iovec iov[1];
```
bossmc: This needs the `= {0}` (since you're not `memset`-ing to 0).

src/namespace_hop.cpp
```diff
@@ -65,17 +66,19 @@ static int recv_file_descriptor(int socket)
...
   if (res <= 0)
   {
     TRC_WARNING("Failed to retrieve cross-namespace socket - recvmsg returned %d (%d %s)\n", res, errno, strerror(errno));
-    return res;
+    return -1;
   }
 
   // Iterate through control message to find if there is a SCM_RIGHTS entry containing file descriptors
+  struct cmsghdr *control_message = NULL;
+
```
bossmc: Extra blank line.

bossmc:
Two comments, otherwise looks good.

PR #409, opened 2015-11-13
[Reviewer: Alex] Allow construction of a localhost-only memcached

rkd-msw:
This adds a new MemcachedStore constructor, which just directs all traffic to 127.0.0.1:11211.

I've left all the rest of the code in although Sprout/Ralf/etc. no longer need much of it - this is because:

* you're going to need it for Astaire
* the more of this we rewrite, the more painful merges will be
* some of it is useful - e.g. trying twice to workaround a memcached connection reuse bug
* my use-case is genuinely a special case of the general use-case - I need to connect to one server with one vbucket and one replica

### DIFF
--- a/src/memcachedstore.cpp
+++ b/src/memcachedstore.cpp
@@ -60,20 +60,22 @@
 /// The data used in memcached to represent a tombstone.
 static const std::string TOMBSTONE = "";
 
-BaseMemcachedStore::BaseMemcachedStore(bool binary,
+BaseMemcachedStore::BaseMemcachedStore(int replicas,
+                                       int vbuckets,
+                                       bool binary,
                                        MemcachedConfigReader* config_reader,
                                        BaseCommunicationMonitor* comm_monitor,
                                        Alarm* vbucket_alarm) :
-  _binary(binary),
-  _updater(NULL),
-  _replicas(2),
-  _vbuckets(128),
-  _options(),
   _view_number(0),
   _servers(),
-  _max_connect_latency_ms(50),
   _read_replicas(_vbuckets),
   _write_replicas(_vbuckets),
+  _binary(binary),
+  _updater(NULL),
+  _replicas(replicas),
+  _vbuckets(vbuckets),
+  _options(),
+  _max_connect_latency_ms(50),
   _comm_monitor(comm_monitor),
   _vbucket_comm_state(_vbuckets),
   _vbucket_comm_fail_count(0),
@@ -97,8 +99,11 @@ BaseMemcachedStore::BaseMemcachedStore(bool binary,
   _options = "--CONNECT-TIMEOUT=10 --SUPPORT-CAS --POLL-TIMEOUT=250";
   _options += (_binary) ? " --BINARY-PROTOCOL" : "";
 
-  // Create an updater to keep the store configured appropriately.
-  _updater = new Updater<void, BaseMemcachedStore>(this, std::mem_fun(&BaseMemcachedStore::update_config));
+  if (config_reader != NULL)
+  {
+    // Create an updater to keep the store configured appropriately.
+    _updater = new Updater<void, BaseMemcachedStore>(this, std::mem_fun(&BaseMemcachedStore::update_config));
+  }
 
   // Initialize vbucket comm state
   for (int ii = 0; ii < _vbuckets; ++ii)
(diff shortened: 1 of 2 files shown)


rkd-msw:
By the way, I think I can merge this to master rather than keeping a long-running branch - shout if you disagree.

src/memcachedstore.cpp
```diff
@@ -107,6 +107,44 @@ BaseMemcachedStore::BaseMemcachedStore(bool binary,
...
+
+  // Create the mutex for protecting vbucket comm state.
+  pthread_mutex_init(&_vbucket_comm_lock, NULL);
+
+  // Set up the fixed options for memcached.  We use a very short connect
+  // timeout because libmemcached tries to connect to all servers sequentially
+  // during start-up, and if any are not up we don't want to wait for any
+  // significant length of time.
+  _options = "--CONNECT-TIMEOUT=10 --SUPPORT-CAS --POLL-TIMEOUT=250";
+  _options += (_binary) ? " --BINARY-PROTOCOL" : "";
```
AlexHockey: You initialize the `_vbucket_comm_lock` but not `_vbucket_comm_state`. Is this right?

rkd-msw: Yep - we'll never touch `_vbucket_comm_state` if there's no vbucket alarm (and we don't want one here), but we'll destroy the lock in the destructor. I could make that conditional on the alarm existing, but it doesn't feel very valuable.

AlexHockey: Quite a lot of code duplication here. Subfunction?

AlexHockey:
Hmm.... I'm not a big fan of this change (sorry!). Some of the members are only initialized if the constructor was called in a certain way are are irrelevant otherwise. I think a better approach would be to define two subclasses of `BaseMemcachedStore` and move the functionality into whichever class actually needs it.

src/memcachedstore.cpp
```diff
@@ -107,6 +107,44 @@ BaseMemcachedStore::BaseMemcachedStore(bool binary,
...
+}
+
```
AlexHockey: Quite a lot of code duplication here. Subfunction?

rkd-msw:
We agreed (after toying briefly with OmniscientMemcachedStore and MonoscientMemcachedStore) on Topology{Aware,Neutral}MemcachedStore.

src/memcachedstore.cpp
AlexHockey: If we're initializing things we never touch lets initialize everything. There is less chance of catching someone out in the future.

rkd-msw:
Better?

src/memcachedstore.cpp
```diff
@@ -1021,24 +1025,48 @@ void BaseMemcachedStore::log_delete_failure(const std::string& fqkey,
...
+{
+  set_fixed_server("127.0.0.1:11211");
+}
+
+void TopologyNeutralMemcachedStore::set_fixed_server(std::string server)
+{
+  _view_number++;
+  _servers = {server};
+  _read_replicas = {{server}};
+  _write_replicas = {{server}};
```
AlexHockey: Wow, I had no idea you could do that!

src/memcachedstore.cpp
```diff
@@ -97,8 +99,11 @@ BaseMemcachedStore::BaseMemcachedStore(bool binary,
   _options = "--CONNECT-TIMEOUT=10 --SUPPORT-CAS --POLL-TIMEOUT=250";
   _options += (_binary) ? " --BINARY-PROTOCOL" : "";
 
-  // Create an updater to keep the store configured appropriately.
-  _updater = new Updater<void, BaseMemcachedStore>(this, std::mem_fun(&BaseMemcachedStore::update_config));
+  if (config_reader != NULL)
+  {
+    // Create an updater to keep the store configured appropriately.
+    _updater = new Updater<void, BaseMemcachedStore>(this, std::mem_fun(&BaseMemcachedStore::update_config));
```
AlexHockey: Should the `_updater` live in the base class. Should it?

AlexHockey:
Looks good, just one comment.

src/memcachedstore.cpp
AlexHockey: Should the `_updater` live in `TopologyAwareMemcachedStore`? Does this even work?

src/memcachedstore.cpp
rkd-msw: Yep, C++11 brace initialization.

src/memcachedstore.cpp
rkd-msw: Yes, it does. See the new commit here - note that I want to talk to you about tombstone_lifetime before merging.

rkd-msw: We agreed that my change to hardcode tombstone_lifetime is fine.

PR #411, opened 2015-11-16
[Reviewer: Rob] Add Chronos mock that doesn't just accept everything

eleanor-merry:
Add a new ChronosConnection Mock

### DIFF
--- a/test_utils/mock_chronos_connection.cpp
+++ b/test_utils/mock_chronos_connection.cpp
@@ -1,5 +1,5 @@
 /**
- * @file mock_chronos_connection.h
+ * @file mock_chronos_connection.cpp
  *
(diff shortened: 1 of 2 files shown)


rkd-msw:
I'm not sure I like the calling convention here - I don't think people are going to remember that:

`MockChronosConnection* m = new MockChronosConnection("chronos");`

accepts everything but

`MockChronosConnection* m = new MockChronosConnection();`

doesn't. Could we either have factory methods to make it obvious, or use VerifyAndClearExpectations?

i.e. 

`MockChronosConnection* m = MockChronosConnection::without_defaults();`

or

```
MockChronosConnection* m = new MockChronosConnection("chronos");
m->clear();
```

PR #412, opened 2015-11-17
[Reviewer: Rob] This includes the ability to simulate data contention by creating an intentionally out-of-date database which is read from if a data contention flag is activated.

AME2:


src/localstore.cpp
```diff
@@ -78,7 +84,13 @@ Store::Status LocalStore::get_data(const std::string& table,
                                    SAS::TrailId trail)
 {
   TRC_DEBUG("get_data table=%s key=%s", table.c_str(), key.c_str());
-
+  std::map<std::string, Record> _db_in_use;
```
rkd-msw: This code should come after the `pthread_mutex_lock(&_db_lock);` - otherwise you run the risk that someone else is adding a record to `_db` at the same time as you're copying it into `db_in_use`.

Relatedly, this copies the contents of `_db` or `_old_db` into `_db_in_use` - if you declare `_db_in_use` as `std::map<std::string, Record>& _db_in_use` (note the extra ampersand in the type), then it's an alias rather than a copy, which is more efficient. (http://www.cprogramming.com/tutorial/references.html explains this in more detail.)

src/localstore.cpp
```diff
@@ -157,6 +169,7 @@ Store::Status LocalStore::set_data(const std::string& table,
       // Supplied CAS is consistent (either because record hasn't expired and
       // CAS matches, or record has expired and CAS is zero) so update the
       // record.
+      _old_db[fqkey] = r;
```
rkd-msw: Can you comment to explain that we're saving off the previous value, so that we can simulate data contention later?

src/localstore.cpp
```diff
@@ -70,6 +72,10 @@ void LocalStore::flush_all()
   pthread_mutex_unlock(&_db_lock);
 }
 
+void LocalStore::force_contention()
```
rkd-msw: It's helpful to future users if you add a comment to explain what this function does - e.g. that this makes the next `get_data` call return out-of-date data, so that a later attempt to store that data fails with contention.

rkd-msw:
Looks good, with a couple of comments inline.

The pull request subject is "This fixes silly bug of not renaming _db to _db_in_use in get_data()", but I think this pull request is a bit broader than that - it adds a force_contention() method to LocalStore so we can test data contention. It's worth renaming the pull request - this makes it easy to browse https://github.com/Metaswitch/cpp-common/pulls?q=is%3Apr+is%3Aclosed in future and see at a glance which changes went in and when.

src/localstore.cpp
```diff
@@ -157,6 +177,11 @@ Store::Status LocalStore::set_data(const std::string& table,
       // Supplied CAS is consistent (either because record hasn't expired and
       // CAS matches, or record has expired and CAS is zero) so update the
       // record.
+      
+      // This write data this is one update out-of-date to _old_db this is for
```
rkd-msw: Typo: I think this should be "This writes data which is one update out-of-date..." or similar.

src/localstore.cpp
```diff
@@ -78,20 +88,30 @@ Store::Status LocalStore::get_data(const std::string& table,
                                    SAS::TrailId trail)
 {
   TRC_DEBUG("get_data table=%s key=%s", table.c_str(), key.c_str());
-
-  Store::Status status = Store::Status::NOT_FOUND;
+    Store::Status status = Store::Status::NOT_FOUND;
```
rkd-msw: Change of indentation - looks accidental?

PR #413, opened 2015-11-18
[Reviewer: Andy] Scalar infinite table

johadalin:
Tested on a single node system and all looks good. 
Going to test further on multi-node system, but i don't think my code will had any effect on that, so should all be good. Would be good if you can check i haven't done anything stupid in the structure of header files.

Other than adding "InfiniteBaseTable::" to function names, i don't believe i've touched anything other than: increment decrement functionality, get_value function, max_row and max_column, and the maps the data is stored in in the child classes.

### DIFF
--- a/include/snmp_infinite_base_table.h
+++ b/include/snmp_infinite_base_table.h
@@ -52,7 +52,7 @@ namespace SNMP
 
 class InfiniteBaseTable
 {
-public:
+protected:
   InfiniteBaseTable(std::string name,
                     std::string oid,
                     uint32_t max_row,
@@ -65,7 +65,6 @@ private:
   static const uint32_t MAX_TAG_LEN = 16;
   static const ssize_t SCRATCH_BUF_LEN = 128;
 
-protected:
   std::string _name;
   oid _tbl_oid[SCRATCH_BUF_LEN];
   size_t _tbl_oid_len;
@@ -73,8 +72,6 @@ protected:
   const uint32_t _max_column;
   netsnmp_handler_registration* _handler_reg;
   uint32_t ROOT_OID_LEN;
-
-private:
 
   static int static_netsnmp_table_handler_fn(netsnmp_mib_handler *handler,
                                              netsnmp_handler_registration *reginfo,
@@ -100,6 +97,7 @@ private:
                      std::unique_ptr<oid[]>& new_oid,
                      uint32_t& new_oid_len);
 };
+
 }
 
 #endif
--- a/include/snmp_infinite_base_table.h
+++ b/include/snmp_infinite_base_table.h
@@ -1,0 +1,60 @@
+/**
+ * @file snmp_infinite_base_table.h
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2015 Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version, along with the "Special Exception" for use of
+ * the program along with SSL, set forth below. This program is distributed
+ * in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details. You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ * The author can be reached by email at clearwater@metaswitch.com or by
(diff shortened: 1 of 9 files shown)


include/snmp_infinite_base_table.h
```diff
@@ -0,0 +1,105 @@
...
+  static const ssize_t SCRATCH_BUF_LEN = 128;
+
+protected:
+  std::string _name;
+  oid _tbl_oid[SCRATCH_BUF_LEN];
+  size_t _tbl_oid_len;
+  const uint32_t _max_row;
+  const uint32_t _max_column;
+  netsnmp_handler_registration* _handler_reg;
+  uint32_t ROOT_OID_LEN;
```
bossmc: Do all of these need to be `protected`?  I doubt the child class needs to see many if any of them.

include/snmp_infinite_base_table.h
```diff
@@ -0,0 +1,105 @@
...
+class InfiniteBaseTable
+{
+public:
+  InfiniteBaseTable(std::string name,
+                    std::string oid,
+                    uint32_t max_row,
+                    uint32_t max_column);
+
+  virtual ~InfiniteBaseTable();
+  virtual Value get_value(std::string, uint32_t, uint32_t, timespec) = 0;
```
bossmc: Why is this `public`?  Surely it shouldn't be called by an external class?

include/snmp_infinite_base_table.h
```diff
@@ -0,0 +1,105 @@
...
+                 std::string& tag,
+                 uint32_t& row,
+                 uint32_t& column);
+
+  void find_next_oid(const oid* req_oid,
+                     const uint32_t& req_oid_len,
+                     std::unique_ptr<oid[]>& new_oid,
+                     uint32_t& new_oid_len);
+};
+}
```
bossmc: I'd put a blank line between these `}`s, for symmetry with how the scopes were opened (though I don't know if that's consistent with other places we use namespaces)...

include/snmp_infinite_scalar_table.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "logger.h"
+
+#ifndef SNMP_INFINITE_SCALAR_TABLE_H
+#define SNMP_INFINITE_SCALAR_TABLE_H
+
+namespace SNMP
+{
+
+class InfiniteScalarTable
```
bossmc: As previously discussed this is a terrible name... though the refactor might not be worth the effort.  I'd prefer something like `InstantaneousTimerCountsTable` or similar (same comment applies to the other classes).  The `base` table is the one table where the name is defensible, but I'd still prefer to see `TagIndexedMetaTableBase` or something so it's clear what function the class provides.

src/snmp_infinite_scalar_table.cpp
```diff
@@ -0,0 +1,111 @@
...
+  public:
+    InfiniteScalarTableImpl(std::string name, // Name of this table, for logging
+                                std::string tbl_oid):// Root OID of this table
+    InfiniteBaseTable(name, tbl_oid, max_row, max_column){}
+
+    virtual ~InfiniteScalarTableImpl(){};
+    
+    void increment(std::string tag)
+    {
+      _scalar_counters[tag]++;
```
bossmc: This doesn't feel thread-safe... is this an atomic integer?

bossmc: Ditto elsewhere.

src/snmp_infinite_scalar_table.cpp
```diff
@@ -0,0 +1,111 @@
...
+      if (_scalar_counters[tag]>0)
+        _scalar_counters[tag]--;
+      else
+        _scalar_counters[tag] = 0;
+    }
+
+  protected:
+    static const uint32_t max_row = 1;
+    static const uint32_t max_column = 2;
+    std::map<std::string, int> _scalar_counters;
```
bossmc: Nope, not atomic...

src/snmp_infinite_timer_count_table.cpp
```diff
@@ -90,453 +68,34 @@ namespace SNMP
...
-      char buf[SCRATCH_BUF_LEN];
-      char buf2[SCRATCH_BUF_LEN];
-      snprint_objid(buf, sizeof(buf), req_oid, req_oid_len);
-      snprint_objid(buf2, sizeof(buf2), new_oid.get(), new_oid_len);
-      TRC_DEBUG("Found next OID, %s -> %s", buf, buf2);
-      return;
+      SimpleStatistics stats;
+      Value result = Value::uint(0);
+      // Update and obtain the relevants statistics structure
+      _timer_counters[tag].get_statistics(row, now, &stats);
```
bossmc: Same thread-safety question.

bossmc:
Reviewed

PR #414, opened 2015-11-19
[Reviewer: Rob] Astaire uses port 11311

AlexHockey:
Change the `TopologyNeutralMemcachedStore` accordingly.


rkd-msw:
Perfection itself.

PR #417, opened 2015-12-04
[Reviewer: Andy] Catch a thrift timeout exception and retry when doing a HA get

AlexHockey:
Andy, as discussed yesterday, our HA get code had a bug where we didn't trap timeout exception properly. Tested in UT and live. 

Not the prettiest of code but I didn't think it was worth the effort to tidy it up. Shout if you disagree.


bossmc:
This looks good (regardless of the copy-pasta)

PR #418, opened 2015-12-05
Reviewer Rob - HTTPStack registers handlers by const string

bossmc:
All our builds have `-fno-write-strings` enabled because this function forgot to put `const` on one of it's arguments.  Disabling compiler checks makes me a sad :panda_face: so here's the fix.


bossmc:
Actually, this is blocking me so I'm merging now, shout if you had any comments and I'll post-apply.

rkd-msw:
+1

PR #423, opened 2015-12-07
[Reviewer: Andy] Divide nanoseconds by 1,000,000 to get milliseconds

rkd-msw:
I also updated a log while debugging this, to say "handled 3 requests in last 2006ms" instead of "handled 3 requests recently" - that's a change worth keeping.


bossmc:
Nice, ship it.

PR #424, opened 2015-12-08
[Reviewer: Andy] Skip the infinite table if we come in too early

rkd-msw:
This resolves https://github.com/Metaswitch/clearwater-infrastructure/issues/259. I've chosen to just skip the infinite table if we come in before the first element:

* this is a lot safer and more maintainable than reworking netsnmp's inclusion/exclusion code
* the current behaviour isn't that unreasonable except when faced with an infinite table
* there's no valid use-case for walking the whole infinite table


bossmc:
Fine (at least the location of the change was easy to find :laughing: )

PR #431, opened 2015-12-10
[Reviewer AJH] Update to resource bundle for fix to SFR477136.

chris-elford-metaswitch:
SAS bug has been fixed so we can use a new resource bundle.

Alex - here is the other half of the fix.


AlexHockey:
Fine.

PR #432, opened 2015-12-14
[Reviewer: Andy] Allow customising the UDP port for DNS

rkd-msw:
Needed for clearwater-fv-test, so we can run dnsmasq on an unprivileged port then talk to it.

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -173,18 +173,21 @@ void DnsCachedResolver::init_from_server_ips(const std::vector<std::string>& dns
 
 
 DnsCachedResolver::DnsCachedResolver(const std::vector<IP46Address>& dns_servers) :
+  _port(53),
   _cache()
 {
   init(dns_servers);
 }
 
 DnsCachedResolver::DnsCachedResolver(const std::vector<std::string>& dns_servers) :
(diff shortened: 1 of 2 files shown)


src/dnscachedresolver.cpp
```diff
@@ -173,18 +173,21 @@ void DnsCachedResolver::init_from_server_ips(const std::vector<std::string>& dns
 
 
 DnsCachedResolver::DnsCachedResolver(const std::vector<IP46Address>& dns_servers) :
+  _port(53),
```
bossmc: This feels bit sub-optimal (you can't have different ports across your collection of DNS servers), do we have (Address + Port) objects we can pass around?

bossmc:
Looks reasonable, one "completeness" question inline.

src/dnscachedresolver.cpp
rkd-msw: Unfortunately this reflects the c-ares API - you can only set one UDP port, not one per server. (See https://github.com/bagder/c-ares/blob/ce3a0ee056d0f52c904dfe27784809903648acc0/ares.h#L556 for the TODO about this.)

As we're only setting ports for FV test usage, not exposing the option to do so in production, I'm not too worried about this.

PR #433, opened 2015-12-17
[Reviewer: Andy] Memcached DNS lookups and connection pooling

AlexHockey:
Please can you review this PR that adds DNS resolution and connection pooling to the topology-neutral memcached store? This should match the design you've already reviewed. I've tested this in the clearwater-fv-test PR I've also sent you.


include/astaire_resolver.h
```diff
@@ -0,0 +1,81 @@
...
+               SAS::TrailId trail);
+
+  /// Default duration to blacklist hosts after we fail to connect to them.
+  static const int DEFAULT_BLACKLIST_DURATION = 30;
+
+private:
+  int _address_family;
+};
+
+#endif
```
bossmc: Does this belong in `cpp-common` or in `astaire` itself?  The name suggests it's pretty Astaire specific.

include/memcachedstore.h
```diff
@@ -241,91 +270,160 @@ class BaseMemcachedStore : public Store
...
+  ///                        specified domain.
   /// @param comm_monitor  - Object tracking memcached communications.
-  /// @param vbucket_alarm - Alarm object to kick if a vbucket is
-  ///                        uncontactable.
-  TopologyAwareMemcachedStore(bool binary,
-                              MemcachedConfigReader* config_reader,
-                              BaseCommunicationMonitor* comm_monitor = NULL,
-                              Alarm* vbucket_alarm = NULL);
+  TopologyNeutralMemcachedStore(const std::string& target_domain,
+                                AstaireResolver* resolver,
```
bossmc: Ah I see now why AstaireResolver is in CPP-common.  Isn't it a subclass of BaseResolver or something that we can reference here?

include/memcachedstore.h
```diff
@@ -241,91 +270,160 @@ class BaseMemcachedStore : public Store
...
+    // Destructor.
+    ~ConnectionHandle();
+
+    // Get the underlying connection.
+    Connection* get();
+
+  private:
+    // Delete the copy constructor to avoid having copies of the handle (which
+    // could cause the connection to get checked back in twice!)
+    ConnectionHandle(ConnectionHandle& rhs) = delete;
```
bossmc: And thus rust was invented...

include/memcachedstore.h
```diff
@@ -241,91 +270,160 @@ class BaseMemcachedStore : public Store
...
+  typedef std::deque<Connection*> ConnectionPoolSlot;
+  typedef std::map<AddrInfo, ConnectionPoolSlot> ConnectionPool;
+  ConnectionPool _conn_pool;
+  pthread_mutex_t _conn_pool_lock;
+
+  // Get a connection to the specified target.
+  ConnectionHandle get_connection(AddrInfo& target);
+
+  // Release the specified connection back to the pool.
+  void release_connection(Connection* conn);
```
bossmc: I'm surprised that this doesn't take a `ConnectionHandle`?  For symmetry with `get_connection()` is this because handles use RIAA?

src/memcachedstore.cpp
```diff
@@ -150,14 +346,16 @@ BaseMemcachedStore::get_replicas(const std::string& key, Op operation)
...
 /// Gets the set of replicas to use for a read or write operation for the
 /// specified vbucket.
-const std::vector<memcached_st*>& BaseMemcachedStore::get_replicas(int vbucket,
-                                                                   Op operation)
+const std::vector<memcached_st*>& TopologyAwareMemcachedStore::get_replicas(int vbucket,
+                                                                            Op operation)
 {
-  BaseMemcachedStore::connection* conn = (connection*)pthread_getspecific(_thread_local);
+  TopologyAwareMemcachedStore::connection* conn =
+    (connection*)pthread_getspecific(_thread_local);
```
bossmc: It's seems weird to fully specify `BaseMemcachedStore::connection` in one place and not in the cast.

Also, `connection` rather than `Connection`?

src/memcachedstore.cpp
```diff
@@ -941,123 +1026,642 @@ void BaseMemcachedStore::delete_with_tombstone(const std::string& fqkey,
...
+  TRC_DEBUG("Release connection to IP: %s, port: %d",
+            conn->target.address.to_string().c_str(),
+            conn->target.port);
+
+  // Get the current time, and mark the connection as having just been used.
+  struct timespec now;
+  clock_gettime(CLOCK_REALTIME, &now);
+  conn->last_used_time_s = now.tv_sec;
+
+  pthread_mutex_lock(&_conn_pool_lock);
```
bossmc: Asymmetric locking!  You lock here and unlock in `free_old_connection` but I think there's no good reason not to lock and unlock here (and `free_old_connection` can require the lock be held).

bossmc:
Well, that was fun, my top-level comment is that all the operations that can be done follow a specific pattern (calculate qualified key, get targets, iterate over targets (bail out if you get a non-retryable `rc`) doing the operation, then (maybe) trickle that change to the other targets) and it would be nice if that was commonized.  You suggested originally that the target selection was intertwined with the retry-loop but the final code is pretty cleanly separated.

src/memcachedstore.cpp
AlexHockey: This is the topology-aware store (the one that connects directly to memcached). All I've done to this is move methods between classes, and I'm not planning on making any unnecessary changes to it.

src/memcachedstore.cpp
AlexHockey: Good spot! `release_connection` and `free_old_connection` used to be one method. But I found a bug so I split them in two (as splitting them in the first place would have prevented the bug), but introduced this new bug!

include/memcachedstore.h
AlexHockey: `BaseResolver` is not an interface, its a bunch of "useful DNS-related functions" that it expects a subclass to assemble into an actually useful implementation. So I need to pass in an actual `AstaireResolver` here.

include/astaire_resolver.h
AlexHockey: It's not a great name, but I couldn't think of a better one. `MemcachedResolver` doesn't feel right as it doesn't apply to the topology-aware store (and the port it uses is not the memcached port), and `TopologyNeutralMemcachedResolver` was a bit of  a mouthful.

include/memcachedstore.h
AlexHockey: Tell me about it. Whenever I write an RAII guard I wish move-semantics were the default in C++. Oh well.

include/memcachedstore.h
AlexHockey: Yes. I will beef up the comment sand rename the method to `_release_connection` to make it clearer that it shouldn't be called willy nilly.

AlexHockey:
Thanks Andy! 

I've had a think about code duplication. At present there are 6 different memcached operations: (set, get, delete) x (topology-aware-store, topology-neutral-store). 

There actually that much commonality between different methods in the topology-neutral store as they all drop different SAS logs and handle error codes in different ways. Some of the common processing has already been refactored out (e.g. target selection) or isn't worth refactoring (e.g. calculating the fully-qualified key). I agree that code that loops over possible targets is duplicated, but refactoring out flow control (for loops, break statements) is tricky without using something like a functor or lambda. I'll have a think about how I might be able to improve this, but I'm not making any promises ;-).

There **is** quite a lot of commonality between the same method on different stores. However I am wary about refactoring this out because:

* We don't have any regression tests for the topology-aware store, so refactoring risks regressions. 
* Depending on how your upgrade discussions go, the topology-aware store may be dead after this release, in which case we shouldn't waste time refactoring it. 

There is also similarity between the `TopologyAwareMemcachedStore` store and the `MemcachedBackend` used by Astaire. Again, we shouldn't do anything on this until we know if the former is going to last beyond the next release.

AlexHockey:
I've applied markups for your specific feedback. I'm going to submit any further refactoring as a new PR.

PR #434, opened 2015-12-18
[Reviewer: Andy] Refactor duplicated code out of the topology-neutral store

AlexHockey:
This PR refactors out some duplicated code from the topology-neutral store:

* Moves construction of the fully-qualified key into one place. 
* Commonalizes the loop-over-targets code into one place, but with the actual communication to memcached done from within a lambda. 

Thoughts?

Tested in the FV tests.

### DIFF
--- a/include/memcachedstore.h
+++ b/include/memcachedstore.h
@@ -118,6 +118,13 @@ protected:
                                                uint32_t flags,
                                                SAS::TrailId trail);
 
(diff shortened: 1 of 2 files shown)


bossmc:
Looks good to me, maybe the `iterate_over_targets` function could have a function comment explaining how it's used (and referencing the three users as examples of using a lambda as the `std::function`)

PR #436, opened 2015-12-23
[Reviewer: Rob] iss214: Adds comment explaning alarm state->severity mapping

AME2:
Fixes https://github.com/Metaswitch/clearwater-etcd/issues/214

### DIFF
--- a/src/json_alarms.cpp
+++ b/src/json_alarms.cpp
@@ -136,7 +136,6 @@ namespace JSONAlarms
           std::string description;
 
           JSON_GET_STRING_MEMBER(*alarms_def_it, "severity", severity);
-          // Here we are reading severities from JSON files. 
           // Alarms are stored in ITU Alarm Table using the severities below.
           // Alarm Model Table stores alarms according to their state. The
           // mapping between state and severity is described in RFC 3877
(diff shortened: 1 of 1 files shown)


src/json_alarms.cpp
```diff
@@ -136,6 +136,12 @@ namespace JSONAlarms
           std::string description;
 
           JSON_GET_STRING_MEMBER(*alarms_def_it, "severity", severity);
+          // Here we are reading severities from JSON files.
```
rkd-msw: I'm not sure we need to explain this - the line above, `JSON_GET_STRING_MEMBER(*alarms_def_it, "severity", severity)` seems to be doing this pretty clearly.

src/json_alarms.cpp
```diff
@@ -136,6 +136,12 @@ namespace JSONAlarms
           std::string description;
 
           JSON_GET_STRING_MEMBER(*alarms_def_it, "severity", severity);
+          // Here we are reading severities from JSON files. 
+          // Alarms are stored in ITU Alarm Table using the severities below.
```
rkd-msw: I think it's good to use the actual terms alarmModelState and ituAlarmPerceivedSeverity - you've done this elsewhere (e.g. https://github.com/Metaswitch/clearwater-infrastructure/pull/281/files), and this is more useful, because it gives some obvious terms that the reader can search on if they go to https://tools.ietf.org/html/rfc3877#section-5.4.

rkd-msw:
Couple of comments, but fine to merge once you've addressed them.

PR #437, opened 2015-12-23
[Reviewer AMC] Don't make ENT logs for client side HTTP errors

steelyvoid:
The problem addressed here is that, if a client attempts to do something that results in a client-side error at the sprout->homestead interface (e.g., "404 not found" from a reg-data or location query), a "Request for <URL> to HTTP server A.B.C.D failed with error <client side error>" ENT log is made.

This is wrong - ENT logs are intended to record server side problems.  Client side problems are recorded and diagnosed using SAS.

Tested by checking that REGISTERs as unknown users no longer generate this ENT log (but REGISTERs when homestead is down continue to do so)

### DIFF
--- a/src/httpconnection.cpp
+++ b/src/httpconnection.cpp
@@ -661,8 +661,10 @@ HTTPCode HttpConnection::send_request(const std::string& path,                 /
       {
         CL_HTTP_COMM_ERR.log(url.c_str(), remote_ip, curl_easy_strerror(rc), rc);
       }
-      else
+      else if (http_rc >= 500)
       {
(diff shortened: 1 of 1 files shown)


bossmc:
Looks sensible, does this resolve https://github.com/Metaswitch/sprout/issues/1300?

steelyvoid:
Yep - should do.  I'll close that down

steelyvoid:
Turns out that Ellie has already fixed this as part of https://github.com/Metaswitch/cpp-common/commit/b5ab5b70946af8ea734c1c88cc80756464a78290, so no need for this PR after all

PR #438, opened 2016-01-04
[Reviewer: Rob] Creates a custom alarm definition (cpp-common)

AME2:


include/alarmdefinition.h
```diff
@@ -81,12 +81,24 @@ namespace AlarmDef {
   };
 
   struct SeverityDetails {
-    Severity    _severity;
+    SeverityDetails() {}
```
rkd-msw: We'd normally:

- indent the initialiser list (e.g. https://github.com/Metaswitch/cpp-common/blob/master/include/baseresolver.h#L60)
- put a blank line after each constructor (https://github.com/Metaswitch/cpp-common/blob/master/include/httpconnection.h#L77)
- put a semicolon after the `{}`

rkd-msw:
A couple of stylistic comments, otherwise good.

One thought - is it worth switching the code at https://github.com/Metaswitch/cpp-common/blob/62342ffcb2adc003c17a10e338956d38b4e3972f/src/json_alarms.cpp#L180 over to use the new constructor? I can't find any other code that constructs either object, so you probably don't need the no-argument constructor if you change that (though it's worth testing that clearwater-snmp-handlers compiles and the tests run, to make sure).

include/test_snmp.h
```diff
@@ -0,0 +1,160 @@
...
+  std::string command = "snmpwalk -v2c -OQn -c clearwater 127.0.0.1:16161 " + oid;
+  std::string mode = "r";
+  FILE* fd = popen(command.c_str(), mode.c_str());
+  char buf[1024];
+  char* fgets_pointer;
+  // fgets returns a null pointer when it reaches EOF.
+  fgets_pointer = fgets(buf, sizeof(buf), fd);
+  char empty[] = "No Such Object available on this agent at this OID";
+  // Checks if the table is empty.
+  if (!fgets_pointer || strstr(buf,empty))
```
rkd-msw: I might reverse this logic:

```
if (fgets_pointer || !strstr(buf, empty))
{
  // We've got data, push it into our vector
  ...
}

return res;
```

This means you only have one return statement, which is good practice - e.g. it's easier to see that you've freed all memory before returning, or to log things out about what you're returning.

include/test_snmp.h
```diff
@@ -0,0 +1,160 @@
+/**
```
rkd-msw: It might make more sense for this file to be in test_utils instead of include.

rkd-msw:
Looks good - are you also changing clearwater-fv-test to use this test_snmp.h file?

rkd-msw:
Looks good.

PR #439, opened 2016-01-07
[Reviewer: Ellie] Update increment/decrement methods to handle integer arguments

johadalin:
I don't think there is anything particularly controversial here. 

I'm not sure if there's a better way to protect against overflow in timer_counter.cpp, so i'd appreciate some feedback there.


src/timer_counter.cpp
```diff
@@ -54,28 +54,46 @@ TimerCounter::TimerCounter():
...
 
+  // To protect against overflow, verify that the current value is greater
+  // than the decrement count. If it isn't, write to statistics decrementing
+  // by the current value instead.
+
   refresh_statistics(five_second.get_current(now), now, five_second.get_interval_ms());
-  write_statistics(five_second.get_current(now), -1);
+  if (five_second.get_current(now)->current_value.load() > count)
+  {
+    write_statistics(five_second.get_current(now), -count);
```
eleanor-merry: You can simplify this by having the check in write_statistics

eleanor-merry:
Looks good

johadalin:
Markups done.
Could you just check the new overflow protection in timer_counter.cpp will work?

src/timer_counter.cpp
```diff
@@ -132,6 +130,13 @@ void TimerCounter::refresh_statistics(SNMP::ContinuousStatistics* data, timespec
 
 void TimerCounter::write_statistics(SNMP::ContinuousStatistics* data, int value_delta)
 {
+  // First check that the value delta will not cause uint stats to overflow.
+  // If value_delta is negative, and larger than the current value, substitute
+  // it with the current value, made negative.
+  if ((value_delta < 0) && ((uint32_t)abs(value_delta) > data->current_value.load()))
+  {
```
eleanor-merry: Why not just set new_value to 0? It's clearer to read what's going on then

PR #441, opened 2016-01-12
[Reviewer: Andy] Integrate ENT logs with communication monitor

### LINKED ISSUE #317: ENT logs often spam
There are situations in which ENT logs on failures that could occur quite frequently - see https://github.com/Metaswitch/sprout/issues/973 for one such example.

The individual conditions are sensible, so I suspect we need to suppress ENT logs when they occur too frequently - e.g. generating one log when it first happens and then a summary log after some period (and nothing else).

eleanor-merry:
Andy, can you review this fix to add ENT logs to the communication monitor (so that the per call ENT logs can be dropped).

Fixes #317

### DIFF
--- a/src/communicationmonitor.cpp
+++ b/src/communicationmonitor.cpp
@@ -88,6 +88,7 @@ void CommunicationMonitor::track_communication_changes(unsigned long now_ms)
       //    in the last 'set_confirm' ms.
       if ((!_error_state) && (succeeded == 0) && (failed != 0))
       {
+        _error_state = true;
         CL_CM_CONNECTION_ERRORED.log(_sender.c_str(),
                                      _receiver.c_str());
 
@@ -95,11 +96,11 @@ void CommunicationMonitor::track_communication_changes(unsigned long now_ms)
         {
           TRC_STATUS("Setting alarm %d", _alarm->index());
           _alarm->set();
-          _error_state = true;
         }
       }
       else if ((_error_state) && (succeeded != 0))
       {
+        _error_state = false;
         CL_CM_CONNECTION_CLEARED.log(_sender.c_str(),
                                      _receiver.c_str());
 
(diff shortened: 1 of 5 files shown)


src/communicationmonitor.cpp
```diff
@@ -75,39 +76,43 @@ void CommunicationMonitor::track_communication_changes(unsigned long now_ms)
...
-                  succeeded, failed);
-        if ((succeeded == 0) && (failed != 0))
+        CL_CM_CONNECTION_ERRORED.log(_sender.c_str(),
+                                     _receiver.c_str());
+
+        if ((_alarm != NULL) && (!_alarm->alarmed()))
         {
           TRC_STATUS("Setting alarm %d", _alarm->index());
           _alarm->set();
+          _error_state = true;
```
bossmc: If we don't have an alarm, we'll drop the log every time round the loop.  Is this deliberate?

src/communicationmonitor.cpp
```diff
@@ -75,39 +76,43 @@ void CommunicationMonitor::track_communication_changes(unsigned long now_ms)
...
-      else
+      else if ((_error_state) && (succeeded != 0))
       {
-        // A communication alarm is currently set so see if it needs to be
-        // cleared. This will be the case if at lease one successful comm
-        // was reported over the interval.
-        TRC_DEBUG("Alarm currently set - successful attempts %d",
-                  succeeded);
-        if (succeeded != 0)
+        CL_CM_CONNECTION_CLEARED.log(_sender.c_str(),
```
bossmc: If there's no alarm, this will never be hit (since we never enter error state).

bossmc:
Generally looks good, I think there's an issue with uses where we don't have an alarm object to work with.

src/communicationmonitor.cpp
eleanor-merry: It's not - I'll move it out of the if

PR #442, opened 2016-01-12
Reviewer AJH: Check for port in astaire resolver

graemerobertson:
The Astaire resolver may need to resolve something of the form <host>:<port> according to the spec, and I don't think it currently does. This addresses that issue.

Graeme

### DIFF
--- a/src/astaire_resolver.cpp
+++ b/src/astaire_resolver.cpp
@@ -63,6 +63,8 @@ void AstaireResolver::resolve(const std::string& host,
                               std::vector<AddrInfo>& targets,
                               SAS::TrailId trail)
 {
+  std::string host_without_port;
+  int port;
   AddrInfo ai;
   int dummy_ttl = 0;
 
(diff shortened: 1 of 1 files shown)


src/astaire_resolver.cpp
```diff
@@ -71,19 +71,29 @@ void AstaireResolver::resolve(const std::string& host,
...
-  if (parse_ip_target(host, ai.address))
+  // Check if host contains a port. Otherwise use the default PORT.
+  std::string host_without_port = host;
+  int port = PORT;
+  size_t pos = host.find(":");
+  if (pos != std::string::npos)
+  {
+    host_without_port = host.substr(0, pos);
+    port = stoi(host.substr(pos+1));
+  }
```
AlexHockey: This does't deal with IPv6 addresses. You need to use `Utils::split_host_port`.

bossmc: As @eleanor-merry would have told you :grinning:

AlexHockey:
Looks OK, though it's worth noting that Astaire's listen port is not configurable so we don;'t actually need this code right now. One comment inline. 

How have you tested this? Do you need to add a testcase to clearwater-fv-test?

PR #445, opened 2016-01-18
[Reviewer: Graeme] Blacklist Astaires when they fail

AlexHockey:
Graeme, please can you review this PR to blacklist an Astaire when it fails? I've checked that all the FV tests still run successfully. I've not added a test to check the node is actually blacklisted, as this is difficult to do in the FV tests without snooping the network traffic.


graemerobertson:
Well that was easy - I could have done it myself! Looks good.

PR #448, opened 2016-01-20
[Reviewer: Rob] Fixes iss1185: UTs now clean up

AME2:


include/baseresolver.h
```diff
@@ -105,6 +105,7 @@ class BaseResolver
   /// Utility function to parse a target name to see if it is a valid IPv4 or IPv6 address.
   bool parse_ip_target(const std::string& target, IP46Address& address);
 
+  void clear_blacklist() {_blacklist.clear();}
```
rkd-msw: Minor point - this should probably take _blacklist_lock. This doesn't matter hugely for this use-case, as it's only being called at the end of UTs, but if this function were called at the same time as some other function manipulating the map, the simultaneous writes to the map could corrupt it and cause crashes.

PR #449, opened 2016-01-20
[No Reviewer] Stats api update

johadalin:
Enabling use of the new Chronos API for tags

### DIFF
--- a/src/chronosconnection.cpp
+++ b/src/chronosconnection.cpp
@@ -46,7 +46,7 @@
 #include "httpconnection.h"
 #include "chronosconnection.h"
 
-const std::vector<std::string> ChronosConnection::EMPTY_TAGS = std::vector<std::string>();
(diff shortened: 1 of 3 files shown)


src/chronosconnection.cpp
```diff
@@ -209,14 +209,21 @@ std::string ChronosConnection::create_body(uint32_t interval,
...
-      writer.String("tags");
+      writer.String("tag-info");
       writer.StartArray();
       {
-        for (std::vector<std::string>::const_iterator it = tags.begin();
-                                                      it != tags.end();
-                                                      ++it)
+        for (std::map<std::string, uint32_t>::const_iterator it = tags.begin();
+                                                       it != tags.end();
+                                                       ++it)
```
chris-elford-metaswitch: Is it worth lining these back up?

PR #452, opened 2016-02-02
[Reviewer AJH] Retry attempt to open log file after 5s.

chris-elford-metaswitch:
As part of issue 383, we should attempt to reopen a log file every 5s until we succeed. This change adds that function, and also unit test infrastructure to allow us to control the results of attempting to open a file.

Tested through Sprout unit tests, and live.

### DIFF
--- a/test_utils/test_interposer.cpp
+++ b/test_utils/test_interposer.cpp
@@ -71,11 +71,18 @@ static const clockid_t supported_clock_ids[] = {CLOCK_REALTIME,
 static std::map<clockid_t, struct timespec> abs_timespecs;
 static time_t abs_time;
 
+// Whether file opening is completely controlled by the test script.
+static bool control_fopen = false;
+
+// When controlling fopen, this stores the file pointer to return.
+FILE* fopen_file_pointer = NULL;
+
 /// The real functions we are interposing.
 static int (*real_getaddrinfo)(const char*, const char*, const struct addrinfo*, struct addrinfo**);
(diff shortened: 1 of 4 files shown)


test_utils/test_interposer.cpp
```diff
@@ -71,11 +71,18 @@ static const clockid_t supported_clock_ids[] = {CLOCK_REALTIME,
 static std::map<clockid_t, struct timespec> abs_timespecs;
 static time_t abs_time;
 
+// Whether file opening is completely controlled by the test script.
+static bool control_fopen = false;
+
+// When controlling fopen, this stores the file pointer to return.
+FILE* fopen_file_pointer = NULL;
```
AlexHockey: Slightly cleaner to make this static as this means no-one outside of the file can access it. 

(it doesn't matter too much in this case since the variable is not defined in a header file so no-one should be able to access it anyway, but using a static makes the intent clearer to the reader)

test_utils/test_interposer.cpp
```diff
@@ -332,3 +339,39 @@ int pthread_cond_timedwait(pthread_cond_t* cond,
 
   return real_pthread_cond_timedwait(cond, mutex, &fixed_time);
 }
+
+
+void cw_control_fopen(FILE* fd)
```
AlexHockey: Please can you use the `cwtest_` prefix like the rest of the interposers?

AlexHockey:
Looks good. A couple of minor comments.

PR #453, opened 2016-02-12
[Reviewer: Rob] Alarm refactoring: Adjusts Alarm class hierarchy so that alarms...

AME2:
with more than one severity are represented as one object of class MultiStateAlarm.

The main changes here include creating the superclass: BaseAlarm that allows us to split different types of alarms into subclasses. Those alarms which only have one possible raised state will be constructed by subclass Alarm. Those alarms which have two or more possible raised states will be constructed by subclass MultiStateAlarm.

MultiStateAlarm is then further subclassed into classes that represent individual alarms. These subclasses gives the MultiStateAlarm objects visibility of just those functions corresponding to states of the alarm that can be raised.

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -74,12 +74,40 @@ private:
   std::string _identifier;
 };
 
+/// @class BaseAlarm
+/// 
+/// Superclass for alarms that allows us to split different types of
+/// alarms into subclasses. Those alarms which only have one possible raised
+/// state will be constructed by subclass Alarm. Those alarms which have two or
+/// more possible raised states will be constructed by subclass
+/// MultiStateAlarm. 
+
+class BaseAlarm
+{
+public:
+  /// Queues a request to generate an alarm state change corresponding to the
+  /// CLEARED severity.
+  virtual void clear();
+  
+  /// Indicates whether the alarm state currently maintained by this object
+  /// corresponds to the non-CLEARED severity.
+  virtual bool alarmed() {return _alarmed.load();}
+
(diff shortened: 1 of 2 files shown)


src/alarm.cpp
```diff
@@ -106,6 +115,56 @@ void Alarm::clear()
   }
 }
 
+void MultiStateAlarm::multi_set_indeterminate()
+{
+  bool previously_alarmed = _alarmed.exchange(true);
+
+  if (!previously_alarmed)
```
rkd-msw: Should we be checking this? It looks as though, if I raised a major alarm and it then became critical, we'd see that `previously_alarmed` was true and not issue the new alarm.

include/alarm.h
```diff
@@ -89,28 +121,68 @@ class Alarm
...
-  /// Indicates that the alarm state currently maintained by this object
-  /// corresponds to the non-CLEARED severity.
-  virtual bool alarmed() {return _alarmed.load();}
-
   /// Returns the index of this alarm.
   virtual int index() const {return _index;}
+  
+  /// Indicates whether the alarm state currently maintained by this object
+  /// corresponds to the non-CLEARED severity.
+  virtual bool alarmed() {return _alarmed.load();}
```
rkd-msw: Should this just be in BaseAlarm? If not, should _alarmed also be in the subclass?

rkd-msw:
Couple of comments, and as discussed, you should move the license-specific alarms elsewhere.

PR #456, opened 2016-02-16
[Reviewer: Chris] Resend status of alarms to alarm handler- cpp-common

AME2:
The main changes here involve changes made to the alarm class hierarchy and creating functions to reraise the alarms every 30 seconds.

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -92,6 +92,8 @@ public:
   /// CLEARED severity.
   virtual void clear();
 
+  /// Uses the _last_state_raised member variable to re-raise the latest state
+  /// of the alarm.
   virtual void reraise_last_state();
 
 protected:
@@ -101,6 +103,9 @@ protected:
   /// Keeps track of whether the alarm is raised
   std::atomic<bool> _alarmed;
 
+  // Keeps track of the latest state of each alarm that has been raised. If the
+  // alarm has just been cleared this would be the corresponding _clear_state
+  // for the alarm.
   AlarmState* _last_state_raised;
 };
 
@@ -123,7 +128,11 @@ public:
 private:
   AlarmManager();
   ~AlarmManager();
+  // This runs on a thread (defined below) and iterates over _global_alarm_list
+  // every 30 seconds. For each alarm it calls the reraise_last_state method.
   virtual void reraise_alarms();
+  // This is used for storing all of the BaseAlarm objects as they get
+  // constructed.
   std::vector<BaseAlarm*> _global_alarm_list;
   pthread_mutex_t _lock;
   pthread_cond_t _terminating_variable;
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -74,12 +74,40 @@ private:
   std::string _identifier;
 };
 
+/// @class BaseAlarm
+/// 
+/// Superclass for alarms that allows us to split different types of
+/// alarms into subclasses. Those alarms which only have one possible raised
+/// state will be constructed by subclass Alarm. Those alarms which have two or
+/// more possible raised states will be constructed by subclass
+/// MultiStateAlarm. 
+
+class BaseAlarm
+{
+public:
+  /// Queues a request to generate an alarm state change corresponding to the
+  /// CLEARED severity.
+  virtual void clear();
+  
+  /// Indicates whether the alarm state currently maintained by this object
+  /// corresponds to the non-CLEARED severity.
+  virtual bool alarmed() {return _alarmed.load();}
+
+protected:
+  /// Keeps track of whether the alarm is raised.
+  std::atomic<bool> _alarmed;
+
+  AlarmState _clear_state;
+  const int _index;
+};
+
 /// @class Alarm
 ///
-/// Encapsulates an alarm active state and its associated alarm clear state.
+/// Encapsulates an alarm's only active state with its associated alarm clear state.
 /// Used to manage the reporting of a fault condition, and subsequent clear
 /// of said condition.
-class Alarm
+
+class Alarm: public BaseAlarm
 {
 public:
   Alarm(const std::string& issuer,
@@ -89,28 +117,59 @@ public:
   virtual ~Alarm() {}
 
   /// Queues a request to generate an alarm state change corresponding to the
-  /// CLEARED severity if a state change for the non-CLEARED severity was
-  /// previously requested via set().
-  virtual void clear();
-
-  /// Queues a request to generate an alarm state change corresponding to the
   /// non-CLEARED severity if a state change for the CLEARED severity was
   /// previously requestd via clear().
   virtual void set();
 
-  /// Indicates that the alarm state currently maintained by this object
-  /// corresponds to the non-CLEARED severity.
-  virtual bool alarmed() {return _alarmed.load();}
-
-  /// Returns the index of this alarm.
+    /// Returns the index of this alarm.
   virtual int index() const {return _index;}
 
 private:
-  const int _index;
-  AlarmState _clear_state;
   AlarmState _set_state;
 
   std::atomic<bool> _alarmed;
+};
+
+/// @class MultiStateAlarm
+///
+/// Encapsulates an alarm's two or more active states with its associated clear
+/// state. Used to manage the reporting of a fault condition, and subsequent
+/// clear of said condition.
+
+class MultiStateAlarm: public BaseAlarm
+{
+public:
+  MultiStateAlarm(const std::string& issuer,
+                  const int index,
+                  std::vector<AlarmDef::Severity> severities);
+
+protected:
+  /// These raise the alarm with the specified severity.
+  virtual void set_indeterminate();
+  virtual void set_warning();
+  virtual void set_minor();
+  virtual void set_major();
+  virtual void set_critical();
+
+private:
+  AlarmState _indeterminate_state;
+  AlarmState _warning_state;
+  AlarmState _minor_state;
+  AlarmState _major_state;
+  AlarmState _critical_state;
+};
+
+/// @class CassandraLicenseErrorAlarm
+///
+/// Represents a Clearwater alarm with multiple raised states. This class gives
+/// the MultiStateAlarm object visibility of just those functions corresponding
+/// to states of the alarm that can be raised.
+
+class CassandraLicenseErrorAlarm: public MultiStateAlarm
+{
+public:
+  set_major() {MultiStateAlarm::set_major()};
+  set_critical() {MultiStateAlarm::set_critical()};
 };
 
 /// @class AlarmReqAgent
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -95,7 +95,7 @@ public:
   
   /// Indicates whether the alarm state currently maintained by this object
   /// corresponds to the non-CLEARED severity.
-  virtual bool alarmed() {return _alarmed.load();}
+  virtual bool alarmed() {return (_last_state_raised == &_clear_state ? false : true);}
 
 protected:
   BaseAlarm(const std::string& issuer,
@@ -104,9 +104,6 @@ protected:
   const int _index;
   AlarmState _clear_state;
   
-  /// Keeps track of whether the alarm is raised
-  std::atomic<bool> _alarmed;
-
   // Keeps track of the latest state of each alarm that has been raised. If the
   // alarm has just been cleared this would be the corresponding _clear_state
   // for the alarm.
@@ -123,15 +120,17 @@ class AlarmManager
 public:
   static AlarmManager& get_instance() {return _instance;}
   bool _terminated;
-  bool _first_alarm_raised;
   void register_alarm(BaseAlarm* alarm); 
   // Used to stop re-raising alarms in UTs
   void alarm_list_clear(void) {_alarm_list.clear();}
+  void start_resending_alarms(void) {_first_alarm_raised = true;}
+  bool has_alarm_been_raised(void) {return _first_alarm_raised;}
 
 private:
   AlarmManager();
   ~AlarmManager();
 
+  bool _first_alarm_raised;
   // Static function called by the reraising alarms thread. This simply calls
   // the 'reraise_alarms' member method
   static void* reraise_alarms_function(void* data);
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -87,7 +87,7 @@ class BaseAlarm
 public:
   /// Queues a request to generate an alarm state change corresponding to the
   /// CLEARED severity.
-  void clear();
+  virtual void clear();
 
   /// Uses the _last_state_raised member variable to re-raise the latest state
   /// of the alarm.
@@ -95,7 +95,7 @@ public:
   
   /// Indicates whether the alarm state currently maintained by this object
   /// corresponds to the non-CLEARED severity.
-  bool alarmed() {return _alarmed.load();}
+  virtual bool alarmed() {return _alarmed.load();}
 
 protected:
   BaseAlarm(const std::string& issuer,
@@ -161,12 +161,12 @@ public:
         const int index,
         AlarmDef::Severity severity);
 
-  ~Alarm() {}
+  virtual ~Alarm() {}
 
   /// Queues a request to generate an alarm state change corresponding to the
   /// non-CLEARED severity if a state change for the CLEARED severity was
   /// previously requested via clear().
-  void set();
+  virtual void set();
 
   /// Returns the index of this alarm.
   int index() const {return _index;}
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -90,11 +90,15 @@ public:
   
   /// Queues a request to generate an alarm state change corresponding to the
   /// CLEARED severity.
-  virtual void clear();
+  void clear();
 
   /// Uses the _last_state_raised member variable to re-raise the latest state
   /// of the alarm.
-  virtual void reraise_last_state();
+  void reraise_last_state();
+  
+  /// Indicates whether the alarm state currently maintained by this object
+  /// corresponds to the non-CLEARED severity.
+  bool alarmed() {return _alarmed.load();}
 
 protected:
   const int _index;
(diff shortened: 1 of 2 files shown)


include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+
+class BaseAlarm
+{
+public:
+  BaseAlarm(const std::string& issuer,
+            const int index);
+  
+  /// Queues a request to generate an alarm state change corresponding to the
+  /// CLEARED severity.
+  virtual void clear();
```
chris-elford-metaswitch: As the code stands, this doesn't need to be virtual as you don't override it in subclasses.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+
+  /// Uses the _last_state_raised member variable to re-raise the latest state
+  /// of the alarm.
+  virtual void reraise_last_state();
```
chris-elford-metaswitch: Ditto

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+/// @class AlarmManager
+///
+/// Singleton class responsible for calling BaseAlarm's reraise_latest_state
+/// function on each alarm every thirty seconds.
+
+class AlarmManager
+{
+public:
+  static AlarmManager& get_instance() {return _instance;}
+  static AlarmManager _instance;
```
chris-elford-metaswitch: This should be private - the instance must never be assigned to from outside the class, and must only be accessed through get_instance.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+  bool _terminated;
+  void add_alarm_to_list(BaseAlarm* alarm);
```
chris-elford-metaswitch: I think 'register_alarm' is a better name as it describes what the function is doing in terms of purpose rather than technical details.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+  static void* reraise_alarms_function(void* data); 
+
+private:
+  AlarmManager();
+  ~AlarmManager();
+  // This runs on a thread (defined below) and iterates over _global_alarm_list
+  // every 30 seconds. For each alarm it calls the reraise_last_state method.
+  virtual void reraise_alarms();
+  // This is used for storing all of the BaseAlarm objects as they get
+  // constructed.
```
chris-elford-metaswitch: Constructed or registered? (See above)

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+  std::vector<BaseAlarm*> _global_alarm_list;
+  pthread_mutex_t _lock;
+  pthread_cond_t _terminating_variable;
+  pthread_t _reraising_alarms_thread;
+};
+
 /// @class Alarm
 ///
-/// Encapsulates an alarm active state and its associated alarm clear state.
+/// Encapsulates an alarm's only active state with its associated alarm clear state.
```
chris-elford-metaswitch: It's good to do this kind of touch-up work when you are in the area.

include/alarm.h
```diff
@@ -89,28 +155,51 @@ class Alarm
   virtual ~Alarm() {}
```
chris-elford-metaswitch: You are using 'virtual' a lot, but it's only necessary when you define a method which needs to be overridden by a subclass. Alarm and MultiSeverityAlarm are not subclassed, so they don't need any virtual methods.

include/alarm.h
```diff
@@ -89,28 +155,51 @@ class Alarm
...
-  /// Indicates that the alarm state currently maintained by this object
-  /// corresponds to the non-CLEARED severity.
-  virtual bool alarmed() {return _alarmed.load();}
-
   /// Returns the index of this alarm.
   virtual int index() const {return _index;}
+  
+  /// Indicates whether the alarm state currently maintained by this object
+  /// corresponds to the non-CLEARED severity.
+  virtual bool alarmed() {return _alarmed.load();}
```
chris-elford-metaswitch: You've defined the same method twice! You should consolidate this into the superclass - one of the main reasons for using inheritance is to re-use code.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
   TRC_DEBUG("%s cleared its alarms", issuer.c_str());
 }
 
+BaseAlarm::BaseAlarm(const std::string& issuer,
```
chris-elford-metaswitch: It's worth thinking a little about how you lay out the file. I'd define all of the methods for one class, then move on to the next.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+
+private:
+  AlarmManager();
+  ~AlarmManager();
+  // This runs on a thread (defined below) and iterates over _global_alarm_list
+  // every 30 seconds. For each alarm it calls the reraise_last_state method.
+  virtual void reraise_alarms();
+  // This is used for storing all of the BaseAlarm objects as they get
+  // constructed.
+  std::vector<BaseAlarm*> _global_alarm_list;
```
chris-elford-metaswitch: This is called _global_alarm_list, but it's not a global variable.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+{
+  ((AlarmManager*)data)->reraise_alarms();
+  return NULL;
+}
+
+void AlarmManager::reraise_alarms()
+{
+  while (!_terminated)
+  {
+    for (std::vector<BaseAlarm*>::iterator it = _global_alarm_list.begin(); it != _global_alarm_list.end(); it++)
```
chris-elford-metaswitch: I'd prefer to use a [range loop](http://en.cppreference.com/w/cpp/language/range-for) here, but let's hold off until after Alex's write-up.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+    {
+      (*it)->reraise_last_state();
+    }
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    ts.tv_sec += 30;
```
chris-elford-metaswitch: We discussed this timing issue before when we went over your design. You are waiting 30s *after* re-syncing all alarms, but the re-sync will take non-zero time. This means that you don't re-sync every 30s.

I think what you should do is:
- store the previous time on the AlarmManager.
- add 30s to it and subtract the current time.
- if that time is more than 0, wait for it.
- otherwise, write a log and sleep until the next valid 30s interval.

include/alarm.h
bossmc: Although, if you want to use polymorphism (treating a subclass as a superclass) you always want the destructor to be virtual (I think g++ warns about this, clang certainly does).

On the other hand, `virtual` is inherited from the parent class (in this case `Alarm`) and thus doesn't need to be specified here again.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+    pthread_cond_timedwait(&_terminating_variable, &_lock, &ts);
+  }
+}
+
+AlarmManager::AlarmManager()
+{
+  _terminated = false;
+  pthread_mutex_init(&_lock, NULL);
```
chris-elford-metaswitch: I don't quite understand what this code is doing. A short high-level comment might help e.g. "Protect x with a condition variable."

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+  _last_state_raised = &_critical_state;
+  _critical_state.issue();
+}
+
+void BaseAlarm::reraise_last_state()
+{
+  _last_state_raised->issue();
+}
+
+void* AlarmManager::reraise_alarms_function(void* data)
```
chris-elford-metaswitch: Add a comment explaining why this function is defined with this signature. I think it needs to fit the signature expected by pthread_create.
- What is the data passed in?
- Why do you return NULL?

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+  pthread_mutex_t _lock;
+  pthread_cond_t _terminating_variable;
```
chris-elford-metaswitch: I'm not sure this name conveys what the condition variable is for - it's used when the thread is not terminating. I might be tempted just to call it _condition.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+private:
+  AlarmManager();
+  ~AlarmManager();
+  // This runs on a thread (defined below) and iterates over _global_alarm_list
+  // every 30 seconds. For each alarm it calls the reraise_last_state method.
+  virtual void reraise_alarms();
+  // This is used for storing all of the BaseAlarm objects as they get
+  // constructed.
+  std::vector<BaseAlarm*> _global_alarm_list;
+  pthread_mutex_t _lock;
```
chris-elford-metaswitch: What does this lock protect? Please add a comment.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+class AlarmManager
+{
+public:
+  static AlarmManager& get_instance() {return _instance;}
+  static AlarmManager _instance;
+  bool _terminated;
+  void add_alarm_to_list(BaseAlarm* alarm); 
+  // Static function called by the reraising alarms thread. This simply calls
+  // the 'reraise_alarms' member method.
+  static void* reraise_alarms_function(void* data);
```
chris-elford-metaswitch: I assumed that this would be private. Is that possible? (My C++ isn't quite good enough to tell.)

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+/// Superclass for alarms that allows us to split different types of
+/// alarms into subclasses. Those alarms which only have one possible raised
+/// state will be constructed by subclass Alarm. Those alarms which have two or
+/// more possible raised states will be constructed by subclass
+/// MultiStateAlarm. 
+
+class BaseAlarm
+{
+public:
+  BaseAlarm(const std::string& issuer,
```
chris-elford-metaswitch: This should be protected - you don't want people creating BaseAlarm objects.

include/alarm.h
```diff
@@ -74,12 +74,78 @@ class AlarmState
...
+  /// Uses the _last_state_raised member variable to re-raise the latest state
+  /// of the alarm.
+  virtual void reraise_last_state();
+
+protected:
+  const int _index;
+  AlarmState _clear_state;
+  
+  /// Keeps track of whether the alarm is raised
+  std::atomic<bool> _alarmed;
```
chris-elford-metaswitch: I'm not convinced by this bool. You are currently exposed to the following threading bug:
1. Thread A sets _alarmed to true.
2. Thread B sets _alarmed to false.
3. Thread B sets _last_state_raised to the cleared state.
4. Thread A sets _last_state_raised to the raised state.
At this point, _last_state_raised is the raised state, but _alarmed is false!

At the same time, _last_state_raised contains all of the information about whether the alarm is raised or not - simply compare it to the cleared state.

I think there is an atomic intptr type which you can use to fix this bug.


All of the information about whether this is alarmed is stored in the _last_state_raised pointer.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+{
+  ((AlarmManager*)data)->reraise_alarms();
+  return NULL;
+}
+
+void AlarmManager::reraise_alarms()
+{
+  while (!_terminated)
```
chris-elford-metaswitch: I'd add a debug level log here.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+  {
+    for (std::vector<BaseAlarm*>::iterator it = _global_alarm_list.begin(); it != _global_alarm_list.end(); it++)
+    {
+      (*it)->reraise_last_state();
+    }
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    ts.tv_sec += 30;
+    pthread_cond_timedwait(&_terminating_variable, &_lock, &ts);
+  }
```
chris-elford-metaswitch: I'd add an info level log here saying that the re-raise thread is terminating.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+}
+
+AlarmManager::AlarmManager()
+{
+  _terminated = false;
+  pthread_mutex_init(&_lock, NULL);
+  pthread_condattr_t cond_attr;
+  pthread_condattr_init(&cond_attr);
+  pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC);
```
chris-elford-metaswitch: Why is this necessary?

include/alarm.h
```diff
@@ -89,28 +155,51 @@ class Alarm
...
-  AlarmState _clear_state;
   AlarmState _set_state;
+};
 
-  std::atomic<bool> _alarmed;
+/// @class MultiStateAlarm
+///
+/// Encapsulates an alarm's two or more active states with its associated clear
+/// state. Used to manage the reporting of a fault condition, and subsequent
+/// clear of said condition.
```
chris-elford-metaswitch: Please add a comment explaining that this must be subclassed, and providing an example. Otherwise, someone will just come along and make all of the set methods public...

include/alarm.h
```diff
@@ -89,28 +155,51 @@ class Alarm
...
+public:
+  MultiStateAlarm(const std::string& issuer,
+                  const int index);
+  
+  /// Indicates whether the alarm state currently maintained by this object
+  /// corresponds to the non-CLEARED severity.
+  virtual bool alarmed() {return _alarmed.load();}
+  
+protected:
+  /// These raise the alarm with the specified severity.
```
chris-elford-metaswitch: It's worth explaining again why these are protected.

chris-elford-metaswitch:
This is mostly good, but a few tweaks are needed. I'd like to see again before you merge.

src/alarm.cpp
```diff
@@ -76,36 +77,145 @@ void AlarmState::clear_all(const std::string& issuer)
...
+  _alarmed.exchange(true);
+  _last_state_raised = &_indeterminate_state;
+  _indeterminate_state.issue();
+}
+
+void MultiStateAlarm::multi_set_warning()
+{
+  _alarmed.exchange(true);
+  _last_state_raised = &_warning_state;
+  _warning_state.issue();
```
chris-elford-metaswitch: You've not added the same logic here as in the Alarm to only send if there is a change.

include/alarm.h
chris-elford-metaswitch: More to the point, is this function even used?

AME2: Yes- by the Cedar communication monitor. I'll consolidate them into the superclass.

src/alarm.cpp
AME2: Yes, that logic said only clear an alarm if it is raised and (importantly) only raise an alarm if it is cleared. So I couldn't repeat that logic because with MultiStateAlarms we may want to raise them again if they are already raised. I'm going to change both bits of logic to base the decision off the _last_state_raised BaseAlarm member variable. If the last state raised of the alarm is different from the current state we raise the alarm.

chris-elford-metaswitch: Sounds good to me!

include/alarm.h
AME2: I think the reason behind these virtual functions was so that we could have a MockAlarm subclass that could define mock methods to override these functions:
https://github.com/Metaswitch/cpp-common/blob/ReraisingAlarms/test_utils/mockalarm.h

include/alarm.h
chris-elford-metaswitch: Please address this comment.

include/alarm.h
chris-elford-metaswitch: Makes sense.

include/alarm.h
AME2: We no longer raise alarms based on the _alarmed variable. We now check the new _last_state_raised variable and if this is different from the state of the alarm we which to raise, we raise it.
I believe the only place that still checks this _alarmed variable is the Cedar communication monitor.

chris-elford-metaswitch: As discussed, this needs fixing.

chris-elford-metaswitch:
I've added a few more comments as I think there are still some bugs here. I'm also interested in how many blank lines you've deleted. We tend to use these to separate sections of code and make it more readable so it's not always right to remove them.

include/alarm.h
chris-elford-metaswitch: You still haven't fixed this, or explained why it's OK.

chris-elford-metaswitch:
Mostly good; just a couple of comments.

You've still not addressed my comment about the _alarmed flag. Last I remember, you were going to investigate fixing the threading issue as a way of ensuring that the interface exposed by this code does what is expected. Even if you think this doesn't need fixing, you should add a comment to explain why it doesn't.

AME2:
I've removed the _alarmed private member function of the class BaseAlarm and have modified the alarmed() public method. This method now looks at the _last_state_raised private member variable to decide whether an alarm is currently active (in a non-CLEARED severity) or not. Then to stop the threading race condition I've included a mutex lock in my set_x methods.
In class BaseAlarm I've also moved the _first_alarm_raised variable into the private namespace and now use the methods start_resending_alarms and has_alarm_been_raised to write and read this variable.

johadalin:
I can't see any big problems in the last couple of commits. 
@rkd-msw , if you have the time, can you take a closer look at this. I don't have a particularly good view of the alarms code, so i may well be missing things.

rkd-msw:
LGTM.

PR #457, opened 2016-02-17
[Reviewer: Adam] Cpp common coverage

eleanor-merry:
Adds an interposer to allow tests to use fake curl and real curl.


test_utils/curl_interposer.cpp
```diff
@@ -0,0 +1,446 @@
...
+static void (*real_curl_easy_cleanup)(CURL* handle);
+static CURL* (*real_curl_easy_init)();
+static CURLcode (*real_curl_global_init)(long flags);
+
+CURLcode curl_easy_perform(CURL* handle)
+{
+  if (!real_curl_easy_perform)
+  {
+    real_curl_easy_perform = (CURLcode (*)(CURL*))dlsym(RTLD_NEXT, "curl_easy_perform");
+  }
```
johadalin: do we need to perform the above if we are controlling curl? should it not  be part of the else branch below?

test_utils/fakecurl.cpp
```diff
@@ -44,73 +42,6 @@
...
-  {
-  }
-
-  virtual ~FakeCurl()
-  {
-  }
-
-  CURLcode easy_perform(FakeCurl* curl);
-};
-
```
johadalin: Is there any reason you've moved the above to the header other than for separation of definition and function?

johadalin:
I think it all looks good. I'll grab you about the questions i have above.

PR #460, opened 2016-02-22
Reviewer EM: Add cause, effect and action to alarms

graemerobertson:


src/json_alarms.cpp
```diff
@@ -177,7 +183,16 @@ namespace JSONAlarms
             return false;
           }
 
-          AlarmDef::SeverityDetails sd(e_severity, description, details); 
+          JSON_GET_STRING_MEMBER(*alarms_def_it, "cause", detailed_cause);
```
eleanor-merry: Are there any length limitations on the new fields?

eleanor-merry:
Looks good. You'll also need to update json_alarm_test in cpp-common-test

src/json_alarms.cpp
graemerobertson: No - they are fields we've added and defined ourselves (and which aren't actually sent on alarms)

PR #468, opened 2016-03-08
[Reviewer: Rob] Changes time limit for resending alarms in UTs to 1000s so they pass …

AME2:
…under Valgrind
The important change here is adding on 1000 seconds to the time limit when Unit Testing (instead of 30).

### DIFF
--- a/src/alarm.cpp
+++ b/src/alarm.cpp
@@ -136,10 +136,13 @@ MultiStateAlarm::MultiStateAlarm(const std::string& issuer,
 {
 }
 
+// We don't have any UTs that use an indeterminate state.
+// LCOV_EXCL_START
 void MultiStateAlarm::set_indeterminate()
 {
   switch_to_state(&_indeterminate_state);
 }
+// LCOV_EXCL_STOP
 
 void MultiStateAlarm::set_warning()
 {
@@ -250,8 +253,10 @@ void AlarmManager::reraise_alarms()
       time_limit.tv_nsec += 10 * 1000 * 1000;
       if (time_limit.tv_nsec >= (1000 * 1000 * 1000))
       {
+        // LCOV_EXCL_START
         time_limit.tv_nsec -= 1000 * 1000 * 1000;
         time_limit.tv_sec += 1;
+        // LCOV_EXCL_STOP
       }
       // We want to temporarily reduce the time_limit for UTs to 10ms, this
       // allows us to pass pthread_cond_timedwait quickly but keeps us in this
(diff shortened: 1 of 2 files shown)


src/alarm.cpp
```diff
@@ -213,7 +213,11 @@ void AlarmManager::reraise_alarms()
   {
     // Sets the limit for when we want the thread to wake up and start
     // re-issueing alarms again.
+#ifdef UNIT_TEST
+    time_limit.tv_sec += 1000;
```
rkd-msw: Could do with a comment to say why we have a higher limit in UT.

src/alarm.cpp
```diff
@@ -227,19 +231,26 @@ void AlarmManager::reraise_alarms()
...
 #ifdef UNIT_TEST
-      time_limit.tv_nsec += 10000000;
-      time_limit.tv_sec -= 30;
+      time_limit.tv_nsec += 10 * 1000 * 1000;
+      if (time_limit.tv_nsec >= (1000 * 1000 * 1000))
+      {
+        time_limit.tv_nsec -= 1000 * 1000 * 1000;
+        time_limit.tv_sec += 1;
+      }
+      time_limit.tv_sec -= 1000;
```
rkd-msw: As discussed, it would be good to have a comment explaining why we subtract 1,000 then add it - this is slightly subtle (at first glance, it feels like a no-op that you can just remove) and tripped me up.

src/alarm.cpp
```diff
@@ -227,19 +231,26 @@ void AlarmManager::reraise_alarms()
...
+            !_terminated)
     {
       // When we are unit testing this function we want to sleep in 10ms
       // increments. This gives the UT a chance to simulate 30 seconds of 
       // time passing to cause all the alarms to be re-raised.
 #ifdef UNIT_TEST
-      time_limit.tv_nsec += 10000000;
-      time_limit.tv_sec -= 30;
+      time_limit.tv_nsec += 10 * 1000 * 1000;
+      if (time_limit.tv_nsec >= (1000 * 1000 * 1000))
```
rkd-msw: Maybe worth a comment saying "Make sure that adding 10ms doesn't bring us over a billion nanoseconds" - although I realise that this is based on code I wrote without that comment!

src/alarm.cpp
rkd-msw: Actually, it might be worth having `int TIME_BETWEEN_CHECKS = 1000` - that way lines 248 and 253 can be `time_limit.tv_sec -= TIME_BETWEEN_CHECKS;` and `time_limit.tv_sec += TIME_BETWEEN_CHECKS;`, which is more readable and has less risk of being confused with the other 1000s in the code.

rkd-msw:
Couple of comments. Does this fix all the builds currently broken in Jenkins?

AME2:
I believe it will, the tests in Clearwater-snmp-handlers and cpp-common-test both pass now under Valgrind (not coverage though).

rkd-msw:
Will the Jenkins build fail due to your changes because of coverage? If so, what are you planning to do about that?

rkd-msw:
Looks good.

PR #470, opened 2016-03-08
Reviewer Rob - Create the alarm_resend thread lazily

bossmc:
Don't spawn threads in things created statically, instead spawn the thread lazily the first time someone calls into `AlarmManager::get_instance()` which happens (among other times) when alarms are defined.  This means you should always wait until after demonizing/forking before defining alarms but this is a pretty reasonable requirement.

Also fixed up a locking mistake (we weren't protecting access to the list of registered alarms even though we iterate over the list in another thread).

Testing done:

- [x] UTs
- [ ] Clearwater Live Test - Doesn't test alarms
- [x] Manual live verification - Tested that alarms are still raised as expected and that the resync thread is running (checked in `gdb`)

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -122,7 +122,7 @@ protected:
 class AlarmManager
 {
 public:
-  static AlarmManager& get_instance() {return _instance;}
+  static AlarmManager& get_instance();
   bool _terminated;
   void register_alarm(BaseAlarm* alarm); 
   // Used to stop re-raising alarms in UTs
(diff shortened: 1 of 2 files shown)


rkd-msw:
Yep, happy with this.

One thought - should we be doing the forking? We're split between doing it ourselves (Sprout, the Python processes) and passing --background to start-stop-daemon to do it for us, and I think the latter would have prevented this bug (& stops us having to worry about this class of bugs in future).

rkd-msw:
Also also, seeing the locking mistake makes me think we should be running more of our stuff under Helgrind to catch this, but I'll have a think about that.

bossmc:
I've raised #473 to cover normalizing our deamonization process.

PR #471, opened 2016-03-09
Reviewer AMC: Expire data if it expires now

graemerobertson:
As discussed, bring store implementations into line for expiry of 0, settling on MemcachedStore's behaviour.

I don't think this will break anything and I've run the Ralf and Sprout UTs successfully. Jenkins will do the rest.


bossmc:
Looks great, couldn't have done it better myself.

PR #472, opened 2016-03-09
Reviewer Rob - Store the AlarmManager in a unique pointer to prevent leaks

bossmc:
Don't leak the `AlarmManager`.  I'll raise an issue for the use of `start-stop-daemon --daemonize`.


rkd-msw:
Worth a comment to explain what happens if we don't have unique_ptr? Otherwise fine.

PR #474, opened 2016-03-11
[Reviewer AMC] Switch ENT logging channel to local7

steelyvoid:
Switch to a less congested logging facility for ENT logs

### DIFF
--- a/include/pdlog.h
+++ b/include/pdlog.h
@@ -1,5 +1,5 @@
 /**
- * @file pdlog.h Enhanced Node Troubleshooting PDLog classes 
+ * @file pdlog.h Enhanced Node Troubleshooting PDLog classes
  *
  * Project Clearwater - IMS in the Cloud
  * Copyright (C) 2014  Metaswitch Networks Ltd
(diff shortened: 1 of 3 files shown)


steelyvoid:
Hi Andy.  Can you rereview this and the sprout pull to check that the refactorisation is what you were expecting?  I'll assign both PRs to you

src/pdlog.cpp
```diff
@@ -0,0 +1,55 @@
+/**
+ * @file pdlog.cpp
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2013  Metaswitch Networks Ltd
```
bossmc: Year

src/pdlog.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <boost/filesystem.hpp>
+
+void PDLogStatic::init(char *pname)
+{
+  // Get the full path of the executable from the first command line argument
+  boost::filesystem::path p = pname;
+
+  // Copy the filename to a string so that we can be sure of its lifespan -
+  // the value passed to openlog must be valid for the duration of the program.
+  std::string *filename = new std::string(p.filename().c_str());
```
bossmc: You leak this... but maybe we don't care.

include/pdlog.h
```diff
@@ -50,32 +50,38 @@ extern "C" {
 #include "syslog_facade.h"
 }
 
+// Namespace for common static ENT logging functions
+namespace PDLogStatic
```
bossmc: It's weird to call these by two names.  Either the comment should say PDLog (and our docs should too) or this namespace should be called ENTLogStatic (or just ENTLog tbh).

bossmc:
Couple of comments, looks generally good.

include/pdlog.h
steelyvoid: Leaving as is following phone discussion

PR #476, opened 2016-03-16
[Reviewer: Rob] Alarm reliability

johadalin:
The main alarm changes. 
- Removed clear_all function
- Added get_alarm_state function, and associated enum
- altered communication_monitor alarm raising
- Alarms now start with severity pointing to NULL

I will link the pull request for cpp-common-test that covers these changes

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -110,7 +110,7 @@ public:
   void reraise_last_state();
   
   /// Returns the current state of the alarm as one of UNKNOWN, CLEARED, or ALARMED.
-  virtual AlarmCondition get_alarm_state();
+  virtual AlarmState::AlarmCondition get_alarm_state();
 
   // If an alarm is currently in a different state to the one we wish to raise
   // the alarm in, we raise the alarm and update _last_state_raised.
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -50,7 +50,7 @@
 /// @enum AlarmStatus
 ///
 /// Enum for the three possible alarm states
-enum Alarm
+enum AlarmStatus
 {
   // The NULL state in which all alarms start
   UNKNOWN,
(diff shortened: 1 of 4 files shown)


include/alarm.h
```diff
@@ -47,6 +47,19 @@
 #include "alarmdefinition.h"
 #include "eventq.h"
 
+/// @enum AlarmStatus
+///
+/// Enum for the three possible alarm states
+enum AlarmStatus
```
rkd-msw: Having AlarmState and AlarmStatus is kind of confusing.

include/alarm.h
```diff
@@ -47,6 +47,19 @@
...
+{
+  // The NULL state in which all alarms start
```
rkd-msw: Don't think "NULL" adds anything here - that's an internal detail. I might say that this state means that we haven't checked for the alarm yet so don't know if it should be active or not.

rkd-msw:
Looks good - bit concerned about the similar naming of AlarmState and AlarmStatus.

johadalin:
Moved AlarmStatus Enum to be within AlarmState, and renamed to AlarmCondition.
Other changes are to handle the new name.

include/alarm.h
```diff
@@ -83,6 +70,21 @@ class AlarmState
   std::string& get_issuer() {return _issuer;}
   std::string& get_identifier() {return _identifier;}
 
+  /// @enum AlarmCondition
+  ///
+  /// Enum for the three possible alarm states
```
rkd-msw: Maybe worth being more explicit that this is about *general* rather than *specific* states - raised or cleared rather than major/minor/critical.

include/alarm.h
```diff
@@ -83,6 +70,21 @@ class AlarmState
...
+  enum AlarmCondition
```
rkd-msw: I might just call it "Condition" (as you already have the "Alarm" prefix in "AlarmState") but don't worry if this is a pain to change.

rkd-msw:
LGTM.

PR #478, opened 2016-03-18
Reviewer EM: Add AVPException class

graemerobertson:
Moving this to diameterstack.cpp so we can use it in other places. I'll update everywhere that it's currently used too.


eleanor-merry:
LGTM

PR #491, opened 2016-03-30
Reviewer Rob - Common daemonize

bossmc:
This commonizes the `daemonize` function from Sprout (and adds the second fork() call to match our python version of this function).

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -49,8 +49,11 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <sys/file.h>
+#include <signal.h>
+#include <sys/stat.h>
 
 #include "utils.h"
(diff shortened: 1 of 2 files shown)


src/utils.cpp
```diff
@@ -377,6 +380,60 @@ int Utils::lock_and_write_pidfile(std::string filename)
...
+  // Create a new session to divorce the child from the tty of the parent.
+  if (setsid() == -1)
+  {
+    return errno;
+  }
+
+  // Clear any restricted umask
+  umask(0);
+
+  // Second fork
```
rkd-msw: Is it worth saying _why_ we need to fork twice?

rkd-msw:
+1

PR #493, opened 2016-04-07
[Reviewer: Alex] Implement get_time in Utils

rkd-msw:
We copy these three lines of code around a lot, so I've just made them a common function.


AlexHockey:
LGTM (though I wish we could just use http://en.cppreference.com/w/cpp/chrono/steady_clock/now)

PR #496, opened 2016-04-13
[Reviewer: Andrew] Allow running clearwater-snmp-handlers UT without /var/run/clearwater

rkd-msw:
Matching change to https://github.com/Metaswitch/clearwater-snmp-handlers/pull/137


AME2:
Looks good

PR #498, opened 2016-04-13
[Reviewer: Alex] Fix Clang warnings

rkd-msw:
Fixes two warnings I saw with Clang 3.8 - sending to you for review as a std::move expert.

* Two of the mockdiameterstack.hpp methods had different signatures to the parent methods. As they weren't used, I've just deleted them.
* Clang told me `moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]` and told me to remove the offending std::move call. I've done so and it seems OK.

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -325,10 +325,10 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
                 ce->domain.c_str(),
                 DnsRRecord::rrtype_to_string(ce->dnstype).c_str());
 
-      results.push_back(std::move(DnsResult(ce->domain,
+      results.push_back(DnsResult(ce->domain,
                                             ce->dnstype,
                                             ce->records,
-                                            ce->expires - time(NULL))));
+                                            ce->expires - time(NULL)));
     }
     else
     {
--- a/test_utils/mockdiameterstack.hpp
+++ b/test_utils/mockdiameterstack.hpp
@@ -44,8 +44,6 @@ class MockDiameterStack : public Diameter::Stack
 {
 public:
   MOCK_METHOD0(initialize, void());
-  MOCK_METHOD1(configure, void(const std::string&));
-  MOCK_METHOD1(advertize_application, void(const Diameter::Dictionary::Application&));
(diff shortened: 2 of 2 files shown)


src/dnscachedresolver.cpp
```diff
@@ -325,10 +325,10 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
                 ce->domain.c_str(),
                 DnsRRecord::rrtype_to_string(ce->dnstype).c_str());
 
-      results.push_back(std::move(DnsResult(ce->domain,
+      results.push_back(DnsResult(ce->domain,
```
bossmc: Shouldn't we be using `emplace_back` here anyway?

AlexHockey: We could do. I don't think it makes a difference here (but I could be wrong). `DnsResult` has a move constructor and the compiler will spot that `DnsResult(...)` is an rvalue and so can pass the value into `push_back` by doing a move rather than a copy.

AlexHockey: Hmmm. Having read around a bit (specifically http://stackoverflow.com/a/26861516) it looks like `emplace_back` may be faster after all (as it avoids the move entirely). I'm not sure its worth changing unless we have regression tests for this code.

AlexHockey:
LGTM. I'll leave it up to you whether to switch to `emplace_back`.

src/dnscachedresolver.cpp
rkd-msw: The warning is that using std::move prevents copy elision - without std::move, won't we elide the copy, and construct the DnsResult directly in-place without any moves or copies?

AlexHockey: It's not clear to me whether copy elision still involves a shallow copy of the object's memory or not. I'm also not sure whether `emplace_back` avoids that copy. But either way this feels like such a small difference we shouldn't worry too much about it.

bossmc: `emplace_back` uses placement new to guarantee no memory is copied (I say "guarantee" but technically the spec makes no guarantees at all... though it suggests that by default it will use placement new).

AlexHockey: Ok cool. So we should ideally switch to `emplace_back` but the gain and the risk are both so small I don't care too much about it.

PR #500, opened 2016-04-14
[Reviewer: Rob] Implement a table for counters indexed by IP address and time period

AlexHockey:
Hi Rob, please can you review this new stats table. This should match the review you reviewed. I've refactored some of the the existing `IPCountRow` code into a new `IPRow` (which actually wasn't as useful as I thought, but never mind), and implemented the new table. 

I've tested this in clearwater-fv-test. A separate PR is on its way.

I'm expecting this PR to break some other components at link-time (due to the new `snmp_ip_row.cpp`) but I'll fix those up without forcing you to review the changes.


rkd-msw:
LGTM.

PR #501, opened 2016-04-16
[Reviewer: Matt] Asynchronous DNS query

rkd-msw:
This implements the fix you suggested in https://github.com/Metaswitch/sprout/issues/1399#issuecomment-210478766.

I haven't done any live testing yet - I was thinking of writing a small tool that started 30 threads, all sharing a resolver, which query for cw-ngv.com and staging.cw-ngv.com (DNS TTL 60 seconds) once a second for ten minutes. I'd then:

* check that they never got an empty or otherwise incorrect result
* run that test program under Helgrind and memcheck
* check the logs and/or use gcov to make sure my new code is being hit

I might artificially slow down DNS responses if they come back too fast for me to test this change.


mirw:
Looks good.  A couple of comments.

*   It looks as though the request that triggers the cache expiry is still happening synchronously.  I think that's OK (1 request can't cause bursting), but just wanted to check you agree that this is the behavior?

*   I can't remember what DNS says about the TTL - are you allowed to continue to use cached entries after they've expired?  An alternative approach would be to actually do the query if the time to expiry is less than a positive value.

    Is there a limit on how long an expired cache entry can remain in the cache?  I'd guess it's one DNS query timeout.  Presumably this means we can still get bursting if a DNS server fails?  (I think that's just an extension of us rejecting requests if the DNS server fails for long enough, so I don't think that's a problem.)

    I think really what I'm saying is that there are 2 parameters here, both of which we might want to tune at some stage.

    *   How long before expiry do we start re-querying?
    *   How long after expiry do we stop returning the cached entry?

    Do you agree?  If so, let's discuss whether it's worth implementing this now - I have had questions about this behavior in the past from users.

rkd-msw:
Expired entries will remain in the cache indefinitely if the DNS server is down - that's existing function at https://github.com/Metaswitch/cpp-common/blob/3c0f9a297645f13a4567b047cec7804cff9a4211/src/dnscachedresolver.cpp#L560. I think that answers a couple of your questions (for example, not honouring the TTL in some cases is existing behaviour).

I don't think the request that triggers the cache expiry is happening synchronously - I think the only time a request happens synchronously is if there is no cached value, and that's unavoidable.

Happy to discuss tomorrow, both to check we're reading the code the same way and to agree whether this is a good thing to implement.

mirw:
I agree with your analysis.  Go ahead and test and merge.

rkd-msw:
I've tested - it worked fine under valgrind and helgrind. I found and fixed two functional issues:

* we create an empty cache entry on the first query (to store the fact that a query is pending), so I've added code to wait for a result if the cache entry contains no records
* at least one thread does need to call wait_for_replies, as this is what calls `ares_process_fd` to handle the DNS responses. So the request that triggers the cache expiry is now happening synchronously - you agreed above that this is OK because one request can't cause bursting.

rkd-msw:
Test code was:

```
#include "dnscachedresolver.h"
#include <cassert>
#include <unistd.h>
#include <thread>

static DnsCachedResolver* d;

void resolver_thread_function()
{
  for (int i = 0; i < 10; i++)
  {
    DnsResult r = d->dns_query("cw-ngv.com", ns_t_a);
    TRC_INFO("Got %ld records\n", r.records().size());
    DnsResult r2 = d->dns_query("staging.cw-ngv.com", ns_t_a);
    TRC_INFO("Got %ld records\n", r2.records().size());
    assert(r.records().size() > 0);
    assert(r2.records().size() > 0);
    sleep(60);
  }
}

int main()
{
  d = new DnsCachedResolver("8.8.8.8");
  std::vector<std::thread*> threads;
  for (int i = 0; i < 30; i++)
  {
    threads.push_back(new std::thread(resolver_thread_function));
  }

  for (std::thread* t: threads)
  {
    t->join();
    delete t;
  }

  delete d;
}
```

PR #502, opened 2016-04-19
Reviewer MIRW: Don't unregister callback handler in multiple places

graemerobertson:
This fixes https://github.com/Metaswitch/homestead/issues/328. By the time the Diameter stack is being shutdown, any applications that registered callback handlers should have unregistered them, so we shouldn't unregister them here - we do drop a warning log if we think there is a listener that hasn't unregistered its callback.

I've tested by running (with https://github.com/Metaswitch/homestead/pull/335) against an HSS and restarting Homestead. There is no crash and Homestead shuts down smoothly and quickly.


mirw:
Looks good.

PR #505, opened 2016-04-22
[Reviewer: Alex] Move thread creation out of constructor

### LINKED ISSUE #504: Cpp processes no longer do anything on a sighup/sigusr1
#### Symptoms
Running 'service <service> reload' should trigger the service to reload its configuration (e.g. cluster settings). Instead, it silently fails.

#### Impact
Elastic scaling doesn't work, re-synchronising data after a crash doesn't work

#### Release and environment 
This works in the Articuno release, and doesn't in the Charizard release

eleanor-merry:
This moves the thread creation out of the signal handler constructors

Fixes #504

### DIFF
--- a/include/communicationmonitor.h
+++ b/include/communicationmonitor.h
@@ -51,6 +51,8 @@
 /// one is detected, an alarm clear state is requested. Note that timing is
 /// driven by calls to the inform_* methods. As such the intervals will not
 /// be very precise at low call volume.
+///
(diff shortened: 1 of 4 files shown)


AlexHockey:
LGTM.

AlexHockey:
Hmmm, one comment actually. `Utils::start_signal_handlers` should have a comment saying you must not call this before you've daemonized.

PR #511, opened 2016-05-03
[Reviewer: Graeme] Don't put signal handler stuff in utils

eleanor-merry:
Don't make utils.cpp dependent on signal_handler.h


graemerobertson:
:frog:

PR #513, opened 2016-05-12
Reviewed: Add some MultiStateAlarm infrastructure

graemerobertson:
The main function in this PR is the ability to lock the HTTP stack.

I've also added a virtual destructor for MultiStateAlarm, and a MockMultiStateAlarm class.

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -212,6 +212,8 @@ class MultiStateAlarm: public BaseAlarm
 public:
   MultiStateAlarm(const std::string& issuer,
                   const int index);
+
+  virtual ~MultiStateAlarm() {}
   
 protected:
   /// These raise the alarm with the specified severity.
--- a/test_utils/mockalarm.h
+++ b/test_utils/mockalarm.h
@@ -51,4 +51,17 @@ public:
   MOCK_METHOD0(get_alarm_state, AlarmState::AlarmCondition());
 };
 
+class MockMultiStateAlarm : public MultiStateAlarm
+{
+public:
+  MockMultiStateAlarm() :
+    MultiStateAlarm("sprout", 0) {}
+
+  MOCK_METHOD0(set_indeterminate, void());
+  MOCK_METHOD0(set_warning, void());
+  MOCK_METHOD0(set_minor, void());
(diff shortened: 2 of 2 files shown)


rkd-msw:
I don't like the use of the "locked" terminology - it's confusing, and I initially thought it would be about mutexes. Maybe "disabled" and "enabled" instead?

src/httpstack.cpp
```diff
@@ -280,7 +280,33 @@ void HttpStack::handler_callback(evhtp_request_t* req,
...
+  if (_locked)
+  {
+    // The HTTP stack is currently locked. Return 503 with the body containing
+    // the _locked_reason_str if one has been provided.
+    TRC_DEBUG("HTTP stack is locked, reason: %s", _locked_reason_str.c_str());
+    if (!_locked_reason_str.empty())
+    {
+      request.add_content(_locked_reason_str.c_str());
+    }
+    send_reply_internal(request, 503, trail);
```
rkd-msw: Do we need to increment a stat here?

include/httpstack.h
```diff
@@ -431,6 +431,13 @@ class HttpStack
     }
   };
 
+  // Setters for _locked and _locked_reason_str.
+  virtual void set_locked(const bool locked) { _locked = locked; }
```
rkd-msw: Could you merge these two - `void disable(std::string reason_str)`?

include/httpstack.h
```diff
@@ -466,6 +473,14 @@ class HttpStack
...
 
   static bool _ev_using_pthreads;
+
+  // When _locked is set to true, the HTTP stack returns 503 to all
+  // requests, with the response body set to the value _locked_reason_str, if
+  // set.
+  //
+  // By default the HTTP stack is not locked.
+  std::atomic<bool> _locked;
+  std::string _locked_reason_str;
```
rkd-msw: This implies that there might be different reasons for disabling the stack. What happens if component A disables it for reason A, then component B disables it for reason B, then component A decides that reason A no longer applies? It feels like it should stay disabled for reason B, but that's not what happens.

PR #515, opened 2016-05-18
[Reviewer: Alex] Add SAS logging to the DNS Cached Resolver

steelyvoid:
This PR adds SAS logs to the DNS Cached Resolver so that the actual requests and responses from the DNS server are traced (as they are already for DNS-based ENUM in sprout).

Tested by making calls to non-existent and existing domains (verifying that NAPTR, SRV and A record lookups attempts are all traced) along with a test with the DNS server disabled (to check that the "no DNS servers responding" condition is traced).

Corresponding PRs in sprout and clearwater-sas-resources to follow.


src/diameterresolver.cpp
```diff
@@ -95,7 +95,7 @@ void DiameterResolver::resolve(const std::string& realm,
     // Realm is specified, so do a NAPTR lookup for the target.
     TRC_DEBUG("Do NAPTR look-up for %s", realm.c_str());
 
-    NAPTRReplacement* naptr = _naptr_cache->get(realm, ttl);
+    NAPTRReplacement* naptr = _naptr_cache->get(realm, ttl, 0);
```
AlexHockey: I think it's right to pass 0 in for the trail ID here (as diameter resolution happens off the call path so there is no trail to specify). However trail ID 0 is a valid trail ID in SAS, so I'm slightly concerned what might happen if we log all diameter resolution attempts to this trail ID. 

The resolution is just to wrap your new SAS events in `if (trail != 0)`.

AlexHockey:
Looks good - I only had one comment.

PR #516, opened 2016-05-20
[Reviewer: Rob] Add more detail to load monitor logs

### LINKED ISSUE #507: Load Monitor should SAS log and trace more detail
#### Symptoms

While investigating load monitor behavior, I noticed that

*   we don't SAS log when we accept a request - it would be useful to do so, recording the number of tokens left in the bucket and the refill rate
*   we "status" log to `/var/log/sprout/sprout*.txt` of the form `load_monitor.cpp:252: Maximum incoming request rate/second increased to 2662.384277 (based on a smoothed mean latency of 1785 and 0 upstream overload responses)` - it would be useful to include the number of requests in, and length of the previous period.

#### Impact

It's harder to establish the correct throttling tuning parameters.

#### Release and environment 

release-94

#### Steps to reproduce

Run under load and look at SAS log and /var/log/sprout/sprout*.txt.

eleanor-merry:
Adds more detail to the load monitor logs and adds SAS logs

Fixes #507, tested live.

### DIFF
--- a/src/load_monitor.cpp
+++ b/src/load_monitor.cpp
@@ -167,8 +167,19 @@ bool LoadMonitor::admit_request(SAS::TrailId trail)
   }
   else
   {
+    float accepted_percent = (accepted + rejected == 0) ?
+                             100.0 :
+                             100 * (((float) accepted) / (accepted + rejected));
+    timespec current_time;
+    clock_gettime(CLOCK_MONOTONIC_COARSE, &current_time);
+    unsigned long time_passed_ms = ((current_time.tv_sec * 1000) +
+                                    (current_time.tv_nsec / 1000000)) -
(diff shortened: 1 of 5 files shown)


src/load_monitor.cpp
```diff
@@ -162,6 +167,10 @@ bool LoadMonitor::admit_request()
   }
   else
   {
+    SAS::Event event(trail, SASEvent::LOAD_MONITOR_REJECTED_REQUEST, trail);
```
rkd-msw: Maybe worth calculating `int ms_passed = (current_time_ms - last_adjustment_time_ms);` and logging out how many requests we've accepted/rejected in the last N milliseconds? I'm not convinced but thought I'd suggest it.

rkd-msw:
:+1:

PR #518, opened 2016-05-23
[Reviewer: Andrew] Take lock when clearing alarm list

rkd-msw:
I think this fixes https://github.com/Metaswitch/cpp-common-test/issues/16 - I ran `LD_LIBRARY_PATH=../usr/lib valgrind --track-origins=yes --malloc-fill=cc --free-fill=df ../build/bin/cpp_common_test_test --gtest_filter="AlarmTest*"` ten times without this fix, and hit that issue four times; I then ran it ten times with this fix and didn't hit it.

I think you added this function originally - it's worth being aware that any time you access a data structure from multiple threads, you need to take a lock, even if you're clearing it - I suspect the issue was that one thread was trying to access elements of the vector that other threads had cleared, and corrupted some other memory when it did so.

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -140,7 +140,7 @@ public:
   bool _terminated;
   void register_alarm(BaseAlarm* alarm); 
   // Used to stop re-raising alarms in UTs
-  void forget_alarm_list(void) {_alarm_list.clear();}
+  void forget_alarm_list(void) {pthread_mutex_lock(&_lock); _alarm_list.clear(); pthread_mutex_unlock(&_lock); }
   void start_resending_alarms(void) { _first_alarm_raised = true; }
   void stop_resending_alarms(void) { _first_alarm_raised = false; }
 


include/alarm.h
```diff
@@ -140,7 +140,7 @@ class AlarmManager
   bool _terminated;
   void register_alarm(BaseAlarm* alarm); 
   // Used to stop re-raising alarms in UTs
-  void forget_alarm_list(void) {_alarm_list.clear();}
+  void forget_alarm_list(void) {pthread_mutex_lock(&_lock); _alarm_list.clear(); pthread_mutex_unlock(&_lock); }
```
AME2: Is it worth a comment saying the lock is already initialised in the constructor of the class?

AME2:
The code looks good, I've just suggested a potential comment.

include/alarm.h
rkd-msw: We don't have that sort of comment at other places we take this lock (https://github.com/Metaswitch/cpp-common/blob/8a18252e910914179a9cae3eee12726989148d49/src/alarm.cpp#L214), and I think it's normal/expected that member variables will be initialised in the constructor.

PR #524, opened 2016-06-08
[Reviewer: Graeme] Don't call fd_sess_getsid on NULL

rkd-msw:
`fd_msg_sess_get` can return 0, to indicate successful completion, but still set the session to NULL (see https://github.com/Metaswitch/freeDiameter/blob/33bdebe6674f3d98ce7b9d6e8d52db934914820d/libfdproto/messages.c#L1518). This PR checks for that case, and doesn't try to do anything with the session.

This caused https://github.com/Metaswitch/freeDiameter/issues/21 (although we no longer hit that issue, normally, as we've stopped SAS logging Device-Watchdog-Requests and therefore stopped going through this code path).

### DIFF
--- a/src/diameterstack.cpp
+++ b/src/diameterstack.cpp
@@ -957,7 +957,8 @@ void Stack::fd_sas_log_diameter_message(enum fd_hook_type type,
   struct session* sess;
   int dummy_is_new;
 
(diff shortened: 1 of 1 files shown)


graemerobertson:
Nice one, looks good.

Can we check that Diameter clients are also no longer affected by https://github.com/Metaswitch/freeDiameter/issues/21 before we close the issue? I know they're almost certainly not (given both the change you've pointed out and this PR) but worth checking.

PR #526, opened 2016-06-13
[Reviewer: Rob] Fix for alarm crashes

### LINKED ISSUE #521: Crash when re-raising alarms
#### Symptoms
I'm seeing Astaire crash reasonably regularly. The stack trace is:
```
#0  0x00007fd98afb0c37 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007fd98afb4028 in __GI_abort () at abort.c:89
#2  0x000000000042778a in signal_handler (sig=11) at main.cpp:211
#3  <signal handler called>
#4  __GI___pthread_mutex_lock (mutex=0x0) at ../nptl/pthread_mutex_lock.c:66
#5  0x0000000000426075 in push_noblock (item=<error reading variable: access outside bounds of object referenced via synthetic pointer>, this=0x0)
    at ../modules/cpp-common/include/eventq.h:239
#6  AlarmReqAgent::alarm_request (this=this@entry=0x63bfc0 <AlarmReqAgent::_instance>, req=std::vector of length 3, capacity 3 = {...})
    at ../modules/cpp-common/src/alarm.cpp:358
#7  0x0000000000426591 in AlarmState::issue (this=0x176bfc0) at ../modules/cpp-common/src/alarm.cpp:71
#8  0x00000000004269b2 in reraise_last_state (this=<optimized out>) at ../modules/cpp-common/src/alarm.cpp:103
#9  AlarmManager::reraise_alarms (this=0x176baa0) at ../modules/cpp-common/src/alarm.cpp:257
#10 0x0000000000426a99 in AlarmManager::reraise_alarms_function (data=<optimized out>) at ../modules/cpp-common/src/alarm.cpp:225
#11 0x00007fd98c0db184 in start_thread (arg=0x7fd989a25700) at pthread_create.c:312
#12 0x00007fd98b07437d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

#### Release and environment 
Seen in the Farfetch'd release.

eleanor-merry:
Rob, can you please review this change to the alarm classes.

This PR covers:
* Renaming the AlarmManager to the AlarmReRaiser
* Stopping the AlarmReqAgent and AlarmReRaiser from being singletons
* Adding a new AlarmManager that's responsible for the AlarmReqAgent and the AlarmManager (i.e. making sure they're created/deleted safely).
* Simplifying the re raising alarm function

This PR doesn't cover any change to the AlarmReqAgent/AlarmReRaiser (beyond simplifying the reraise function)

Fixes #521

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -66,7 +66,7 @@ class AlarmReqAgent
 public:
   enum
   {
-    MAX_Q_DEPTH = 100
+    MAX_Q_DEPTH = 1000
   };
 
   /// Queue an alarm request to be forwarded to snmpd.
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -69,14 +69,11 @@ public:
     MAX_Q_DEPTH = 100
   };
 
-  /// Initialize ZMQ context and start agent thread.
-  bool start();
-
-  /// Gracefully stop the agent thread and remove ZMQ context.
-  void stop();
-
   /// Queue an alarm request to be forwarded to snmpd.
   void alarm_request(std::vector<std::string> req);
+
+  // The AlarmManager is the only class allowed to create the AlarmReqAgent
+  friend class AlarmManager;
 
 private:
   AlarmReqAgent();
@@ -207,6 +204,9 @@ public:
   // Tell the Alarm Manager about an alarm
   void register_alarm(BaseAlarm* alarm); 
 
+  // The AlarmManager is the only class allowed to create the AlarmReRaiser
+  friend class AlarmManager;
+
 private:
   AlarmReRaiser();
   ~AlarmReRaiser();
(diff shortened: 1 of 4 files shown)


include/alarm.h
```diff
@@ -47,14 +47,74 @@
...
+/// Class which provides an agent thead to accept queued alarm requests from
+/// clients and forward them via ZMQ to snmpd (which will actually generate the
+/// inform message(s)).
+
+class AlarmReqAgent
+{
+public:
+  enum
+  {
+    MAX_Q_DEPTH = 100
```
rkd-msw: I realise this isn't new, but isn't it a bad idea? Won't we drop alarms? Should we set it to 10,000 or infinity or something like that?

rkd-msw:
Does this run fine under valgrind? (I faintly remember that the re-raising thread triggered too often when running slowly under valgrind, hence the 1,000-second pause in UT.)

Otherwise good, one comment about something you didn't introduce.

eleanor-merry:
Yes - the reraise thread is now fully controlled in UT (using MockPThreadCondVar, controlled time, and block_till_waiting).

include/alarm.h
eleanor-merry: I don't think this wants to be unlimited. I'll set it to 1000

PR #530, opened 2016-06-28
[Reviewer: Graeme] Add unregister_alarm call in alarm destructor

rkd-msw:
I think the crash you hit when we tried a version of this code last week might have been due to not saving off the `alarm_manager` constructor argument - that seems like a possible thing not to do, and if I comment out that line, I also hit a crash at the same point.

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -192,6 +192,9 @@ protected:
   // alarm has just been cleared this would be the corresponding _clear_state
   // for the alarm.
   AlarmState* _last_state_raised;
+
+  // The manager this alarm is registered with.
+  AlarmManager* _alarm_manager;
 };
 
 /// @class AlarmReRaiser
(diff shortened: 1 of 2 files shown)


include/alarm.h
```diff
@@ -192,6 +192,9 @@ class BaseAlarm
   // alarm has just been cleared this would be the corresponding _clear_state
   // for the alarm.
   AlarmState* _last_state_raised;
+
+  // The manager this alarm is registered with.
+  AlarmManager* _alarm_manager;
```
eleanor-merry: Can you add a comment to the alarm manager to say that it must be deleted after any alarms are deleted

include/alarm.h
```diff
@@ -204,6 +207,9 @@ class AlarmReRaiser
   // Tell the Alarm Manager about an alarm
   void register_alarm(BaseAlarm* alarm); 
 
+  // Tell the AlarmManager an alarm has been deleted
```
graemerobertson: Can you make AlarmManager/Alarm Manager consistent with the comment above? It would also be nice if the commenting style in this file was consistent but whatever.

graemerobertson:
I'm pretty sure I did save off the AlarmManager but whatever, if the tests pass then that's great and it looks like this will fix the issue. There are a couple of comments to address but then merge away.

PR #531, opened 2016-06-30
[Reviewer: Matt W] Allow daemons to provide path for stdout/stderr

richardwhiuk:
Small change to allow daemons to specify the path for stdout and stderr to be used when the daemon process is forked. Also change to append to these files instead of just blasting away the contents.

This is the first half of a fix to Metaswitch/sprout#965.


mirw:
Presumably it's OK to open `/dev/null` to append - it feels like it, but have you tested?

PR #532, opened 2016-07-01
[Reviewer: MIRW] Commonalize logging and daemonization

richardwhiuk:
There's a bunch of common logging and creation of a daemon which is replicated in every Clearwater server:

- [ ] chronos
- [ ] sprout
- [ ] ralf
- [ ] memento
- [ ] cedar
- [ ] homestead
- [ ] astaire
- [ ] stats-aggregator

This moves a small amount of the common code into cpp-common. It was the largest common set I could move without having to pass function calls to cpp-common to allow the server to hook functions in.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -542,3 +542,51 @@ int Utils::daemonize(std::string out, std::string err)
 
   return 0;
 }
+
+void Utils::daemon_log_setup(int argc,
+                             char* argv[],
+                             bool daemon,
+                             std::string& log_directory,
+                             int log_level,
+                             bool log_to_file)
+{
+  // Work out the program name from argv[0], stripping anything before the
+  // final slash.
+  char* prog_name = argv[0];
+  char* slash_ptr = rindex(argv[0], '/');
+  if (slash_ptr != NULL)
+  {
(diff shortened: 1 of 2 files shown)


mirw:
Looks good.  No comments.

src/utils.cpp
```diff
@@ -542,3 +542,51 @@ int Utils::daemonize(std::string out, std::string err)
...
+                                prefix + "_err.log");
+    }
+    else
+    {
+      errnum = Utils::daemonize();
+    }
+
+    if (errnum != 0)
+    {
+      TRC_ERROR("Failed to convert to daemon, %d (%s)", errnum, strerror(errnum));
```
eleanor-merry: I know this is a copy-paste of what we had previously - but surely this tries to log before we've set the logging functions up?

richardwhiuk: It does, but that's not as bad as it sounds, because we start with a statically initialised logger going to stdout. As we've failed to daemonize, that should end up in syslog I think.

mirw: I think where it ends up depends on the OS, but otherwise agreed.

PR #534, opened 2016-07-05
[Reviewer: Alex] Reference count rows of the IP time counter table

richardwhiuk:
Instead of ignoring add methods that occur when a row already exists,
keep track of them, and only remove a row when the same number have
occurred.

I've got an additional FV test in https://github.com/Metaswitch/clearwater-fv-test/pull/48

### DIFF
--- a/src/snmp_ip_time_based_counter_table.cpp
+++ b/src/snmp_ip_time_based_counter_table.cpp
@@ -131,17 +131,34 @@ public:
     // the write lock.
     pthread_rwlock_wrlock(&_counters_lock);
 
-    std::map<std::string, IPEntry*>::iterator entry = _counters_by_ip.find(ip);
-    if (entry == _counters_by_ip.end())
-    {
-      // IP address does not already exist - add an entry in the counts map and
-      // the associated SNMP rows.
-      TRC_DEBUG("Adding IP rows for: %s", ip.c_str());
+    std::map<std::string, uint32_t>::iterator ref_entry = _ref_count_by_ip.find(ip);
 
-      _counters_by_ip[ip] = new IPEntry();
-      add(std::make_pair(ip, TimePeriodIndexes::scopePrevious5SecondPeriod));
-      add(std::make_pair(ip, TimePeriodIndexes::scopeCurrent5MinutePeriod));
-      add(std::make_pair(ip, TimePeriodIndexes::scopePrevious5MinutePeriod));
+    if (ref_entry == _ref_count_by_ip.end())
+    {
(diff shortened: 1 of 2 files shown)


src/snmp_ip_time_based_counter_table.cpp
```diff
@@ -153,18 +170,41 @@ class IPTimeBasedCounterTableImpl : public IPTimeBasedCounterTable,
...
+    }
+    else
+    {
+      ref_entry->second --;
+
+      // If we have removed the last reference to this entry, remove it from the
+      // counts table.
+      if (ref_entry->second == 0)
+      {
+
```
AlexHockey: Unnecessary blank line

src/snmp_ip_time_based_counter_table.cpp
```diff
@@ -281,6 +321,10 @@ class IPTimeBasedCounterTableImpl : public IPTimeBasedCounterTable,
   // multiple threads so is protected by a lock.
   std::map<std::string, IPEntry*> _counters_by_ip;
   pthread_rwlock_t _counters_lock;
+
+  // A reference count for each IP address, keeping track of how many times
+  // it's been added and removed. This is protected by _counters_lock
```
AlexHockey: Feels like we should rename `_counters_lock` to just `_table_lock` or similar.

src/snmp_ip_time_based_counter_table.cpp
```diff
@@ -131,17 +131,34 @@ class IPTimeBasedCounterTableImpl : public IPTimeBasedCounterTable,
...
+        TRC_DEBUG("Adding IP rows for: %s", ip.c_str());
+
+        _counters_by_ip[ip] = new IPEntry();
+        add(std::make_pair(ip, TimePeriodIndexes::scopePrevious5SecondPeriod));
+        add(std::make_pair(ip, TimePeriodIndexes::scopeCurrent5MinutePeriod));
+        add(std::make_pair(ip, TimePeriodIndexes::scopePrevious5MinutePeriod));
+      }
+      else
+      {
+        TRC_ERROR("Entry for %s doesn't exist in reference table, but does exist in count table",
```
AlexHockey: Feels more like a `TRC_WARNING` (given the only impact of hitting this is your stats might be a bit wrong).

AlexHockey:
I've reviewed. Only a few minor comments. 

Personally I would have been tempted to put the refernce count in the `IPEntry` object so that we don't have two maps that could get out of sync. However I don't care too strongly.

PR #537, opened 2016-07-18
Reviewer: [AJH] HTTP Graylisting 

TJW2:


include/baseresolver.h
```diff
@@ -98,8 +98,16 @@ class BaseResolver
...
+  {
+    blacklist(ai, _default_blacklist_duration, _default_graylist_duration);
+  }
+  void blacklist(const AddrInfo& ai, int blacklist_ttl)
+  {
+    blacklist(ai, blacklist_ttl, _default_graylist_duration);
+  }
+  void blacklist(const AddrInfo& ai, int blacklist_ttl, int graylist_ttl);
+  /// Returns false only if the associated Host has state State::WHITE
   bool blacklisted(const AddrInfo& ai);
```
AlexHockey: I thought we were going to make this method private?

TJW2: That's correct, I'll make that change. Do you think it's worth moving all the blacklist-related methods to the bottom of the header file, to keep them together, rather than having some at the top and some at the bottom?

include/baseresolver.h
```diff
@@ -98,8 +98,16 @@ class BaseResolver
...
+  void blacklist(const AddrInfo& ai)
+  {
+    blacklist(ai, _default_blacklist_duration, _default_graylist_duration);
+  }
+  void blacklist(const AddrInfo& ai, int blacklist_ttl)
+  {
+    blacklist(ai, blacklist_ttl, _default_graylist_duration);
+  }
+  void blacklist(const AddrInfo& ai, int blacklist_ttl, int graylist_ttl);
+  /// Returns false only if the associated Host has state State::WHITE
```
AlexHockey: We normally have a blank line before a comment unless the comment is the first line in a block. This is because the comment normally refers to the code that follows it, so having a preceding blank line makes it clear the comment is separate from the code that goes before. 

It's sometime appropriate to have a comment follow on directly from some code, but those instances are much less common.

TJW2: I thought the same thing, and made the same change just after I pushed this - I didn't think it was worth pushing again!

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
...
+    /// Destructor
+    ~Host();
+
+    /// Enum representing the current state of a Host in the blacklist
+    /// system. Whitelisted, graylisted and not being probed, graylisted and
+    /// being probed, and blacklisted respectively.
+    enum struct State {WHITE, GRAY_NOT_PROBING, GRAY_PROBING, BLACK};
+    /// Returns a string representation of the given state.
+    static std::string state_to_string(State state);
+    /// Returns the current state of a Host.
```
AlexHockey: This is a member method, so it returns the current state of *this* host.

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
...
+    State get_state(time_t current_time);
+
+    /// Indicates that the combination associated with this Host has been
```
AlexHockey: You've already mentioned that that host is associated with a specific IP/transport/port combination. No need to repeat yourself.

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
   /// combinations which have been blacklisted because the destination is
   /// unresponsive (either TCP connection attempts are failing or a UDP
   /// destination is unreachable).
-  pthread_mutex_t _blacklist_lock;
-  typedef std::map<AddrInfo, time_t> Blacklist;
-  Blacklist _blacklist;
+
+  /// Private class to hold data and methods associated to a transport/IP
```
AlexHockey: It might be more natural to phrase this as "IP/transport/port" (port is associated with the transport, so put those closer together).

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
...
+  /// address/port combination in the blacklist system. Each Host is associated
+  /// with exactly one combination.
+  class Host
+  {
+  public:
+    /// Constructor
```
AlexHockey: It would be good to mention that these are in units of seconds.

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
...
+    /// Indicates that the combination associated with this Host is being probed
+    /// by the given user.
+    void probing(pthread_t user_id);
+    /// Indicates that the combination associated with this Host has gone
+    /// untested by the given user.
+    void untested(pthread_t user_id);
+
+  private:
+    /// The time at which this Host is to be removed from the blacklist and
+    /// placed onto the graylist.
```
AlexHockey: Might be worth specifying that this is in units of "seconds since the epoch" (which means seconds since the start of 1970).

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
...
+  /// in the blacklist system, and Host::State::WHITE otherwise. _hosts_lock
+  /// must be held when calling this method.
+  Host::State host_state(const AddrInfo& ai) {return host_state(ai, time(NULL));}
+  Host::State host_state(const AddrInfo& ai, time_t current_time);
+
+public:
+  /// Indicates that the given AddrInfo has responded.
+  void success(const AddrInfo& ai);
+  /// Indicates that the calling thread has left the given AddrInfo untested.
+  void untested(const AddrInfo& ai);
```
AlexHockey: I would put these up with the `blacklist` method.

include/baseresolver.h
```diff
@@ -241,10 +254,73 @@ class BaseResolver
   /// combinations which have been blacklisted because the destination is
   /// unresponsive (either TCP connection attempts are failing or a UDP
   /// destination is unreachable).
```
AlexHockey: It would be good to update this comment to refer to hosts rather than the blacklist.

src/baseresolver.cpp
```diff
@@ -466,42 +492,31 @@ IP46Address BaseResolver::to_ip46(const DnsRRecord* rr)
 }
 
 /// Adds an address, port, transport tuple to the blacklist.
-void BaseResolver::blacklist(const AddrInfo& ai, int ttl)
+void BaseResolver::blacklist(const AddrInfo& ai,
+                             int blacklist_ttl, int graylist_ttl)
```
AlexHockey: It's clearer if you put the two parameters on different lines.

src/baseresolver.cpp
```diff
@@ -466,42 +492,31 @@ IP46Address BaseResolver::to_ip46(const DnsRRecord* rr)
...
-    {
-      // Blacklist entry has yet to expire.
-      rc = true;
-    }
-    else
-    {
-      // Blacklist entry has expired, so remove it.
-      _blacklist.erase(i);
-    }
+    rc = true;
```
AlexHockey: You could just do `rc = (host_state(ai) == Host::State::BLACK)`

src/baseresolver.cpp
```diff
@@ -466,42 +492,31 @@ IP46Address BaseResolver::to_ip46(const DnsRRecord* rr)
...
   }
-  pthread_mutex_unlock(&_blacklist_lock);
-  char buf[100];
-  TRC_DEBUG("%s:%d transport %d is %sblacklisted",
-            inet_ntop(ai.address.af, &ai.address.addr, buf, sizeof(buf)),
-            ai.port, ai.transport, rc ? "" : "not ");
```
AlexHockey: You've removed a log here. This is fine if the log is not useful, but it's not clear to me that is the case here.

src/baseresolver.cpp
```diff
@@ -870,3 +885,167 @@ int BaseResolver::SRVWeightedSelector::total_weight()
 {
   return _tree[0];
 }
+
+BaseResolver::Host::Host(int blacklist_ttl,int graylist_ttl) :
+  _blacklist_expiry_time(time(NULL) + blacklist_ttl),
+  _graylist_expiry_time(time(NULL) + blacklist_ttl + graylist_ttl),
```
AlexHockey: You call `time` twice here, which could return two different results. I would prefer if you moved this initialization to the constructor body and only call `time` once.

include/baseresolver.h
TJW2: We had a problem here before with requiring the Host::State enum, which is defined in Host towards the end of the file. It may be possible to pre-declare the enum, or perhaps move the definition of Host.

src/baseresolver.cpp
AlexHockey: A reason this is important is that if `graylist_ttl` is 0 we don;t want to do graylisting, but this code could mean `_graylist_expiry_time` ends up being 1s after `_blacklist_expiry_time` even if `graylist_ttl` is 0.

src/baseresolver.cpp
TJW2: The host_state function now provides a log specifying the state of the AddrInfo in the blacklist system. I removed the log here to avoid duplication - would it be worth putting it back?

src/baseresolver.cpp
```diff
@@ -870,3 +885,167 @@ int BaseResolver::SRVWeightedSelector::total_weight()
...
+  if (get_state(time(NULL)) == State::GRAY_NOT_PROBING)
+  {
+    _being_probed = true;
+    _probing_user_id = user_id;
+  }
+}
+
+void BaseResolver::Host::untested(pthread_t user_id)
+{
+  if (get_state(time(NULL)) == State::GRAY_PROBING &&
```
AlexHockey: We normally put brackets around each part of a conditional to avoid issues with operator precedence. SO this would look like:

```cpp
if ((get_state(time(NULL)) == State::GRAY_PROBING) && (user_id == _probing_user_id))
{
  ...
```

src/baseresolver.cpp
```diff
@@ -870,3 +885,167 @@ int BaseResolver::SRVWeightedSelector::total_weight()
...
+  }
+  else
+  {
+    return State::WHITE;
+  }
+}
+
+void BaseResolver::Host::success()
+{
+  if (get_state(time(NULL)) == State::GRAY_PROBING)
```
AlexHockey: It feels wrong to be testing for the probing state but not also testing for the user ID.

src/baseresolver.cpp
```diff
@@ -870,3 +885,167 @@ int BaseResolver::SRVWeightedSelector::total_weight()
...
+    i->second.success();
+  }
+
+  pthread_mutex_unlock(&_hosts_lock);
+}
+
+void BaseResolver::probing(const AddrInfo& ai)
+{
+  char buf[100];
+  TRC_DEBUG("Probing %s:%d transport %d",
```
AlexHockey: Well, we're not actually probing it here. Maybe say "Selected host ... for probing".

src/baseresolver.cpp
TJW2: I think we agreed initially that we would not test for user ID on a success call, as provided the host has been successfully connected to, we don't mind who actually did it. This is easy enough to change though, if you'd like.

src/baseresolver.cpp
```diff
@@ -870,3 +885,167 @@ int BaseResolver::SRVWeightedSelector::total_weight()
...
+  {
+    i->second.probing(pthread_self());
+  }
+}
+
+void BaseResolver::untested(const AddrInfo& ai)
+{
+  char buf[100];
+  TRC_DEBUG("%s:%d transport %d returned untested",
+            inet_ntop(ai.address.af, &ai.address.addr, buf, sizeof(buf)),
```
AlexHockey: I think we should refactor out the code that formats an AddrInfo for logging. Perhaps it should live in the `AddrInfo` class?

TJW2: I agree, especially given this code is repeated several times with the additional logging I have added.

src/httpconnection.cpp
```diff
@@ -653,6 +654,7 @@ HTTPCode HttpConnection::send_request(const std::string& path,                 /
       }
 
       // Success!
+      _resolver->success(*i);
```
AlexHockey: This code looks correct. However I think whenever we talk to a host we should flag it as either blacklisted or successful, so I think you also need some code ~15 lines below.

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
+  time_t current_time = time(NULL);
 
-  for (std::vector<DnsRRecord*>::const_iterator i = result.records().begin();
-       i != result.records().end();
-       ++i)
+  pthread_mutex_lock(&_hosts_lock);
+
+  // Loop through the records in the result to select a GRAY_NOT_PROBING target,
+  // if it exists
+  std::vector<DnsRRecord*>::const_iterator j;
```
AlexHockey: There is usually a better way to name an iterator that just "i". Here, `result_to_probe` would probably be clearer.

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
+  for (j = result.records().begin(); j != result.records().end(); ++j)
   {
-    ai.address = to_ip46(*i);
-    if (!blacklisted(ai))
+    ai.address = to_ip46(*j);
+    if (host_state(ai, current_time) == Host::State::GRAY_NOT_PROBING)
     {
-      // Address isn't blacklisted, so copy across to the target list.
+      TRC_DEBUG("Added a graylisted target for probing");
```
AlexHockey: This debug log probably doesn't add much over the log that `probing` produces.

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
   std::string added_from_blacklist_str;
+  time_t current_time = time(NULL);
 
-  for (std::vector<DnsRRecord*>::const_iterator i = result.records().begin();
-       i != result.records().end();
-       ++i)
+  pthread_mutex_lock(&_hosts_lock);
+
+  // Loop through the records in the result to select a GRAY_NOT_PROBING target,
+  // if it exists
```
AlexHockey: Comments should explain why you're doing something, not what you're doing. Here it would be better to say something like "If possible we want the first returned target to be a graylisted target that is not being probed. See if one exists and if so add it to the returned targets."

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
-      blacklist_str = blacklist_str + (*i)->to_string() + ";";
+      break;
     }
+  }
 
-    if (targets.size() >= (size_t)retries)
-    {
-      // We have enough targets so stop looking at records.
-      TRC_DEBUG("Have enough targets");
+  pthread_mutex_unlock(&_hosts_lock);
```
AlexHockey: I don't think it's safe to unlock the mutex here.

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
 
-      if (trail != 0)
+  // Loop through the records in the result picking non-blacklisted targets.
+  for (std::vector<DnsRRecord*>::const_iterator i = result.records().begin();
+       i != result.records().end();
+       ++i)
+  {
+    if (i != j)
```
AlexHockey: I think you should also rename the `i` iterator to `curr_result` or similar. This would make this if test `if (curr_result != result_to_probe)` which is much clearer.

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
+    {
+      ai.address = to_ip46(*i);
+      if (host_state(ai, current_time) == Host::State::WHITE)
+      {
+        // Address isn't blacklisted, so copy across to the target list.
```
AlexHockey: Well, it's not blacklisted or graylisted. Maybe just "host is healthy"?

src/baseresolver.cpp
```diff
@@ -362,49 +365,72 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
-    if (!blacklisted(ai))
+    ai.address = to_ip46(*j);
+    if (host_state(ai, current_time) == Host::State::GRAY_NOT_PROBING)
     {
-      // Address isn't blacklisted, so copy across to the target list.
+      TRC_DEBUG("Added a graylisted target for probing");
+      probing(ai);
       targets.push_back(ai);
-      targetlist_str = targetlist_str + (*i)->to_string() + ";";
-      TRC_DEBUG("Added a server, now have %ld of %d", targets.size(), retries);
```
AlexHockey: I think we need to include these two lines of logging.

include/baseresolver.h
AlexHockey: My main concern is keeping all public methods together, and all private methods together. However if several methods have the same visibility and similar function, they should be close to each other.

include/baseresolver.h
AlexHockey: None of these methods depend on the `Host` class at all so they should be safe to move up.

src/baseresolver.cpp
AlexHockey: Ah, that's OK then.

src/baseresolver.cpp
AlexHockey: If we don't care who's calling success, why do we care whether the host is being probed or not? Maybe we should just make this test "is the host non-black".

src/baseresolver.cpp
```diff
@@ -870,3 +885,167 @@ int BaseResolver::SRVWeightedSelector::total_weight()
...
+  case State::GRAY_PROBING:
+    return "GRAY_PROBING";
+  case State::BLACK:
+    return "BLACK";
+  default:
+    return "UNKNOWN";
+  }
+}
+
+BaseResolver::Host::State BaseResolver::Host::get_state(time_t current_time)
```
AlexHockey: You call this method a lot like this: `get_state(time(NULL))`. It would be cleaner to have an override that takes no parameters, and gets the time for you.

AlexHockey:
I've finished reviewing. This looked pretty good. I don't actually have much feedback on the code itself - nice job! I had more feedback on the commenting but this is totally expected.

One piece of feedback that doesn't fit in anywhere else: I think we should have a comment explaining how graylisting works and what sequence of method calls could happen (e.g. `blacklist` followed by `success`/`blacklist`/`untested`). I would probably put this is a comment just before the definition of the `success`/`blacklist`/`untested` methods. 

Please let me know when you've addressed my feedback, and I'll take another look.

include/baseresolver.h
TJW2: I've rearranged the code, and now public and protected methods are kept together.

include/baseresolver.h
TJW2: I've updated this comment.

include/baseresolver.h
TJW2: I've rephrased this to be clearer.

include/baseresolver.h
TJW2: I've changed this here, and also in other comments.

include/baseresolver.h
TJW2: I've expanded the constructor comment to be more specific.

include/baseresolver.h
TJW2: I've added to the comment here to make it clearer.

include/baseresolver.h
TJW2: I've moved these methods up.

include/baseresolver.h
TJW2: I've updated this comment.

src/baseresolver.cpp
TJW2: I've changed this as you suggested.

src/baseresolver.cpp
TJW2: I've changed this to your suggestion.

src/baseresolver.cpp
TJW2: Good point - I've changed the code so that these times are set up in the constructor, with the same time value.

src/baseresolver.cpp
TJW2: I've changed this as you suggest.

src/baseresolver.cpp
TJW2: Good point, I've made this change.

src/baseresolver.cpp
TJW2: I've renamed j to result_to_probe.

src/baseresolver.cpp
TJW2: I've removed this log.

src/baseresolver.cpp
TJW2: I've changed this comment as you suggest.

src/baseresolver.cpp
TJW2: I've moved the mutex to beyond the second loop in this method.

src/baseresolver.cpp
TJW2: I've renamed i to current_result.

src/baseresolver.cpp
TJW2: I've changed this, and also renamed variables in this section to use 'healthy' and 'unhealthy'.

src/baseresolver.cpp
TJW2: This is the newly-created first loop of a_resolve, so I'm not sure why Git is showing this up here. Regardless, there isn't logging where there should be, so I've added some.

src/baseresolver.cpp
TJW2: I've changed this as you suggested.

src/baseresolver.cpp
TJW2: I've renamed this method 'select_for_probing', and the 'probing' method on host to 'selected_for_probing'. I've also updated some comments to reflect this change.

include/baseresolver.h
```diff
@@ -85,6 +86,22 @@ struct AddrInfo
...
   }
+
+  std::string address_and_port_to_string() const
+  {
+    std::stringstream os;
+    char buf[100];
+    os << inet_ntop(address.af, &address.addr, buf, sizeof(buf));
+    os << ":" << port;
+    return os.str();
+  }
```
AlexHockey: Nice one.

include/baseresolver.h
```diff
@@ -98,9 +115,20 @@ class BaseResolver
...
+  void blacklist(const AddrInfo& ai, int blacklist_ttl)
+  {
+    blacklist(ai, blacklist_ttl, _default_graylist_duration);
+  }
+  void blacklist(const AddrInfo& ai, int blacklist_ttl, int graylist_ttl);
+
+  /// Indicates that the given AddrInfo has responded.
+  void success(const AddrInfo& ai);
+  /// Indicates that the calling thread has left the given AddrInfo untested.
```
AlexHockey: Blank line before this comment please.

include/baseresolver.h
```diff
@@ -237,14 +270,96 @@ class BaseResolver
...
+  /// graylist.
+  ///
+  /// Hosts on the graylist are given out by the a_resolve method to only one
+  /// client, unless insufficient non-blacklisted hosts are available. A host
+  /// moves to the whitelist if the client probing this host connects
+  /// successfully, or if a specified time elapses. If a client given a host for
+  /// probing and does not attempt to connect to it, it is made available for
+  /// giving out by a_resolve once more. If a client attempts but fails to
+  /// connect to a host, it is moved to the blacklist.
+  ///
```
AlexHockey: I think the last paragraph is logically distinct from the rest of the comment, so there should be a blank line here.

src/baseresolver.cpp
```diff
@@ -264,10 +267,7 @@ void BaseResolver::srv_resolve(const std::string& srv_name,
             ai.address = active_addr[ii].back();
             active_addr[ii].pop_back();
             targets.push_back(ai);
-            char buf[100];
-            std::string target = inet_ntop(ai.address.af,
-                                           &ai.address.addr,
-                                           buf, sizeof(buf));
+            std::string target = ai.address_and_port_to_string();
             std::string tg = "[" + target + ":" + std::to_string(ai.port) + "] ";
```
AlexHockey: I think you end up adding the port twice - once in `address_and_port_to_string` and once in the line above. There are a few other examples of this - please could you fix them up?

src/baseresolver.cpp
```diff
@@ -362,73 +357,96 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
     {
-      targets.push_back(blacklisted_targets[ii]);
-      char buf[100];
-      std::string blacklistee = inet_ntop(blacklisted_targets[ii].address.af,
-                                          &blacklisted_targets[ii].address.addr,
-                                          buf, sizeof(buf));
-      std::string bl = "[" + blacklistee + ":" + std::to_string(blacklisted_targets[ii].port) + "]";
-      added_from_blacklist_str = added_from_blacklist_str + bl;
+      targets.push_back(unhealthy_targets[ii]);
+      std::string blacklistee = unhealthy_targets[ii].address_and_port_to_string();
```
AlexHockey: You've changed the other variables to use the term unhealthy (which is fine) but left this as blacklistee, which is a bit inconsistent.

AlexHockey:
I've reviewed. Again, these changes are generally very good. A few cosmetic things, plus a minor logging bug (but one you should still fix up).

PR #540, opened 2016-07-27
[Reviewer: RKD] Add ability to force error in local store set

sathiyan-sivathas:
So that I could write the UT in the other pull request.

### DIFF
--- a/src/localstore.cpp
+++ b/src/localstore.cpp
@@ -51,6 +51,7 @@ LocalStore::LocalStore() :
   _data_contention_flag(false),
   _db_lock(PTHREAD_MUTEX_INITIALIZER),
   _db(),
+  _force_error_flag(false),
   _old_db()
 {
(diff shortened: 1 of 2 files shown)


src/localstore.cpp
```diff
@@ -151,6 +159,17 @@ Store::Status LocalStore::set_data(const std::string& table,
 
   Store::Status status = Store::Status::DATA_CONTENTION;
 
+  // This is for the purpose of testing data SETs failing.  If the flag is set
+  // to true, then we'll just return an error.
+  if (_force_error_flag == true)
+  {
+    TRC_DEBUG("Force an error on the SET");
+    _force_error_flag = false;
+    status = Store::Status::ERROR;
```
rkd-msw: You could just `return Store::Status::ERROR` - that's a bit more concise.

rkd-msw:
Looks good, one comment.

PR #541, opened 2016-07-27
[Reviewer: Ellie] Adds support for enterprise alarms MIB

AME2:
The big changes here include:

- Updating the JSON parser to read the new fields, do some validity checking on them and then assign them to the new member variables of the SeverityDetails and AlarmDefinition structs.
- Adding methods to the AlarmTableDef class to retrieve the new values in the above structs (that are part of the AlarmTableDef object) and tests these get parsed correctly by the JSON parser. https://github.com/Metaswitch/clearwater-snmp-handlers/pull/157
- Testing that the JSON parser rejects JSON fields that are too long. https://github.com/Metaswitch/cpp-common-test/pull/26
- Updating the Python JSON parser to test that the new enterprise MIB fields are under 4096 characters. https://github.com/Metaswitch/python-common/pull/64

Notes:

- The way I've done the alarm name processing (SPROUT_PROCESS_FAILURE -> Sprout process failure) means that the name of each alarm will start with a capital letter. This is good for most alarms but not for the etcd ones. My opinion is that it's not worth the hassle to change it.

### DIFF
--- a/include/json_alarms.h
+++ b/include/json_alarms.h
@@ -75,9 +75,13 @@ namespace JSONAlarms
                                  std::string& error,
                                  std::vector<AlarmDef::AlarmDefinition>& alarms);
 
+  // Processes the name of an alarm into a human readable format, for example
+  // "SPROUT_PROCESS_FAILURE" becomes "Sprout process failure"
   std::string process_alarm_name(std::string raw_name);
 
-  std::string prepare_error_message(std::string field, int max_length, int index);
+  // Prepares a string that can be used to report the error that a specific
+  // field has exceeded the character limit
+  std::string exceeded_character_limit_error(std::string field, int max_length, int index);
 
   // Writes a header file that includes the alarm IDs and their index
   void write_header_file(std::string name, std::map<std::string, int> alarms);
--- a/include/json_alarms.h
+++ b/include/json_alarms.h
@@ -77,6 +77,8 @@ namespace JSONAlarms
 
   std::string process_alarm_name(std::string raw_name);
 
+  std::string prepare_error_message(std::string field, int max_length, int index);
+
   // Writes a header file that includes the alarm IDs and their index
   void write_header_file(std::string name, std::map<std::string, int> alarms);
 };
--- a/include/json_alarms.h
+++ b/include/json_alarms.h
@@ -75,6 +75,8 @@ namespace JSONAlarms
                                  std::string& error,
                                  std::vector<AlarmDef::AlarmDefinition>& alarms);
 
+  std::string process_alarm_name(std::string raw_name);
+
   // Writes a header file that includes the alarm IDs and their index
   void write_header_file(std::string name, std::map<std::string, int> alarms);
 };
--- a/src/json_alarms.cpp
+++ b/src/json_alarms.cpp
@@ -180,9 +180,9 @@ namespace JSONAlarms
 
           // We check if any extended details have been included in each
           // alarms's JSON file.
-          if (alarms_def_it->HasMember("extended details"))
+          if (alarms_def_it->HasMember("extended_details"))
           {
-            JSON_GET_STRING_MEMBER(*alarms_def_it, "extended details", extended_details);
+            JSON_GET_STRING_MEMBER(*alarms_def_it, "extended_details", extended_details);
             if (extended_details.length() > 4096)
             {
               char error_text[100];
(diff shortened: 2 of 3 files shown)


src/json_alarms.cpp
```diff
@@ -174,6 +178,26 @@ namespace JSONAlarms
             return false;
           }
 
+          // We check if any extended details have been included in each
+          // alarms's JSON file.
+          if (alarms_def_it->HasMember("extended details"))
```
eleanor-merry: Spaces? extended_details surely?

src/json_alarms.cpp
```diff
@@ -35,6 +35,8 @@
  */
 
 #include "json_alarms.h"
+#include <algorithm>
+#include <string>
```
eleanor-merry: Why have you needed to add this?

src/json_alarms.cpp
```diff
@@ -182,21 +206,65 @@ namespace JSONAlarms
             error = std::string(error_text);
             return false;
           }
+          
+          // We check if a extended description have been included in each
```
eleanor-merry: *an extended description has

src/json_alarms.cpp
```diff
@@ -182,21 +206,65 @@ namespace JSONAlarms
...
+          // alarms's JSON file.
+          if (alarms_def_it->HasMember("extended description"))
```
eleanor-merry: Same spaces comment

src/json_alarms.cpp
```diff
@@ -213,7 +281,15 @@ namespace JSONAlarms
         }
         else 
         {
-          AlarmDef::AlarmDefinition ad = {index,
+          // Process the raw_name of each alarm (e.g. SPROUT_PROCESS_FAILURE)
```
eleanor-merry: This should be put into a new function - let's discuss tomorrow

src/json_alarms.cpp
```diff
@@ -174,6 +178,26 @@ namespace JSONAlarms
...
           }
 
+          // We check if any extended details have been included in each
+          // alarms's JSON file.
+          if (alarms_def_it->HasMember("extended details"))
+          {
+            JSON_GET_STRING_MEMBER(*alarms_def_it, "extended details", extended_details);
+            if (extended_details.length() > 4096)
+            {
+              char error_text[100];
```
eleanor-merry: I'm not a big fan of this logging method - let's discuss tomorrow

eleanor-merry:
Generally looks good, some comments inline

src/json_alarms.cpp
AME2: Yes I reread the spec and it did specify underscores instead of spaces. I'll change this.

src/json_alarms.cpp
AME2: Will do, I've had a search and the only other way I can see of converting a string to lowercase is iterating over the string character by character which I don't think is any better.

include/json_alarms.h
```diff
@@ -75,6 +75,10 @@ namespace JSONAlarms
                                  std::string& error,
                                  std::vector<AlarmDef::AlarmDefinition>& alarms);
 
+  std::string process_alarm_name(std::string raw_name);
```
eleanor-merry: Can you please add function comments for these?

src/json_alarms.cpp
```diff
@@ -168,35 +171,85 @@ namespace JSONAlarms
           JSON_GET_STRING_MEMBER(*alarms_def_it, "details", details);
           if (details.length() > 255)
           {
-            char error_text[100];
-            sprintf(error_text, "alarm %d: 'details' exceeds %d char limit", index, 255);
-            error = std::string(error_text);
+            error = prepare_error_message("details", 255, index);
```
eleanor-merry: This is a very generic function name for a function that creates a very specific error message - can you change the name?

include/json_alarms.h
eleanor-merry: name not raw_name?

src/json_alarms.cpp
```diff
@@ -168,35 +171,85 @@ namespace JSONAlarms
...
           if (description.length() > 255)
           {
-            char error_text[100];
-            sprintf(error_text, "alarm %d: 'description' exceeds %d char limit", index, 255);
-            error = std::string(error_text);
+            error = prepare_error_message("description", 255, index);
             return false;
           }
+          
+          // We check if an extended description have been included in each
```
eleanor-merry: *has

eleanor-merry:
Generally looks good, some comments inline

PR #546, opened 2016-08-17
[Reviewer: AJH] Add Connection Pooling to MemcachedStore

TJW2:
Refactors out connection pooling code from the TopologyNeutralMemcachedStore class into new template connection pooling classes suitable for various types of connection, and integrates these new classes back into TopologyNeutralMemcachedStore.

### DIFF
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -44,6 +44,8 @@
 #include <time.h>
 
 #include "log.h"
+
+// Required as AddrInfo is defined here
 #include "baseresolver.h"
 
 template <typename T>
@@ -71,11 +73,15 @@ struct ConnectionInfo
 };
 
 /// Abstract template class storing a pool of connection objects, in "slots",
-/// with each distinct target having its own slot.
+/// with each distinct target having its own slot. Each connection is wrapped in
+/// a ConnectionInfo, and stored in the pool as a pointer.
 ///
 /// Connections can be retrieved from and replaced in the pool, at the front of
 /// the slot. Connections that have gone unused for a while are removed
 /// periodically from the back of the slots.
+///
+/// Retrieved connections are wrapped in ConnectionHandle objects, which, when
+/// destroyed, handle returning the connection to the pool.
 template <typename T>
 class ConnectionPool
 {
@@ -194,20 +200,23 @@ ConnectionHandle<T> ConnectionPool<T>::get_connection(AddrInfo target)
             target.address.to_string().c_str(),
             target.port);
 
+  ConnectionInfo<T>* conn_info_ptr;
+
   pthread_mutex_lock(&_conn_pool_lock);
 
   typename Pool::iterator slot_it = _conn_pool.find(target);
 
-  ConnectionInfo<T>* conn_info_ptr;
-
   if ((slot_it != _conn_pool.end()) && (!slot_it->second.empty()))
   {
+    // If there is a connection in the pool for the given AddrInfo, retrieve it
     conn_info_ptr = slot_it->second.front();
     slot_it->second.pop_front();
     TRC_DEBUG("Found existing connection %p in pool", conn_info_ptr);
   }
   else
   {
+    // If there is no connection in the pool for the given AddrInfo, create a
+    // new one
     TRC_DEBUG("No existing connection in pool, create one");
     conn_info_ptr = new ConnectionInfo<T>(create_connection(target), target);
     TRC_DEBUG("Created new connection %p", conn_info_ptr);
@@ -295,6 +304,7 @@ ConnectionHandle<T>::ConnectionHandle(ConnectionInfo<T>* conn_info_ptr,
 template <typename T>
 ConnectionHandle<T>::~ConnectionHandle()
 {
+  // On destruction, release the connection back into the pool
   _conn_pool_ptr->release_connection(_conn_info_ptr);
 }
 
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -139,8 +139,6 @@ private:
   ConnectionPool<T>* _conn_pool_ptr;
 };
 
-#endif
-
 template<typename T>
 ConnectionPool<T>::ConnectionPool(time_t max_idle_time_s) :
   _max_idle_time_s(max_idle_time_s)
@@ -306,3 +304,5 @@ AddrInfo ConnectionHandle<T>::get_target()
 {
   return _conn_info.target;
 }
+
+#endif
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -81,7 +81,7 @@ class ConnectionPool
 
 public:
   ConnectionPool(time_t max_idle_time_s);
-  ~ConnectionPool() {}
+  virtual ~ConnectionPool() {}
 
   /// Retrieves a connection for the given target from the pool if it exists,
   /// and creates one otherwise. Returns this connection wrapped in a
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -46,14 +46,15 @@
 #include "log.h"
 
 // Required as AddrInfo is defined here
-#include "baseresolver.h"
+#include "utils.h"
 
 template <typename T>
 class ConnectionHandle;
 
-/// Abstract struct storing a connection object and associated information. T is
-/// required to be copy- and move- constructable; it is the intention that T
-/// will be a pointer to a connection object on the heap.
+/// Abstract struct storing a connection object and associated information. The
+/// underlying connection type T is required to be copy- and move- constructable;
+/// this condition is satisified by letting T be a pointer to a connection
+/// object on the heap.
 template <typename T>
 struct ConnectionInfo
 {
@@ -67,7 +68,8 @@ struct ConnectionInfo
 
   ConnectionInfo(T conn, AddrInfo target) :
     conn(conn),
-    target(target)
+    target(target),
+    last_used_time_s(0)
   {
   }
 };
@@ -82,6 +84,10 @@ struct ConnectionInfo
 ///
 /// Retrieved connections are wrapped in ConnectionHandle objects, which, when
 /// destroyed, handle returning the connection to the pool.
+///
+/// This class should be subclassed for a specific connection type, with the
+/// subclass responsible for implementing the creation and destruction of the
+/// type T connection object when required by the pool.
 template <typename T>
 class ConnectionPool
 {
@@ -94,6 +100,12 @@ class ConnectionPool
 
 public:
   ConnectionPool(time_t max_idle_time_s);
+
+  /// The pool cannot be safely emptied in this destructor, as an implementation
+  /// of the destroy_connection method is required to safely destroy type T
+  /// connection objects. Subclass destructors should call the
+  /// destroy_connection_pool method of this class to ensure the pool is
+  /// destroyed correctly.
   virtual ~ConnectionPool() {}
 
   /// Retrieves a connection for the given target from the pool if it exists,
@@ -171,12 +183,14 @@ template<typename T>
 void ConnectionPool<T>::destroy_connection_pool()
 {
   pthread_mutex_lock(&_conn_pool_lock);
-  // Iterate over the slots in the pool
+  // Iterate over the slots in the pool. The typename keyword is required to
+  // clarify the type declaration to the compiler.
   for (typename Pool::iterator slot_it = _conn_pool.begin();
        slot_it != _conn_pool.end();
        ++slot_it)
   {
-    // Iterate over the ConnectionInfo objects in the current slot
+    // Iterate over the ConnectionInfo objects in the current slot. The typename
+    // keyword is required to clarify the type declaration to the compiler.
     for (typename Slot::iterator conn_info_it = slot_it->second.begin();
          conn_info_it != slot_it->second.end();
          ++conn_info_it)
@@ -185,8 +199,7 @@ void ConnectionPool<T>::destroy_connection_pool()
       // ConnectionInfo
       destroy_connection((*conn_info_it)->conn);
       // Destroy the current ConnectionInfo
-      delete *conn_info_it;
-      *conn_info_it = NULL;
+      delete *conn_info_it; *conn_info_it = NULL;
     }
   }
   _conn_pool.clear();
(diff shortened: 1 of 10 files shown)


include/memcachedstore.h
```diff
@@ -67,8 +68,6 @@ class BaseMemcachedStore : public Store
   // we expect there to be servers.
   bool has_servers() { return true; };
 
-  void set_max_connect_latency(unsigned int ms);
```
TJW2: This method is never called, so it is removed - there is no support for setting the max_connect_latency in the new MemcachedConnectionPool

include/memcachedstore.h
```diff
@@ -476,7 +391,7 @@ class TopologyNeutralMemcachedStore : public BaseMemcachedStore
   memcached_return_t iterate_through_targets(
     std::vector<AddrInfo>& targets,
     SAS::TrailId trail,
-    std::function<memcached_return_t(ConnectionHandle&)> fn);
+    std::function<memcached_return_t(ConnectionHandle<memcached_st*>&)> fn);
```
TJW2: This method is only used internally by MemcachedStore, so it's safe to change its parameters

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+
+#include <map>
+#include <deque>
+#include <pthread.h>
+#include <time.h>
+
+#include "log.h"
+
+// Required as AddrInfo is defined here
+#include "baseresolver.h"
```
AlexHockey: It feels like you should refactor this class out into a separate file.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+
+template <typename T>
+class ConnectionHandle;
+
+/// Abstract struct storing a connection object and associated information. T is
```
AlexHockey: Probably worth mentioning that T is the type of the underlying connection.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+/// required to be copy- and move- constructable; it is the intention that T
```
AlexHockey: I don't think it is *intended* for T to be a pointer, but it *can* be.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+
+  // An AddrInfo object storing details of the target of the connection
+  AddrInfo target;
+
+  // The time in seconds that the connection was last used
+  time_t last_used_time_s;
+
+  ConnectionInfo(T conn, AddrInfo target) :
+    conn(conn),
+    target(target)
```
AlexHockey: Initialize the `time_t`?

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+/// Abstract template class storing a pool of connection objects, in "slots",
+/// with each distinct target having its own slot. Each connection is wrapped in
+/// a ConnectionInfo, and stored in the pool as a pointer.
+///
+/// Connections can be retrieved from and replaced in the pool, at the front of
+/// the slot. Connections that have gone unused for a while are removed
+/// periodically from the back of the slots.
+///
+/// Retrieved connections are wrapped in ConnectionHandle objects, which, when
+/// destroyed, handle returning the connection to the pool.
```
AlexHockey: Would be good to mention that this class should be subclassed, and the subclass is responsible for creating/destroying the actual connection when prompted by the `ConnectionPool`

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+template <typename T>
+class ConnectionPool
+{
+  // The ConnectionHandle class must call the protected release_connection
+  // method of this class.
+  friend class ConnectionHandle<T>;
+
+  using Slot = std::deque<ConnectionInfo<T>*>;
+  using Pool = std::map<AddrInfo, Slot>;
```
AlexHockey: Nice!

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+{
+  pthread_mutex_init(&_conn_pool_lock, NULL);
+}
+
+template<typename T>
+void ConnectionPool<T>::destroy_connection_pool()
+{
+  pthread_mutex_lock(&_conn_pool_lock);
+  // Iterate over the slots in the pool
+  for (typename Pool::iterator slot_it = _conn_pool.begin();
```
AlexHockey: The `typename` may confuse people so it would be good to add a comment saying it is just needed to clarify what is going on to the compiler (basically something saying "this code does what you expect it to!)

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+    for (typename Slot::iterator conn_info_it = slot_it->second.begin();
+         conn_info_it != slot_it->second.end();
+         ++conn_info_it)
+    {
+      // Safely destroy the connection object contained in the current
+      // ConnectionInfo
+      destroy_connection((*conn_info_it)->conn);
+      // Destroy the current ConnectionInfo
+      delete *conn_info_it;
+      *conn_info_it = NULL;
```
AlexHockey: Good practice to put these two statements on the same line (so they are less likely to get separated if the code is changed in future).

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+       slot_it != _conn_pool.end();
+       ++slot_it)
+  {
+    if (!slot_it->second.empty())
+    {
+      // Connections are always checked in/out at the front of the slot, so the
+      // oldest one is at the back
+      ConnectionInfo<T>* oldest_conn_info_ptr = slot_it->second.back();
+
+      if(current_time > oldest_conn_info_ptr->last_used_time_s + _max_idle_time_s)
```
AlexHockey: Space before the `(` please

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+      {
+        TRC_DEBUG("Free idle connection to IP: %s, port: %d (time now is %ld, last used %ld)",
+                  oldest_conn_info_ptr->target.address.to_string().c_str(),
+                  oldest_conn_info_ptr->target.port,
```
AlexHockey: I thin `AddrInfo` has a `to_string` method now, so you should use that.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+  {
+    if (!slot_it->second.empty())
+    {
+      // Connections are always checked in/out at the front of the slot, so the
+      // oldest one is at the back
+      ConnectionInfo<T>* oldest_conn_info_ptr = slot_it->second.back();
+
+      if(current_time > oldest_conn_info_ptr->last_used_time_s + _max_idle_time_s)
+      {
+        TRC_DEBUG("Free idle connection to IP: %s, port: %d (time now is %ld, last used %ld)",
```
AlexHockey: The last two format parameters are the wrong type.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+                  oldest_conn_info_ptr->target.address.to_string().c_str(),
+                  oldest_conn_info_ptr->target.port,
+                  ctime(&current_time),
+                  ctime(&(oldest_conn_info_ptr->last_used_time_s)));
```
AlexHockey: This is not safe. According to http://www.cplusplus.com/reference/ctime/ctime/:

> Return Value
> A C-string containing the date and time information in a human-readable format.
>
> The returned value points to an internal array whose validity or value may be altered by any subsequent call to asctime or ctime.

So the second call to ctime may invalidate the result of the first call.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+
+        // Delete the connection as it is too old
+        slot_it->second.pop_back();
+        destroy_connection(oldest_conn_info_ptr->conn);
+        delete oldest_conn_info_ptr;
```
AlexHockey: Would be good to NULL out the `oldest_conn_info_ptr` here.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+{
+  // On destruction, release the connection back into the pool
+  _conn_pool_ptr->release_connection(_conn_info_ptr);
+}
+
+template <typename T>
+ConnectionHandle<T>::ConnectionHandle(ConnectionHandle<T>&& conn_handle) :
+  _conn_info_ptr(conn_handle._conn_info_ptr),
+  _conn_pool_ptr(conn_handle._conn_pool_ptr)
+{
```
AlexHockey: I think you need to NULL out the pointers in the object that has been moved from otherwise the connection would be released twice. Do you have a UT case to cover this?

include/memcachedconnectionpool.h
```diff
@@ -0,0 +1,72 @@
...
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+#ifndef MEMCACHEDCONNECTIONPOOL_H__
+#define MEMCACHEDCONNECTIONPOOL_H__
+
+// Compilation fails when surrounding these two includes with extern "C". The
+// code seems to work without it, but this may cause problems in future.
```
AlexHockey: What is the point of this comment? I think it's to give future developer a hint if they see errors due to undefined references at compile time. It would be good if the comment were a bit more specific.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+  // The ConnectionHandle class must call the protected release_connection
+  // method of this class.
+  friend class ConnectionHandle<T>;
+
+  using Slot = std::deque<ConnectionInfo<T>*>;
+  using Pool = std::map<AddrInfo, Slot>;
+
+public:
+  ConnectionPool(time_t max_idle_time_s);
+  virtual ~ConnectionPool() {}
```
AlexHockey: IT would be useful to explain here why you don't empty the pool in this destructor.

include/connectionpool.h
```diff
@@ -0,0 +1,338 @@
...
+
+protected:
+  /// Creates a type T connection for the given target
+  virtual T create_connection(AddrInfo target) = 0;
+
+  /// Safely destroys a type T connection
+  virtual void destroy_connection(T conn) = 0;
+
+  /// Safely destroys the connection pool, leaving it empty. This method must be
+  /// called from the destructor of all subclasses
```
AlexHockey: Worth elaborating on this (i.e. the subclass needs to not have been destroyed when the pool is emptied, and what the consequences are if this were not called).

include/memcachedstore.h
AlexHockey: Sounds good.

include/memcachedconnectionpool.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "connectionpool.h"
+
+class MemcachedConnectionPool : public ConnectionPool<memcached_st*>
+{
+public:
+  MemcachedConnectionPool(time_t max_idle_time_s, std::string options) :
+    ConnectionPool<memcached_st*>(max_idle_time_s),
+    _options(options),
+    _max_connect_latency(50)
```
AlexHockey: This used to be called `_max_connect_latency_ms`. The units are useful so please can you change the name back.

src/memcachedconnectionpool.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "memcachedconnectionpool.h"
+
+memcached_st* MemcachedConnectionPool::create_connection(AddrInfo target)
+{
+  // Create and set up a memcached connection
+  memcached_st* conn = memcached(_options.c_str(), _options.length());
+  memcached_behavior_set(conn,
+                          MEMCACHED_BEHAVIOR_CONNECT_TIMEOUT,
+                          _max_connect_latency);
```
AlexHockey: Alignment looks off

AlexHockey:
This looks really good. Well done. I've put some comments inline but most were cosmetic. 

A couple of high level things:

* When you make a pull request, could you put some detail in the description (containing what the PR is for, what testing you've done, and any other useful information). We sometimes go back to old PRs to figure out why a change was made, and this information is handy. 
* It's good to try and identify where you're doing something weird (or not doing something expected), and add a comment to satisfy the reader. For example when your connection pool destructor didn't free any objects.

include/connectionpool.h
TJW2: I've moved this class into the utils.h file, alongside the definition of the IP46Address struct, as this seemed like the most logical place for it. As utils.h is imported by BaseResolver, there are no issues with backwards compatibility here.

include/connectionpool.h
TJW2: I've added to the comment to clarify this.

include/connectionpool.h
TJW2: I've modified the comment to make it clear that T can be any copy- or move-constructable type, and that this condition is satisfied nicely by using a pointer to a connection object on the heap.

include/connectionpool.h
TJW2: The last_used_time_s variable is now set to zero on initialization.

include/connectionpool.h
TJW2: I've added a comment to this effect.

include/connectionpool.h
TJW2: I've added a comment to this effect.

include/connectionpool.h
TJW2: I've modified this entire TRC_DEBUG call to resolve the problems that you have mentioned.

include/connectionpool.h
TJW2: I've added a UT to test this, as well as setting the relevant pointers to NULL on moving.

include/connectionpool.h
TJW2: I've added a comment to this effect.

include/connectionpool.h
TJW2: This will be covered in the comment on the destructor of ConnectionPool.

include/memcachedconnectionpool.h
TJW2: I've made this comment more specific.

include/connectionpool.h
```diff
@@ -0,0 +1,362 @@
...
+      {
+        /// Create strings required for debug logging
+        std::string addr_info_str = oldest_conn_info_ptr->target.address_and_port_to_string();
+        std::string current_time_str = ctime(&current_time);
+        std::string last_used_time_s_str = ctime(&(oldest_conn_info_ptr->last_used_time_s));
+
+        TRC_DEBUG("Free idle connection to target: %s (time now is %s, last used %s)",
+                  addr_info_str.c_str(),
+                  current_time_str.c_str(),
+                  last_used_time_s_str.c_str());
```
AlexHockey: Nice. The one extra change I would make is to wrap this in `if (Log::enabled(Log::DEBUG_LEVEL))`. This prevents the program from creating the strings if debug logging is not turned on.

AlexHockey:
Thanks Tom. One small comment.

PR #547, opened 2016-08-18
[Reviewer: EM] Add Connection Pooling to HttpConnection

TJW2:
A refactor of the HttpConnection class in httpconnection.h and httpconnection.cpp to improve the send_request method, and integrate connection pooling.

### DIFF
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -117,6 +117,10 @@ protected:
   /// Creates a type T connection for the given target
   virtual T create_connection(AddrInfo target) = 0;
 
+  /// Safely destroys a type T connection, which targets the given AddrInfo,
+  /// using the destroy_connection method
+  virtual void destroy_connection_with_target(AddrInfo target, T conn);
+
   /// Safely destroys a type T connection
   virtual void destroy_connection(T conn) = 0;
 
@@ -128,6 +132,7 @@ protected:
   /// to the pool or safely destroys it as specified by the second parameter.
   virtual void release_connection(ConnectionInfo<T>* conn_info,
                                   bool return_to_pool);
+
 private:
   /// Removes one connection that has gone unused for more than the max idle
   /// time, if any such connections exist
@@ -165,7 +170,7 @@ public:
   // Gets the AddrInfo object contained within _conn_info
   AddrInfo get_target();
 
-  void set_release_to_pool(bool return_to_pool);
+  void set_return_to_pool(bool return_to_pool);
 
 private:
   // A ConnectionInfo containing the connection
@@ -204,7 +209,7 @@ void ConnectionPool<T>::destroy_connection_pool()
     {
       // Safely destroy the connection object contained in the current
       // ConnectionInfo
-      destroy_connection((*conn_info_it)->conn);
+      destroy_connection_with_target((*conn_info_it)->target, (*conn_info_it)->conn);
       // Destroy the current ConnectionInfo
       delete *conn_info_it; *conn_info_it = NULL;
     }
@@ -248,6 +253,13 @@ ConnectionHandle<T> ConnectionPool<T>::get_connection(AddrInfo target)
 }
 
 template<typename T>
+void ConnectionPool<T>::destroy_connection_with_target(AddrInfo target,
+                                                       T conn)
+{
+  destroy_connection(conn);
+}
+
+template<typename T>
 void ConnectionPool<T>::release_connection(ConnectionInfo<T>* conn_info_ptr,
                                            bool return_to_pool)
 {
@@ -271,7 +283,7 @@ void ConnectionPool<T>::release_connection(ConnectionInfo<T>* conn_info_ptr,
   else
   {
     // Safely destroy the connection and its associated ConnectionInfo
-    destroy_connection(conn_info_ptr->conn);
+    destroy_connection_with_target(conn_info_ptr->target, conn_info_ptr->conn);
     delete conn_info_ptr; conn_info_ptr = NULL;
   }
 
@@ -313,7 +325,7 @@ void ConnectionPool<T>::free_old_connection()
 
         // Delete the connection as it is too old
         slot_it->second.pop_back();
-        destroy_connection(oldest_conn_info_ptr->conn);
+        destroy_connection_with_target(oldest_conn_info_ptr->target, oldest_conn_info_ptr->conn);
         delete oldest_conn_info_ptr; oldest_conn_info_ptr = NULL;
 
         // Delete the entire slot if it is now empty
@@ -382,7 +394,7 @@ AddrInfo ConnectionHandle<T>::get_target()
 }
 
 template <typename T>
-void ConnectionHandle<T>::set_release_to_pool(bool release_to_pool)
+void ConnectionHandle<T>::set_return_to_pool(bool release_to_pool)
 {
   _release_to_pool = release_to_pool;
 }
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -125,7 +125,7 @@ protected:
   void destroy_connection_pool();
 
   /// Releases the given connection back into the pool
-  void release_connection(ConnectionInfo<T>* conn_info);
+  virtual void release_connection(ConnectionInfo<T>* conn_info);
 
 private:
   /// Removes one connection that has gone unused for more than the max idle
--- a/include/connectionpool.h
+++ b/include/connectionpool.h
@@ -3,7 +3,7 @@
  * pooling
  *
  * Project Clearwater - IMS in the Cloud
- * Copyright (C) 2013  Metaswitch Networks Ltd
+ * Copyright (C) 2016  Metaswitch Networks Ltd
  *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the
@@ -180,6 +180,9 @@ public:
   // Gets the AddrInfo object contained within _conn_info
   AddrInfo get_target();
 
+  // Sets the value of the return to pool flag. When true, the connection is
+  // returned to the pool on destruction of the handle, and when false, the
+  // connection is destroyed on destruction of the handle.
   void set_return_to_pool(bool return_to_pool);
 
 private:
@@ -189,9 +192,9 @@ private:
   // A pointer to the ConnectionPool that created this object
   ConnectionPool<T>* _conn_pool_ptr;
 
-  // True if the connection should be released to the pool on destruction of the
+  // True if the connection should be returned to the pool on destruction of the
   // handle, and false if it should be destroyed. Defaults to true.
-  bool _release_to_pool;
+  bool _return_to_pool;
 };
 
 template<typename T>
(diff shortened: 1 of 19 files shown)


include/connectionpool.h
```diff
@@ -164,12 +180,18 @@ class ConnectionHandle
   // Gets the AddrInfo object contained within _conn_info
   AddrInfo get_target();
 
+  void set_return_to_pool(bool return_to_pool);
```
eleanor-merry: Can you add a function comment here?

src/httpconnectionpool.cpp
```diff
@@ -0,0 +1,157 @@
+/**
+ * @file httpconnectionpool.cpp  Implementation of derived class for HTTP
```
eleanor-merry: http_connection_pool, and copyright year is wrong

src/httpconnectionpool.cpp
```diff
@@ -0,0 +1,157 @@
...
+ * software and licensed under the OpenSSL Licenses.
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+#include "httpconnectionpool.h"
+#include "httpconnection.h"
+
+// LCOV_EXCL_START
```
eleanor-merry: Why is this file excluded from coverage?

src/httpconnectionpool.cpp
```diff
@@ -0,0 +1,157 @@
...
+  _timeout_ms = HttpConnection::calc_req_timeout_from_latency((load_monitor != NULL) ?
+                                                              load_monitor->get_target_latency_us() :
+                                                              DEFAULT_LATENCY_US);
+
+  TRC_STATUS("  Connection will use a response timeout of %ldms", _timeout_ms);
+}
+
+CURL* HttpConnectionPool::create_connection(AddrInfo target)
+{
+    CURL* conn = curl_easy_init();
```
eleanor-merry: Indentation

src/httpconnectionpool.cpp
```diff
@@ -0,0 +1,157 @@
...
+    curl_easy_setopt(conn, CURLOPT_VERBOSE, 1L);
+
+    return conn;
+}
+
+void HttpConnectionPool::increment_statistic(AddrInfo target, CURL* conn)
+{
+  if (_stat_table)
+  {
+    // Increment the statistic
```
eleanor-merry: Indentation

src/httpconnectionpool.cpp
```diff
@@ -0,0 +1,157 @@
...
+    _stat_table->get(ip_address)->increment();
+  }
+}
+
+void HttpConnectionPool::decrement_statistic(AddrInfo target, CURL* conn)
+{
+  if (_stat_table)
+  {
+    // Decrement the statistic
+      char buf[100];
```
eleanor-merry: Indentation

test_utils/testableconnectionpool.h
```diff
@@ -49,5 +49,8 @@ class TestableConnectionPool : public ConnectionPool<int>
 
 protected:
   MOCK_METHOD1(create_connection, int(AddrInfo target));
+  MOCK_METHOD2(increment_statistic, void(AddrInfo target, int conn));
+
```
eleanor-merry: Extra blank line

src/httpconnection.cpp
```diff
@@ -99,25 +70,25 @@ HttpConnection::HttpConnection(const std::string& server,
...
-  _stat_table(stat_table)
+  _stat_table(stat_table),
+  _conn_pool(load_monitor, stat_table)
 {
-  pthread_key_create(&_curl_thread_local, cleanup_curl);
   pthread_key_create(&_uuid_thread_local, cleanup_uuid);
   pthread_mutex_init(&_lock, NULL);
   curl_global_init(CURL_GLOBAL_DEFAULT);
+
+  if (_load_monitor)
```
eleanor-merry: What's this if statement and no_stats vector for?

src/httpconnectionpool.cpp
```diff
@@ -0,0 +1,157 @@
...
+HttpConnectionPool::HttpConnectionPool(LoadMonitor* load_monitor,
+                                       SNMP::IPCountTable* stat_table) :
+  ConnectionPool<CURL*>(MAX_IDLE_TIME_MS),
+  _stat_table(stat_table)
+{
+  _timeout_ms = HttpConnection::calc_req_timeout_from_latency((load_monitor != NULL) ?
+                                                              load_monitor->get_target_latency_us() :
+                                                              DEFAULT_LATENCY_US);
+
+  TRC_STATUS("  Connection will use a response timeout of %ldms", _timeout_ms);
```
eleanor-merry: Does this TRC_STATUS still make sense?

src/httpconnection.cpp
```diff
@@ -462,48 +340,58 @@ HTTPCode HttpConnection::send_get(const std::string& path,
...
+                                      std::string& doc,
+                                      const std::string& username,
+                                      SAS::TrailId trail,
+                                      std::vector<std::string> headers_to_add,
+                                      std::map<std::string, std::string>* response_headers)
+{
+  HTTPCode http_code;
+  CURLcode rc;
 
   // Create a UUID to use for SAS correlation and add it to the HTTP message.
```
eleanor-merry: Comment out of date

src/httpconnection.cpp
```diff
@@ -584,30 +421,53 @@ HTTPCode HttpConnection::send_request(const std::string& path,                 /
...
+
+    curl_easy_setopt(curl, CURLOPT_HTTPHEADER, extra_headers);
+
+    // Set general curl options
+    set_curl_options_general(curl, body, doc);
+
+    // Set response header curl options
+    set_curl_options_response(curl, response_headers);
+
+    //Set request-type specific curl options
```
eleanor-merry: Missing space

src/httpconnection.cpp
```diff
@@ -643,39 +505,43 @@ HTTPCode HttpConnection::send_request(const std::string& path,                 /
     {
       TRC_ERROR("%s failed at server %s : %s (%d) : fatal",
                 url.c_str(), remote_ip, curl_easy_strerror(rc), rc);
-      sas_log_curl_error(trail, remote_ip, i->port, method_str, url, rc, 0);
+      sas_log_curl_error(trail, remote_ip, target_it->port, method_str, url, rc, 0);
     }
 
+    http_code = curl_code_to_http_code(curl, rc);
```
eleanor-merry: Why has this moved?

include/connectionpool.h
```diff
@@ -117,21 +117,37 @@ class ConnectionPool
...
+  virtual void release_connection(ConnectionInfo<T>* conn_info,
+                                  bool return_to_pool);
 
 private:
   /// Removes one connection that has gone unused for more than the max idle
   /// time, if any such connections exist
   void free_old_connection();
 
+  /// Creates a type T connection and calls increment statistic
+  T create_connection_and_incr(AddrInfo target);
```
eleanor-merry: I'm not convinced by this layout. I was thinking that create_connection itself would call increment_statistic. Let's discuss this.

include/connectionpool.h
```diff
@@ -117,21 +117,37 @@ class ConnectionPool
   /// Creates a type T connection for the given target
   virtual T create_connection(AddrInfo target) = 0;
 
+  /// Called just after a connection is created - can be overridden to increment
+  /// a relevant statistic
+  virtual void increment_statistic(AddrInfo target, T conn) {}
```
eleanor-merry: Now that I've seen the code, I'm not sure this is the nicest way to split this up. I'm leaning towards making a destroy_connection call that takes a AddrInfo as well as a conn, and a create_connection call that just takes a AddrInfo. I think this works for memcached as well as destroy connection is only called when you're holding a ConnectionInfo (which includes the AddrInfo). Let's discuss

eleanor-merry:
Looks good. I've made some minor style-type comments, and one bigger comment about the statistics method.

include/connectionpool.h
TJW2: I've added a comment explaining the return to pool flag, and also cleared up some inconsistencies in naming.

src/httpconnectionpool.cpp
TJW2: I've fixed the copyright year across the new files that I've created. In terms of the file name, I copied the style of httpconnection and httpresolver - should these be changed also?

src/httpconnectionpool.cpp
TJW2: I've moved this block to the correct indentation.

src/httpconnectionpool.cpp
TJW2: I've moved this block to the correct indentation.

test_utils/testableconnectionpool.h
TJW2: I've removed this line.

src/httpconnection.cpp
TJW2: This is carried over from the previous code, which had two separate constructors, one of which was not passed a load monitor. A grep for "no_stats" in cpp-common returns only this result, and the vector would drop out of scope in the old constructor anyway. Can I safely assume that this is leftover from some previous feature, and remove it?

src/httpconnectionpool.cpp
TJW2: I've changed the message to reflect that this is now an option set on the HttpConnectionPool. On a related note, would it be worth moving the calc_req_timout_from_latency static method from HttpConnection into HttpConnectionPool? HttpConnectionPool would still need to be a friend of HttpConnection, but the method seems to logically belong with the connection pool now, and is only called from there.

src/httpconnection.cpp
TJW2: I've changed the comment to reflect that the UUID is no longer added at this point.

src/httpconnection.cpp
TJW2: In the old code, the curl pointer from the last attempt to connect persisted after the for loop had completed. With the new code, the curl pointer is set inside the loop, and the ConnectionHandle falls out of scope after each iteration of the loop. I think it would be possible to move this back outside the loop, but it wouldn't be that nice. It seemed to make sense to finish using the curl object as soon as possible.

src/httpconnection.cpp
TJW2: I've fixed this.

src/httpconnectionpool.cpp
TJW2: This was left over from before the class was integrated into HttpConnection, and should have been removed. I've now removed it.

src/httpconnection.cpp
eleanor-merry: Yup

src/httpconnection.cpp
eleanor-merry: Cool, makes sense

src/httpconnectionpool.cpp
eleanor-merry: Yup, makes sense

TJW2: As per our discussion, I've moved this method to HttpConnectionPool.

eleanor-merry:
LGTM

PR #548, opened 2016-08-18
Add extra reserved block

ChrisRees:


ChrisRees:
Eleanor this is the corresponding change to reserve an Alarm block.
Should be ok to merge, but figured you should see it anyway.

eleanor-merry:
:shipit:

PR #549, opened 2016-08-19
[Reviewer: AJH] Add HTTP Lazy Targeting

TJW2:
Add resolve methods to BaseResolver and HttpResolver that return an iterator enabling targets to be selected lazily, using the most up-to-date information available to the resolver in the state of the hosts.


src/baseresolver.cpp
```diff
@@ -337,133 +337,185 @@ void BaseResolver::srv_resolve(const std::string& srv_name,
...
+    std::string blacklist_str = "[" + result.address_and_port_to_string() + "]";
+    added_from_unhealthy_str = added_from_unhealthy_str + blacklist_str;
+    TRC_DEBUG("Added an unhealthy server, now have %ld of %d",
+              targets.size(),
+              targets_count);
+  }
+
+  if (_trail != 0)
+  {
+    SAS::Event event(_trail, SASEvent::BASERESOLVE_A_RESULT, 0);
```
TJW2: I'm not sure that this is the correct place for logging this event. Where should SAS logging be performed, given that A/AAAA record resolution and target selection are now separate?

src/baseresolver.cpp
```diff
@@ -337,133 +337,185 @@ void BaseResolver::srv_resolve(const std::string& srv_name,
...
+  AddrInfo ai;
+  ai.port = port;
+  ai.transport = transport;
 
-  // Do A/AAAA lookup.
-  DnsResult result = _dns_client->dns_query(hostname, (af == AF_INET) ? ns_t_a : ns_t_aaaa, trail);
-  ttl = result.ttl();
+  // Create an AddrInfo object for each returned DnsRRecord, and add them to the
+  // query results vector
+  for (std::vector<DnsRRecord*>::const_iterator result_it = dns_result.records().begin();
```
TJW2: I'm a little worried about efficiency here. I think that in cases with no or few graylisted records, a similar number of AddrInfo objects are created as previously, and this approach leads to clearer and simpler code, but I'm not sure we won't lost some performance.

src/httpclient.cpp
```diff
@@ -381,16 +387,16 @@ HTTPCode HttpClient::send_request(RequestType request_type,
   std::string host = host_from_server(server);
   int port = port_from_server(server);
 
+#ifdef UNIT_TEST
+  // Resolve the host using a FakeHttpResolver. Here we rely on _resolver being
+  // a FakeHttpResolver during unit testing. This is not particularly nice, but
+  // does enable the use of a FakeIterator.
+  TRC_DEBUG("Using FakeIterator");
+  FakeIterator target_it = dynamic_cast<FakeHttpResolver*>(_resolver)->resolve_iter(host, port, trail);
```
TJW2: This is fairly unpleasant, but was the best we could come up with in the current framework. Take a look at my general comment for a potential resolution here.

src/httpresolver.cpp
```diff
@@ -84,6 +81,15 @@ void HttpResolver::resolve(const std::string& host,
...
-    a_resolve(host, _address_family, port, TRANSPORT, max_targets, targets, dummy_ttl, trail);
+    Iterator it = resolve_iter(host, port, trail);
+    targets = it.take(max_targets);
   }
 }
+
+BaseResolver::Iterator HttpResolver::resolve_iter(const std::string& host,
+                                                 int port,
+                                                 SAS::TrailId trail)
+{
```
TJW2: Here we don't check for the possibility that `host' is already an IP address. I'm not sure how important this is, but see my general comment for a possible resolution.

TJW2:
I've had problems with testing HttpClient/HttpConnection with the lazy targeting Iterator; the tests now work, but not in the nicest way. Since Iterator contains a reasonable amount of code, it was desirable to use a mock or fake iterator for testing, but object slicing in HttpClient meant that this was difficult. 

A potential solution is to enable Iterator to function without a BaseResolver, in which case it is passed a vector of AddrInfo objects and simply returns them in order. This would enable relatively simple testing, albeit without mocks, and would let HttpResolver return a simple Iterator containing a single target in the event that the resolve_iter method is passed a host that is already an IP address.

However, it really feels like this simpler Iterator should be a separate object, to avoid code duplication/complication in the existing Iterator, and just because they perform different functions. This puts us back to square one with regard to the object slicing problems in testing.

I'd like to discuss this if possible to try and come up with a cleaner solution, although given the timings this may need to fall into next sprint - the code does at least work at present.

include/baseresolver.h
```diff
@@ -85,6 +84,12 @@ class BaseResolver
   static bool parse_ip_target(const std::string& target, IP46Address& address);
 
   void clear_blacklist();
+
+
```
AlexHockey: Double blank line.

include/baseresolver.h
```diff
@@ -50,9 +50,8 @@
 #include "utils.h"
 #include "sas.h"
 
-/// Overrides Google Test default print method to avoid compatibility issues
-/// with valgrind
-void PrintTo(const AddrInfo& ai, std::ostream* os);
+class BaseAddrIterator;
```
AlexHockey: Why do you need to forward declare the `BaseAddrIterator`?

include/baseresolver.h
```diff
@@ -319,4 +333,72 @@ class BaseResolver
...
+class BaseAddrIterator
+{
+public:
+  BaseAddrIterator() {}
+  virtual ~BaseAddrIterator() {}
+
+  /// Should return a vector containing at most targets_count AddrInfo targets.
+  virtual std::vector<AddrInfo> take(int targets_count) = 0;
+
+  /// Calls the take method with targets_count set to one. If the returned
```
AlexHockey: The first sentence of this comment is an implementation detail.

include/sasevent.h
```diff
@@ -138,7 +138,7 @@ namespace SASEvent {
   const int MEMCACHED_TRY_HOST = COMMON_BASE + 0x00010E;
 
   const int BASERESOLVE_SRV_RESULT = COMMON_BASE + 0x000200;
-  const int BASERESOLVE_A_RESULT = COMMON_BASE + 0x000201;
+  const int BASERESOLVE_TARGET_SELECT = COMMON_BASE + 0x000201;
```
AlexHockey: This name should refer to A records

src/baseresolver.cpp
```diff
@@ -347,123 +347,24 @@ void BaseResolver::a_resolve(const std::string& hostname,
...
-  if (trail != 0)
-  {
-    SAS::Event event(trail, SASEvent::BASERESOLVE_A_RESULT, 0);
-    event.add_var_param(hostname);
-    event.add_var_param(targetlist_str);
-    event.add_var_param(unhealthy_str);
-    event.add_var_param(added_from_unhealthy_str);
-    SAS::report_event(event);
-  }
+  return new LazyAddrIterator(result, this, port, transport, trail);
```
AlexHockey: Do we need to `std::move` the `result` or is the compiler smart enough to automatically call the move constructor?

src/baseresolver.cpp
```diff
@@ -1044,3 +945,177 @@ void BaseResolver::untested(const AddrInfo& ai)
...
+  // Determine the number of elements to be returned, and create an iterator
+  // pointing to the location at which to split _targets.
+  int actual_targets_count = std::min(targets_count, int(_targets.size()));
+  std::vector<AddrInfo>::iterator targets_it = _targets.begin();
+  std::advance(targets_it, actual_targets_count);
+
+  // Set targets to the first actual_targets_count elements of _targets, and
+  // remove those elements from _targets.
+  std::vector<AddrInfo> targets(_targets.begin(), targets_it);
+  _targets = std::vector<AddrInfo>(targets_it, _targets.end());
```
AlexHockey: This code isn't very efficient if there are lots of entries in `_targets` as you copy the entire vector each time you call take. It might have been better to implement it more like the lazy iterator (i.e. reverse the addresses and slice off the end of the vector. 

I don't think it's worth making this change (given that this class is only used for testing for with small numbers of entries), but something to bear in mind in the future.

src/baseresolver.cpp
```diff
@@ -1044,3 +945,177 @@ void BaseResolver::untested(const AddrInfo& ai)
...
+  // Strings for logging purposes.
+  std::string graylisted_targets_str;
+  std::string whitelisted_targets_str;
+  std::string blacklisted_targets_str;
+
+  std::string found_blacklisted_str;
+
+  // This lock must be held to safely call the host_state method of
+  // BaseResolver.
+  pthread_mutex_lock(&(_resolver->_hosts_lock));
```
AlexHockey: Reaching into the base resolver to take its lock is a bit nasty. Is there anything else we could do here?

src/baseresolver.cpp
```diff
@@ -1044,3 +945,177 @@ void BaseResolver::untested(const AddrInfo& ai)
...
+  // If there are any graylisted records, the Iterator should return one first,
+  // and then no more.
+  if (_first_call)
+  {
+    _first_call = false;
+
+    // Iterate over the records.
+    for (std::vector<AddrInfo>::iterator result_it = _unused_results.begin();
+         result_it != _unused_results.end();
+         ++result_it)
```
AlexHockey: It doesn't matter too much but reverse iterating is probably more efficient (As it means you remove from near the end of the vector rather than near the front.

src/baseresolver.cpp
```diff
@@ -1044,3 +945,177 @@ void BaseResolver::untested(const AddrInfo& ai)
...
+  {
+    ai.address = _resolver->to_ip46(*result_it);
+    _unused_results.push_back(ai);
+  }
+
+  // Shuffle the results for load balancing purposes
+  std::random_shuffle(_unused_results.begin(), _unused_results.end());
+}
+
+std::vector<AddrInfo> LazyAddrIterator::take(int targets_count)
```
AlexHockey: `num_requested_targets` might be better. `targets_count` sounded a bit like an internal helper variable.

src/httpclient.cpp
```diff
@@ -404,12 +396,18 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
-  std::vector<AddrInfo>::const_iterator target_it;
+  AddrInfo target;
 
-  for (target_it = targets.begin(); target_it != targets.end(); ++target_it)
+  // Iterate over the targets returned by target_it until a successful
+  // connection is made, a specified number of failures is reached, or the
+  // targets are exhausted. If only one target is available, it should be tried
+  // twice.
+  for (int attempts = 0;
+       target_it->next(target) || attempts == 1;
```
AlexHockey: This relies on `next` not modifying it's argument if it returns false. I think this is fine, but you should document that clearly in the header file.

test_utils/mockhttpresolver.h
```diff
@@ -40,18 +40,16 @@
...
 {
 public:
-  MockHttpResolver() : HttpResolver(nullptr,0,0,0) {}
+  MockHttpResolver() : HttpResolver(nullptr, 0, 0, 0) {}
   ~MockHttpResolver() {}
-  MOCK_METHOD5(resolve, void(const std::string& host,
-                             int port,
-                             int max_targets,
-                             std::vector<AddrInfo>& targets,
-                             SAS::TrailId trail));
```
AlexHockey: As you still have a `resolve` method you might as well keep the mock method around.

AlexHockey:
I've reviewed. This generally looks pretty good. A few comments inline. 

I must admit I don't understand why it was so much easier to convert the DNS results in AddrInfos when the iterator is constructed. You only look at the unused addresses a couple of times so it feels like it would have been quite easy to leave them as DNSResults until you needed to return them. However, I don't think the performance of what you've done will be too bad, so let's leave it as it is for now.

src/baseresolver.cpp
TJW2: The DnsResult is passed by reference, so I don't think any copying or moving takes place.

src/httpclient.cpp
TJW2: I've made a comment to this effect on the next method of BaseAddrResolver.

include/baseresolver.h
TJW2: The return type of a_resolve_iter is BaseAddrIterator*, so this needs to be forward declared - however, LazyAddrIterator does not need to be forward declared, so I have removed this.

src/baseresolver.cpp
TJW2: I had considered this. An additional complication is that the take method would have reverse the vector of targets to be returned so as to keep them in the order in which they were passed in to the iterator.

In the case that most calls to take are for a single element, there would be an increase in efficiency here, but, as you say, at the moment this class is only used with small numbers of entries.

src/baseresolver.cpp
TJW2: We could create a method on BaseResolver that takes the lock, calls into host state, releases the lock, and returns the obtained state. This wouldn't be hard to implement.

include/baseresolver.h
AlexHockey: You could just put the definition of BaseAddrIterator here rather than the forward declaration. Though I suppose it's also nice to keep the interface an implementations together. I don't care too much.

src/baseresolver.cpp
AlexHockey: Ah, OK, I'd missed that. Thanks!

src/baseresolver.cpp
AlexHockey: Yeah, that doesn't sound great either. I've had a bit of a think and the only other solutions I came up with were pretty complex. So let's just live with this. At least the lock is defined in the same file.

PR #550, opened 2016-08-19
[Reviewer: Ellie] Clarify latency comment in HttpConnection

mirw:
As you pointed out, the calculated HTTP connection latency is actually only half of the total time (as it doesn't include the retry).  The comment at the top of httpconnection.cpp is misleading - this fixes it.


eleanor-merry:
Sure

PR #551, opened 2016-08-23
[Reviewer: RKD] Modify debug string creation to improve performance

TJW2:
Modify BaseResolver so that strings required for debug logging are created only when debug logging is switched on; unnecessarily creating these strings was causing a performance hit.


rkd-msw:
Looks good, thanks!

PR #553, opened 2016-08-25
[Reviewer: Richard] Make IP address checks wider

eleanor-merry:
is_ipv6 just reported if an address was an IPv6 address or not. By itself, this wasn't particularly useful, so I've changed it so that it reports whether the address is valid, and whether it has a port.

Tested in UTs.


include/utils.h
```diff
@@ -339,6 +340,10 @@ namespace Utils
         tokens.push_back(token);
         num_tokens++;
       }
+      else if (count_empty_tokens)
+      {
+        tokens.push_back("");
```
richardwhiuk: I think should be increment num_tokens here, otherwise splitting with max tokens will probably give you a different result than you'd expect (i.e. split ("::", sep => ":", max_tokens => 2) will give you "", "", "", instead of "", ":".

In which case, I think you should do `if (token.length() > 0 || count_empty_tokens)`

include/utils.h
```diff
@@ -299,14 +299,15 @@ namespace Utils
...
+  /// at the end of tokens.
   template <class T>  //< container that has T::push_back(std::string)
   void split_string(const std::string& str_in,  //< string to scan (will not be changed)
                     char delimiter,  //< delimiter to use
                     T& tokens,  //< tokens will be added to this list
                     const int max_tokens = 0,  //< max number of tokens to push; last token will be tail of string (delimiters will not be parsed in this section)
                     bool trim = false,  //< trim the string at both ends before splitting?
-                    bool check_for_quotes = false) //< only use delimiters not in quotes
+                    bool check_for_quotes = false,
+                    bool count_empty_tokens = false) //< whether empty tokens are counted
```
richardwhiuk: I think include_empty_tokens is clearer than count_empty_tokens, which sounds like you are going to return a count of them, instead of including them in the result.

include/utils.h
```diff
@@ -355,6 +360,10 @@ namespace Utils
     {
       tokens.push_back(token);
     }
+    else if (count_empty_tokens)
+    {
+      tokens.push_back("");
```
richardwhiuk: Surely this would be better as `if (token.length() > 0 || count_empty_tokens)` rather than the same effective logic twice.

include/utils.h
```diff
@@ -556,6 +565,18 @@ namespace Utils
...
                         int log_level,
                         bool log_to_file);
+
+  enum IPAddressType {
+    IPV4_ADDRESS,
+    IPV4_ADDRESS_WITH_PORT,
+    IPV6_ADDRESS,
+    IPV6_ADDRESS_WITH_PORT,
+    INVALID
+  };
```
richardwhiuk: I think we care about the difference between `::1`, `[::1]` and `[::1]:80` - i.e. IPv6 address, bracketed IPv6 address and bracketed IPv6 address with port.

src/utils.cpp
```diff
@@ -590,3 +590,31 @@ void Utils::daemon_log_setup(int argc,
 
   TRC_STATUS("Log level set to %d", log_level);
 }
+
+Utils::IPAddressType Utils::parse_ip_address(std::string address)
+{
+  // Check if we have a port
+  std::string host;
+  int port;
+  bool with_port = Utils::split_host_port(address, host, port);
```
richardwhiuk: I'm slightly worried that about what split_host_port does with something like `::1:1` - I think it'll strip the empty components and return host of `1` and port of `1`, instead of considering it as an IPv6 address equivalent to `0000:0000:0000:0000:0000:0000:0001:0001`

include/utils.h
```diff
@@ -556,6 +565,18 @@ namespace Utils
...
+
+  // Takes a string and reports what type of IP address it is
+  IPAddressType parse_ip_address(std::string address);
```
richardwhiuk: Previously this function was in the header file, so it was suitable for in-lining. I don't think this is a big deal as I doubt we do this in performance critical code, but worth bearing in mind.

src/utils.cpp
eleanor-merry: It won't do the split, as this isn't an address that has a port on it (so it's ::1:1 that gets passed to inet_pton)

richardwhiuk:
Mostly this looks fine - some stylistic comments, some concerns over purely bracketed IPv6 addresses and an edge case.

As discussed, your about to send me the UTs.

include/utils.h
eleanor-merry: Agree - I'll add IPV6_ADDRESS_BRACKETED as well

include/utils.h
eleanor-merry: Fixed

include/utils.h
eleanor-merry: Sure

src/utils.cpp
richardwhiuk: Looking at https://github.com/Metaswitch/cpp-common/blob/03777a60307233f7921a5e6d6219bf7db4cfa1cf/src/utils.cpp#L317

close_bracket will be npos.

split_string will discard the empty parts, so `host_port_parts[0]` will be `"1"` and `host_ports_parts[1]` will be `"1"`

`port` will then be `1` and `host` will be `"1"`

inet_pton will get called with a host of "1", which will fail with both AF_INET and AF_INET6

And then we'll return INVALID I think.

eleanor-merry: It won't discard the empty parts (because of the new check in this PR to count empty tokens), so host_port_parts size won't be 2, so we'll bail out of split_host_port

PR #554, opened 2016-08-26
[Reviewer: Richard] Check for IPv6 address in parse_ip_target

eleanor-merry:
Use the new parse_ip_address to correctly determine when we have an IP address

Fixes https://github.com/Metaswitch/sprout/issues/1317

(Not tested at all yet)


src/baseresolver.cpp
```diff
@@ -527,13 +527,18 @@ bool BaseResolver::parse_ip_target(const std::string& target, IP46Address& addre
   std::string ip_target = target;
   Utils::trim(ip_target);
 
-  if (inet_pton(AF_INET6, ip_target.c_str(), &address.addr.ipv6) == 1)
+  Utils::IPAddressType address_type = Utils::parse_ip_address(target);
```
richardwhiuk: This function used to fill in address.addr.ipv6 here, and below address.addr.ipv4 which is owned by the target, which the caller probably cares about being correct. This no longer does that.

richardwhiuk:
I don't think this works - we aren't parsing the address into address. Unless this is never called with that being cared about (in which case we can probably remove it as a argument), I don't think this works.

src/baseresolver.cpp
eleanor-merry: True - thinking about this, I'll just move to unbracketing the IP address, and leaving this as it was

PR #555, opened 2016-08-26
[Reviewer: EM] Add function to bracket IPv6 addresses

richardwhiuk:
Ellie,

As discussed, we'd like a common function in cpp-common to get an IPv6 address in brackets. This does that.

There's a separate pull request coming for the associated tests in cpp-common-test.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -577,6 +577,9 @@ namespace Utils
   // doesn't include a port.
   std::string uri_ip_address(std::string address, int default_port = 0);
 
+  // Removes the brackets from an IPv6 address - e.g. [::1] -> ::1
+  std::string remove_brackets_from_ip(std::string address);
+
 } // namespace Utils
 
 #endif /* UTILS_H_ */
--- a/include/utils.h
+++ b/include/utils.h
@@ -571,6 +571,12 @@ namespace Utils
   // Takes a string and reports what type of IP address it is
   IPAddressType parse_ip_address(std::string address);
 
+  // Takes an IP address and returns it suitable for use in a URI,
+  // i.e, an IPv6 address will be returned in brackets.
+  // If the optional port is passed, this will be appended if the address
+  // doesn't include a port.
+  std::string uri_ip_address(std::string address, int default_port = 0);
+
 } // namespace Utils
 
 #endif /* UTILS_H_ */
--- a/include/utils.h
+++ b/include/utils.h
@@ -565,7 +565,8 @@ namespace Utils
     IPV6_ADDRESS,
     IPV6_ADDRESS_BRACKETED,
     IPV6_ADDRESS_WITH_PORT,
-    INVALID
+    INVALID,
+    INVALID_WITH_PORT
   };
 
   // Takes a string and reports what type of IP address it is
@@ -575,11 +576,13 @@ namespace Utils
   // i.e, an IPv6 address will be returned in brackets.
   // If the optional port is passed, this will be appended if the address
   // doesn't include a port.
-  std::string uri_ip_address(std::string address, int default_port = 0);
+  std::string uri_address(std::string address, int default_port = 0);
 
   // Removes the brackets from an IPv6 address - e.g. [::1] -> ::1
   std::string remove_brackets_from_ip(std::string address);
 
+  // Does the passed in address have brackets?
+  bool is_bracketed_address(const std::string& address);
 } // namespace Utils
 
 #endif /* UTILS_H_ */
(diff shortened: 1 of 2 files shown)


richardwhiuk:
I'll add unbracket first.

src/utils.cpp
```diff
@@ -591,6 +591,47 @@ void Utils::daemon_log_setup(int argc,
...
+
+std::string Utils::uri_ip_address(std::string address, int default_port)
+{
+  Utils::IPAddressType addrtype = parse_ip_address(address);
+
+  if (default_port == 0)
+  {
+    if (addrtype == IPAddressType::IPV6_ADDRESS)
+    {
+      address = "[" + address + "]";
```
eleanor-merry: Is this really what we want to do? Why bracket if there's no port?

src/utils.cpp
```diff
@@ -591,6 +591,47 @@ void Utils::daemon_log_setup(int argc,
   TRC_STATUS("Log level set to %d", log_level);
 }
 
+std::string Utils::remove_brackets_from_ip(std::string address)
+{
+  bool with_brackets = ((address.size() >= 2) &&
```
eleanor-merry: Not a fan of the repeated code from parse_ip_address, but I don't really care

src/utils.cpp
```diff
@@ -591,6 +591,47 @@ void Utils::daemon_log_setup(int argc,
...
+                        (address[0] == '[') &&
+                        (address[address.size() - 1] == ']'));
+
+  return with_brackets ? address.substr(1, address.size() - 2) : address;
+}
+
+std::string Utils::uri_ip_address(std::string address, int default_port)
+{
+  Utils::IPAddressType addrtype = parse_ip_address(address);
```
eleanor-merry: I'm wondering if we should bail out here if we don't have an IP address (e.g. just return an empty string)

eleanor-merry:
Looks good, some comments inline

src/utils.cpp
richardwhiuk: https://www.ietf.org/rfc/rfc2732.txt suggests that if you include a IPv6 address in a URL - which is where I think you need it in bracketed form, you should do so in square brackets.

eleanor-merry: Oo, good to know

src/utils.cpp
richardwhiuk: Hmm, maybe I should add a 'is_bracketed_address' function. Then I can use that in both places, and it might be genuinely useful.

src/utils.cpp
richardwhiuk: I think I actually need to create a parse_address, which can also return AddressType::HOSTNAME and AddressType::HOSTNAME_WITH_PORT, and then have this turn any address into a URI form with port...

richardwhiuk:
Can you have a quick glance over the changes I've made?

src/utils.cpp
```diff
@@ -591,6 +591,48 @@ void Utils::daemon_log_setup(int argc,
...
+    if (addrtype == IPAddressType::IPV6_ADDRESS)
+    {
+      address = "[" + address + "]";
+    }
+  }
+  else
+  {
+    std::string port = std::to_string(default_port);
+
+    if (addrtype == IPAddressType::IPV4_ADDRESS ||
```
eleanor-merry: Check for !(types with port) to make it clearer what's happening?

eleanor-merry: Ah, I see, brackets make that harder

src/utils.cpp
```diff
@@ -591,6 +591,48 @@ void Utils::daemon_log_setup(int argc,
...
+bool Utils::is_bracketed_address(const std::string& address)
+{
+  return ((address.size() >= 2) &&
+          (address[0] == '[') &&
+          (address[address.size() - 1] == ']'));
+}
+
+std::string Utils::remove_brackets_from_ip(std::string address)
+{
+  return is_bracketed_address(address) ? address.substr(1, address.size() - 2) : address;
```
eleanor-merry: Line length (and below as well)

include/utils.h
```diff
@@ -565,12 +565,24 @@ namespace Utils
     IPV6_ADDRESS,
     IPV6_ADDRESS_BRACKETED,
     IPV6_ADDRESS_WITH_PORT,
-    INVALID
+    INVALID,
```
eleanor-merry: I don't like this, but I'm OK with it

eleanor-merry:
Sure

PR #557, opened 2016-09-06
[Reviewer: GDR] Experimental Result Codes must be scoped by vendor

richardwhiuk:
Graeme,

As discussed, the current experimental result API on a Diameter Message only returns the experimental result code, as opposed to also including the vendor ID.

As experimental result codes are scoped by vendor ID, we should also include the vendor ID in the response - a better name for 'Experimental Result' would be 'Vendor Specific Result' I think.

Note, that the vendor ID may differ between the VENDOR_SPECIFIC_APPLICATION_ID and the EXPERIMENTAL_RESULT AVPs - although I suspect that's in practice not very likely - so the existing call to vendor_id() is not appropriate here.

As such, I've changed cpp-common to have a experimental_result function which will provide the code and the vendor ID of the first Experimental Result found, returning true if found, and false otherwise. It will return false in the event of a malformed EXPERIMENTAL_RESULT - which is against spec.

This is part of a co-ordinate checkin with homestead which uses the new function.

### DIFF
--- a/include/diameterstack.h
+++ b/include/diameterstack.h
@@ -419,7 +419,7 @@ public:
     result = _result;
     return (result != 0);
   }
-  int32_t experimental_result_code() const;
+  bool experimental_result(int32_t& experimental_result_code, uint32_t& vendor_id) const;
   int32_t vendor_id() const;
   inline std::string impi() const
   {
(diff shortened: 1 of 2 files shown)


graemerobertson:
Looks good

PR #558, opened 2016-09-07
[Reviewer: RJW2] BGCF-20 Break out WeightedSelector into a separate class

benritchie:
Rich
Can you review this please? - its the change you suggested as part of the Load-balancing you reviewed.  
Sprout UT's have been run.  is there any other component likely to be affected by this?
Thanks

### DIFF
--- a/include/weightedselector.h
+++ b/include/weightedselector.h
@@ -1,0 +1,166 @@
+  /**
+ * @file weightedselector.h  Declaration of base class for DNS resolution.
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2013  Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version, along with the "Special Exception" for use of
+ * the program along with SSL, set forth below. This program is distributed
+ * in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details. You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ * The author can be reached by email at clearwater@metaswitch.com or by
+ * post at Metaswitch Networks Ltd, 100 Church St, Enfield EN2 6BQ, UK
+ *
+ * Special Exception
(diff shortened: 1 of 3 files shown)


include/weightedselector.h
```diff
@@ -0,0 +1,166 @@
+  /**
```
richardwhiuk: This line shouldn't be indented.

src/baseresolver.cpp
```diff
@@ -177,7 +178,7 @@ void BaseResolver::srv_resolve(const std::string& srv_name,
       srvs.reserve(i->second.size());
 
       // Build a cumulative weighted tree for this priority level.
-      SRVWeightedSelector selector(i->second);
+      WeightedSelector<BaseResolver::SRV> selector(i->second);
```
richardwhiuk: I think this can just be `WeightedSelector<SRV>`.

include/weightedselector.h
```diff
@@ -0,0 +1,166 @@
...
+
+#ifndef WEIGHTEDSELECTOR_H__
+#define WEIGHTEDSELECTOR_H__
+
+#include <vector>
+
+#include "utils.h"
+#include "baseresolver.h"
+
+  /// The WeightedSelector class is a class used to implement resource
```
richardwhiuk: These lines shouldn't be indented.

include/weightedselector.h
```diff
@@ -0,0 +1,166 @@
...
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+#ifndef WEIGHTEDSELECTOR_H__
+#define WEIGHTEDSELECTOR_H__
+
+#include <vector>
+
+#include "utils.h"
+#include "baseresolver.h"
```
richardwhiuk: I don't think you should need either of these imports.

richardwhiuk:
This mostly looks fine - some stylistic comments.

The UTs for the BaseResolver are in https://github.com/Metaswitch/cpp-common-test - specifically this is effectively tested in `src/ut/baseresolver_test.cpp` - you should run those - `make full_test` against the updated `cpp-common code` from the base directory will run the tests under valgrind.

It might be nice to add some generic unit tests just to test the template class, but you haven't reduced our coverage here, so I'm happy if you don't feel you have time to do that.

I assume this is sufficient for you integrate this externally?

include/weightedselector.h
richardwhiuk: Oh and 'The WeightSelector is a template class used to implement resource...' would be nicer so we aren't repeating class twice in a single sentence - currently it reads oddly.

include/weightedselector.h
benritchie: I think utils.h is needed to get access to std::string
Agree the other isn't.  - removed

richardwhiuk: `#include <string>` then?

PR #559, opened 2016-09-08
[Reviewer: RJW2]  Minor Additions for WeightedSelector

jack-atack:
As you know BJR2 made some changes yesterday.  A couple of minor amendments have been made to WeightedSelector and a get_weight function has had to be added to baseresolver.  Could you please check these are ok?

### DIFF
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -170,7 +170,10 @@ protected:
     int port;
     int priority;
     int weight;
-    int get_weight() const {return weight;};
+    int get_weight() const
+    {
+      return weight;
+    }
   };
   typedef std::map<int, std::vector<SRV> > SRVPriorityList;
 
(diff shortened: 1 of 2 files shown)


include/baseresolver.h
```diff
@@ -170,6 +170,7 @@ class BaseResolver
     int port;
     int priority;
     int weight;
+    int get_weight() const {return weight;};
```
richardwhiuk: This should be formatted as:

```cpp
int get_weight() const
{
  return weight;
}
```

richardwhiuk:
Looks good. I'm not really sure why you renamed get_rand_node(), but I guess the name makes marginally more sense with routes instead of SRV records. Presumably this has broken the tests in cpp-common-test, given it looks like get_rand_node was being mocked out?

jack-atack:
get_rand_node() was a typo. The function wasn't in the original baseresolver code, it was inserted purely to let us mock out a line of code.

PR #561, opened 2016-09-20
[Reviewer: Alex] Enhance Cassandra connections

sebrexmetaswitch:
Allow homestead to connect to Cassandra on more than one host, using DNS A records

 - move to connection pooling, rather than one persistent connection per thread
 - use DNS resolver to decide which host to connect to for each operation

Unit tests aren't done yet

### DIFF
--- a/src/cassandra_connection_pool.cpp
+++ b/src/cassandra_connection_pool.cpp
@@ -38,6 +38,7 @@
 #include "cassandra_connection_pool.h"
 #include "cassandra_store.h"
 
+// LCOV_EXCL_START - UTs do not cover the creation/deletion on Clients
 CassandraConnectionPool::CassandraConnectionPool() :
   ConnectionPool<CassandraStore::Client*>(MAX_IDLE_TIME_S)
 {
@@ -69,4 +70,5 @@ void CassandraConnectionPool::destroy_connection(AddrInfo target, CassandraStore
 {
   delete conn; conn = NULL;
 }
+// LCOV_EXCL_STOP
 
--- a/src/cassandra_connection_pool.cpp
+++ b/src/cassandra_connection_pool.cpp
@@ -1,0 +1,69 @@
+/**
+ * @file cassandra_connection_pool.cpp  Implementation of derived class for
+ * Cassandra connection pooling.
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2016 Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version, along with the "Special Exception" for use of
+ * the program along with SSL, set forth below. This program is distributed
+ * in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details. You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ * The author can be reached by email at clearwater@metaswitch.com or by
+ * post at Metaswitch Networks Ltd, 100 Church St, Enfield EN2 6BQ, UK
+ *
+ * Special Exception
+ * Metaswitch Networks Ltd  grants you permission to copy, modify,
+ * propagate, and distribute a work formed by combining OpenSSL with The
+ * Software, or a work derivative of such a combination, even if such
+ * copying, modification, propagation, or distribution would otherwise
+ * violate the terms of the GPL. You must comply with the GPL in all
+ * respects for all of the code used other than OpenSSL.
+ * "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
+ * Project and licensed under the OpenSSL Licenses, or a work based on such
+ * software and licensed under the OpenSSL Licenses.
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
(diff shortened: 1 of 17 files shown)


src/cassandra_connection_pool.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "cassandra_store.h"
+
+CassandraConnectionPool::CassandraConnectionPool() :
+  ConnectionPool<CassandraStore::Client*>(MAX_IDLE_TIME_S)
+{
+}
+
+CassandraStore::Client* CassandraConnectionPool::create_connection(AddrInfo target)
+{
+  //We require the address as a string
```
AlexHockey: Missing space between `//` and `We`

src/cassandra_store.cpp
```diff
@@ -138,61 +149,88 @@ Store::Store(const std::string& keyspace) :
...
   _max_queue(0),
   _thread_pool(NULL),
-  _comm_monitor(NULL),
-  _thread_local()
+  _comm_monitor(NULL)
 {
-  // Create the thread-local storage that stores cassandra connections.
-  // delete_client is a destroy callback that is called when the thread exits.
-  pthread_key_create(&_thread_local, delete_client);
+  _conn_pool = new CassandraConnectionPool();
```
AlexHockey: This would probably be better initialized in the member initialization list:
```cpp
_conn_pool(new CassandraConnectionPool())
```

src/cassandra_store.cpp
```diff
@@ -138,61 +149,88 @@ Store::Store(const std::string& keyspace) :
...
+  BaseAddrIterator* target_it = _resolver->resolve_iter(_cass_hostname,
+                                                        _cass_port,
+                                                        0);
+  AddrInfo target;
+  bool retry = true;
+  int attempt_count = 0;
+
+  // Iterate over targets and try to connect. We only try the next target if
+  // there is a transport error, and we limit the loop to 2 targets.
+  while (retry && attempt_count < 2 && (target_it->next(target)))
```
AlexHockey: We normally put compound conditional expressions in brackets, and don't use brackets otherwise. i.e. 

```cpp
while (retry && (attempt_count < 2) && target_it->next(target))
```

src/cassandra_store.cpp
```diff
@@ -138,61 +149,88 @@ Store::Store(const std::string& keyspace) :
...
   {
-    TRC_ERROR("Store caught unknown exception!");
-    rc = UNKNOWN_ERROR;
+    retry = false;
+    attempt_count++;
+    rc = OK;
+
+    ConnectionHandle<Client*> conn_handle = get_client(target);
+    client = conn_handle.get_connection();
```
AlexHockey: You should probably define the `Client*` here - that way you can't use the pointer when you've dropped the `ConnectionHandle`

src/cassandra_store.cpp
```diff
@@ -138,61 +149,88 @@ Store::Store(const std::string& keyspace) :
...
+    try
+    {
+      if (!client->is_connected())
+      {
+        TRC_DEBUG("Connecting to %s", target.to_string().c_str());
+        client->connect();
+        client->set_keyspace(_keyspace);
+      }
```
AlexHockey: Why is `get_client` not doing this?

src/cassandra_store.cpp
```diff
@@ -138,61 +149,88 @@ Store::Store(const std::string& keyspace) :
...
+      TRC_ERROR("Store caught InvalidRequestException: %s", ire.why.c_str());
+      rc = INVALID_REQUEST;
+    }
+    catch(...)
+    {
+      TRC_ERROR("Store caught unknown exception!");
+      rc = UNKNOWN_ERROR;
+    }
+
+    conn_handle.set_return_to_pool(false);
```
AlexHockey: So we never return the client to the pool??

src/cassandra_store.cpp
```diff
@@ -272,56 +310,17 @@ Store::~Store()
...
-    client = NULL;
-    pthread_setspecific(_thread_local, NULL);
-  }
-}
-
-
-void Store::delete_client(void* client)
-{
-  delete (Client*)client; client = NULL;
+  return conn_handle;
```
AlexHockey: You could just return the value without storing it in a local variable.

src/cassandra_store.cpp
```diff
@@ -330,26 +329,47 @@ bool Store::do_sync(Operation* op, SAS::TrailId trail)
...
   Client* client = NULL;
-  ResultCode cass_result = OK;
+  ResultCode cass_result = UNKNOWN_ERROR;
   std::string cass_error_text = "";
 
   // Set up whether the perform should be retried on failure.
-  // Only try once, unless there's connection error, in which case try twice.
-  bool retry = false;
+  // Only retry on connection errors
+  bool retry = true;
```
AlexHockey: This comment and the line of code are at odds. You say we only retry in a specific case, but then set retry to true unconditionally. I think the previous comment was more accurate.

src/cassandra_store.cpp
```diff
@@ -330,26 +329,47 @@ bool Store::do_sync(Operation* op, SAS::TrailId trail)
...
+                                                        _cass_port,
+                                                        trail);
+  AddrInfo target;
+
+  // Iterate over targets until either we succeed in connecting, run out of
+  // targets or hit the maximum (2).
+  // If there is only one target, try it twice.
+  while (retry &&
+         attempt_count <= 2 &&
+         (target_it->next(target) || attempt_count == 1))
```
AlexHockey: Not enough bracketing around the compound conditional expressions.

src/cassandra_store.cpp
```diff
@@ -330,26 +329,47 @@ bool Store::do_sync(Operation* op, SAS::TrailId trail)
...
+      client = conn_handle.get_connection();
 
-      // Get a client to execute the operation.
-      client = get_client();
+      if (!client->is_connected())
+      {
+        TRC_DEBUG("Connecting to %s", target.to_string().c_str());
+        client->connect();
+        client->set_keyspace(_keyspace);
+      }
```
AlexHockey: Again, this feels like it should be in get_client, which should be moved back within the try/catch.

PR #562, opened 2016-09-20
Add new KPIs from TS 32.454

MatMeredith:


src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -59,6 +59,11 @@ class SuccessFailCountByRequestTypeRow: public TimeAndOtherTypeBasedRow<SuccessF
     uint_fast32_t attempts = counts->attempts.load();
     uint_fast32_t successes = counts->successes.load();
     uint_fast32_t failures = counts->failures.load();
+    uint_fast32_t success_percent = 0;
+    if (successes + failures > 0)
```
graemerobertson: I'd just make this check `if (successes > 0)`. (And if I didn't do that, I'd bracket `successes + failures`)

Same below.

src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -59,6 +59,11 @@ class SuccessFailCountByRequestTypeRow: public TimeAndOtherTypeBasedRow<SuccessF
...
+    {
+      success_percent = successes * 100000 / (successes + failures);
```
graemerobertson: I'd bracket `successes * 100000` - and in fact, that might literally change the value of success_percent? I'm not sure which order the operators are currently applied, but we definitely want to do the multiplication first.

Also, shouldn't this be 1000000 rather than 100000? Either way, I think it should be broken down and commented why we're doing this. So actually `(successes * 100 * 10000)`.

Does this break coverage in cpp-common-test, and regardless of whether it does or not can/should this be UTed?

Same below.

MatMeredith:
All good points -- thanks!   Fixed.

MatMeredith:
As far as I can see we don't currently have any UT for the SNMP tables at all.    Obviously we would have some ideally.   Can I propose merging as is, and if I have spare time at the end of the sprint coming back to this?

graemerobertson:
You're right, we don't - https://github.com/Metaswitch/cpp-common-test/issues/3.

However, as that issue mentions, we do have https://github.com/Metaswitch/clearwater-fv-test/blob/master/src/test_snmp.cpp, which you can v easily enhance to test this code I think. Anyway, you can merge this.

PR #564, opened 2016-09-21
[Reviewer: Ellie] Add function to convert binary to hex

rkd-msw:
This moves it out of Homestead so Sprout can use it.

---
This change is [Reviewable](https://reviewable.io/reviews/metaswitch/cpp-common/564)

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -269,6 +269,22 @@ void Utils::create_random_token(size_t length,       //< Number of characters.
   }
 }
 
+std::string Utils::hex(const uint8_t* data, size_t len)
+{
+  static const char* const hex_lookup = "0123456789abcdef";
+  std::string result;
+  result.reserve(2 * len);
+  for (size_t ii = 0; ii < len; ++ii)
+  {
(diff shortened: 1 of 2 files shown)


src/utils.cpp
```diff
@@ -269,6 +269,22 @@ void Utils::create_random_token(size_t length,       //< Number of characters.
...
+  result.reserve(2 * len);
+  for (size_t ii = 0; ii < len; ++ii)
+  {
+    const uint8_t b = data[ii];
+    result.push_back(hex_lookup[b >> 4]);
+    result.push_back(hex_lookup[b & 0x0f]);
+  }
+  return result;
+}
+
```
eleanor-merry: Excessive lines

eleanor-merry:
LGTM. Are there are UTs that need to move to cpp-common-test?

rkd-msw:
No, this was a private method in Homestead so it wasn't UTed directly, only indirectly through UTs on the other methods.

rkd-msw:
(And it's still covered in UT indirectly, in both Sprout and Homestead now)

PR #565, opened 2016-10-03
BGCF-12 Updating SAS Resource bundle version number for failover logs.

ChrisRees:
Updated the Resource bundle number, to that which includes the new failover logs.


ChrisRees:
This should correspond to the other PR's with you.

benritchie:
looks good :)

PR #566, opened 2016-10-07
[Reviewer: Ellie] Changes success percent stats to report 100% when there are no errors

AME2:
Only weird thing here is that we report percentages in units of 10,000 of a percent to maintain accuracy whilst still using integers.

### DIFF
--- a/src/snmp_success_fail_count_by_request_type_table.cpp
+++ b/src/snmp_success_fail_count_by_request_type_table.cpp
@@ -60,7 +60,14 @@ public:
     uint_fast32_t successes = counts->successes.load();
     uint_fast32_t failures = counts->failures.load();
     uint_fast32_t success_percent = 0;
-    if (successes > 0)
+    if (attempts == uint_fast32_t(0))
+    {
+      // If there are no attempts made we report the Success Percent as being
+      // 100% to indicate that there have been no errors.
+      // Note that units for Success Percent are actually 10,000's of a percent.
+      success_percent = 100 * 10000;
+    }
+    else if (successes > 0)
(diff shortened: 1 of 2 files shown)


src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -60,7 +60,14 @@ class SuccessFailCountByRequestTypeRow: public TimeAndOtherTypeBasedRow<SuccessF
     uint_fast32_t successes = counts->successes.load();
     uint_fast32_t failures = counts->failures.load();
     uint_fast32_t success_percent = 0;
-    if (successes > 0)
+    if (attempts == uint_fast32_t(0))
```
eleanor-merry: I'd have done 
```
uint_fast32_t success_percent = 100 * 10000;
if (failures > 0)
...
```

(but I don't care too much)

src/snmp_success_fail_count_by_request_type_table.cpp
```diff
@@ -60,7 +60,14 @@ class SuccessFailCountByRequestTypeRow: public TimeAndOtherTypeBasedRow<SuccessF
...
+    {
+      // If there are no attempts made we report the Success Percent as being
+      // 100% to indicate that there have been no errors.
+      // Note that units for Success Percent are actually 10,000's of a percent.
+      success_percent = 100 * 10000;
```
eleanor-merry: Can you change the name of success_percent to make it clear it works in 10,000ths of a percent? (It's not a name you've added, but it makes your change harder to parse)

PR #570, opened 2016-10-12
[Reviewer: Alex] Make HTTP socket world-writeable

rkd-msw:
I'll test live in my morning.

### DIFF
--- a/src/httpstack.cpp
+++ b/src/httpstack.cpp
@@ -36,6 +36,7 @@
 
 #include "httpstack.h"
 #include <cstring>
+#include <sys/stat.h>
 #include "log.h"
 
 bool HttpStack::_ev_using_pthreads = false;
@@ -249,6 +250,10 @@ void HttpStack::bind_unix_socket(const std::string& bind_path)
     throw Exception("evhtp_bind_socket (unix)", rc);
     // LCOV_EXCL_STOP
   }
+
+  // Socket needs to be world-writeable for nginx to use it
+  mode_t all_access = S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP | S_IWGRP | S_IXGRP | S_IROTH | S_IWOTH | S_IXOTH;
+  chmod(full_bind_address.c_str(), all_access);
 }
 
 void HttpStack::start(evhtp_thread_init_cb init_cb)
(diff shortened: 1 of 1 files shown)


AlexHockey:
I tried to reproduce the problem on a test system without this fix. I stopped sprout, deleted /tmp/sprout-http-mgmt-socket, then started sprout. The socket was writable by all:

```
$ ls -l /tmp/sprout-http-mgmt-socket
srwxrwxrwx 1 sprout sprout 0 Oct 12 08:53 /tmp/sprout-http-mgmt-socket
```

Are you sure this is needed?

rkd-msw:
The difference between our versions is this commit, added after CCv9: https://github.com/Metaswitch/cpp-common/commit/708de28bb3257de0d9267a24e998d647fdfd4a0f . It adds `umask(0)` into the daemonisation process. This means that:

- code on the dev branch that daemonises will work
- code on the dev branch that doesn't daemonise (e.g. docker) won't work - it'll use the default umask of 022
- code on the CCv9 branch won't work - it'll use the default umask of 022

So I think this fix is right, and also we should fix up the `daemonise` function to set a umask of 022 - or set umask(0) everywhere whether we daemonise or not.

src/httpstack.cpp
```diff
@@ -249,6 +250,10 @@ void HttpStack::bind_unix_socket(const std::string& bind_path)
     throw Exception("evhtp_bind_socket (unix)", rc);
     // LCOV_EXCL_STOP
   }
+
+  // Socket needs to be world-writeable for nginx to use it
+  mode_t all_access = S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP | S_IWGRP | S_IXGRP | S_IROTH | S_IWOTH | S_IXOTH;
+  chmod(full_bind_address.c_str(), all_access);
```
AlexHockey: Don't you need to chmod `bind_path` rather than `full_bind_address`?

AlexHockey:
In that case I'm happy, barring the one markup inline. Please can you raise an issue covering the fact we set our umask inconsistently. Having daemons be different fells like it will be a constant source of pain.

src/httpstack.cpp
rkd-msw: Ah, that's why it didn't work when I tested it this morning. Thanks!

PR #572, opened 2016-10-12
[Reviewer: Alex] Don't set the umask as a side-effect of daemonising

rkd-msw:
Fixes the issue from https://github.com/Metaswitch/cpp-common/pull/570#issuecomment-253198617. I'll probably merge after the end of the UK day then keep an eye on the live tests to make sure this doesn't break anything. (We only introduced this `umask(0)` change in March so it shouldn't be too disruptive to revert it.)

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -591,6 +591,9 @@ void Utils::daemon_log_setup(int argc,
       exit(0);
     }
   }
+  
+  mode_t old_umask = umask(0022);
+  TRC_STATUS("umask set to 022 (was %o)", old_umask);
 
   Log::setLoggingLevel(log_level);
 
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -539,9 +539,6 @@ int Utils::daemonize(std::string out, std::string err)
     return errno;
   }
 
-  // Clear any restricted umask
-  umask(0);
-
   // Second fork
   pid = fork();
   if (pid == -1)


AlexHockey:
My understanding is that standard daemons reset their umask. Why so we think this is the right fix (as opposed to, say, setting the umask to 0 regardless of whether we daemonize)?

rkd-msw:
I don't see why standard daemons would need to reset their umask - http://man7.org/linux/man-pages/man2/umask.2.html says that "A child process created via fork(2) inherits its parent's umask.  The umask is left unchanged by execve(2)".

If we do end up doing what you suggest, I think we should set the umask to 022, not 0 - I don't think anything a daemon creates should be world-writeable by default.

AlexHockey:
Forking is not the same as daemonizing. However I've found http://stackoverflow.com/questions/3095566/linux-daemonize which does suggest that resetting the umask is an optional step. 

However it feels to me that we should set the umask to 022 everywhere rather than picking up the default and leaving ourselves susceptible to breakages caused by running in different environments. I suppose the first step though would be to remove the umask call from daemonize though. Thoughts?

rkd-msw:
I've realised that everything calls daemon_log_setup, whether it daemonises or not, to do some common setup stuff, so I don't need to edit main.cpp in every process - I've just added `umask(022)` in there.

AlexHockey:
Should log_daemon_setup *really* be setting the umask? :-(

rkd-msw:
Closing, have raised https://github.com/Metaswitch/cpp-common/issues/584 to cover fixing differently.

PR #573, opened 2016-10-14
[Reviewer: Ellie] Allow DnsCachedResolver to load static DNS records

sebrexmetaswitch:
- constructors now take an optional filename
 - it will try to read this as a JSON file, storing the static records in memory
 - when doing a dns_query, these records will be examined first.

Currently, only CNAME records are supported
 - if one is found, the DNS query is performed on the redirected host rather than the original

Testing done:
 - UTs (in cpp-common-test)
 - some live testing (though more will follow)

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -256,12 +256,18 @@ DnsCachedResolver::~DnsCachedResolver()
 // }
 void DnsCachedResolver::reload_static_records()
 {
+  if (_dns_config_file == "")
+  {
+    // No config file specified, just return
+    return;
+  }
+
   std::ifstream fs(_dns_config_file.c_str());
 
   if (!fs)
   {
-    // File doesn't exist - expected if static records not in use
-    TRC_DEBUG("No static DNS records specified");
+    TRC_ERROR("DNS config file %s missing", _dns_config_file.c_str());
+    CL_DNS_FILE_MISSING.log();
     return;
   }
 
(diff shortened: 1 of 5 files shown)


src/dnscachedresolver.cpp
```diff
@@ -207,6 +226,114 @@ DnsCachedResolver::~DnsCachedResolver()
...
+            }
+            else
+            {
+              TRC_DEBUG("Two CNAME entries found for hostname %s", hostname.c_str());
+              CL_DNS_FILE_DUPLICATES.log();
+            }
+          }
+          else
+          {
+            TRC_DEBUG("Found unsupported record type: %s", type.c_str());
```
eleanor-merry: TRC_ERROR

src/dnscachedresolver.cpp
```diff
@@ -207,6 +226,114 @@ DnsCachedResolver::~DnsCachedResolver()
...
+          }
+        }
+      }
+
+      _static_records.insert(std::make_pair(hostname, records));
+    }
+  }
+  catch (JsonFormatError err)
+  {
+    TRC_ERROR("Error parsing dns config file - records must be an array.");
```
eleanor-merry: Can you pass in the file name rather than say dns config file

src/dnscachedresolver.cpp
```diff
@@ -207,6 +226,114 @@ DnsCachedResolver::~DnsCachedResolver()
...
+    TRC_ERROR("Unable to parse dns config file: %s\nError: %s",
+              dns_config.c_str(),
+              rapidjson::GetParseError_En(doc.GetParseError()));
+    CL_DNS_FILE_MALFORMED.log();
+    return;
+  }
+
+  try
+  {
+    for (rapidjson::Value::ConstMemberIterator hosts_it = doc.MemberBegin();
```
sebrexmetaswitch: I think this will make it invalid JSON, since the hostname is actually the name of the JSON object, and the array is the value.

src/dnscachedresolver.cpp
```diff
@@ -207,6 +226,114 @@ DnsCachedResolver::~DnsCachedResolver()
...
+}
+
+void DnsCachedResolver::read_records_from_file()
+{
+  std::ifstream fs(_dns_config_file.c_str());
+
+  if (!fs)
+  {
+    // File doesn't exist or can't be opened, just return
+    return;
```
sebrexmetaswitch: This is actually the expected case for almost all uses of the DNSCachedResolver, since the _dns_config_file is not specified for most instances. So I've added a DEBUG log

sebrexmetaswitch:
Thanks, I've implemented most of those suggestions with the exception of the two comments replied to above.

Can you take a quick look and check you're happy with the new PD log strings? Just want to check they line up with what you were envisaging.

sebrexmetaswitch:
As discussed, I've now:
 - changed the structure of the JSON file
 - changed DnsCachedResolver to allow it to reload the static DNS records without requiring a restart.

Could you review for me please?

src/dnscachedresolver.cpp
```diff
@@ -207,6 +226,166 @@ DnsCachedResolver::~DnsCachedResolver()
...
+            }
+          }
+          else
+          {
+            TRC_ERROR("Found unsupported record type: %s", type.c_str());
+          }
+        }
+        catch (JsonFormatError err)
+        {
+          TRC_ERROR("Bad DNS record specified for hostname %s in DNS config file %s",
```
eleanor-merry: PD log please

src/dnscachedresolver.cpp
```diff
@@ -122,6 +128,9 @@ void DnsCachedResolver::init(const std::vector<IP46Address>& dns_servers)
   _dns_servers = dns_servers;
   _cache_lock = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
 
+  // Read any static dns records from the config file
+  reload_static_records();
```
eleanor-merry: You could remove this, then have the updater run on start up. That way only those callers that want to use static records will ever call this function.

PR #574, opened 2016-10-20
Reviewer RKD: Syslog changes

graemerobertson:
This PR removes the `syslog_facade.h` file which is no longer necessary (since we no longer use `LOG_*` methods), and moves the call the `openlog` from the PDLog class to our generic initialise logging method.

This requires changes to every project that creates ENT logs - woop.


src/utils.cpp
```diff
@@ -573,6 +574,20 @@ void Utils::daemon_log_setup(int argc,
     prog_name = slash_ptr + 1;
   }
 
+  // Copy the program name to a string so that we can be sure of its lifespan -
+  // the memory passed to openlog must be valid for the duration of the program.
+  //
+  // Note that we don't save syslog_identity here, and so we're technically leaking
+  // this object. However, its effectively part of static initialisation of
+  // the process - it'll be freed on process exit - so its not leaked in practice.
```
rkd-msw: its -> it's

src/utils.cpp
```diff
@@ -573,6 +574,20 @@ void Utils::daemon_log_setup(int argc,
...
+  std::string *syslog_identity = new std::string(prog_name);
```
rkd-msw: `std::string*`? (i.e. space after the *)

rkd-msw:
LGTM.

rkd-msw:
I have a thought on how you can easily make changes to every project - come and chat to me if that looks tedious.

PR #575, opened 2016-10-20
[Reviewer: RJW2] Add scope indexed SNMP tables

sathiyan-sivathas:
This adds two new tables that are indexed by time period and scope (node type):
`EventAccumulatorByScopeTable`
`CounterByScopeTable`

These depend on the new time and scope based row:
`TimeAndScopeBasedRow`

See the sprout pull request for the testing that I've done.

### DIFF
--- a/include/load_monitor.h
+++ b/include/load_monitor.h
@@ -40,7 +40,7 @@
 #include <time.h>
 #include <pthread.h>
 #include "snmp_continuous_accumulator_table.h"
-#include "snmp_scalar.h"
+#include "snmp_abstract_scalar.h"
 #include "sas.h"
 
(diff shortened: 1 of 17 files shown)


richardwhiuk:
I'm sure all of this code works fine and implements the function we need.

However, it all looks like it's copy and pasted, and duplicated. I could make some detailed comments about specific sections, but I think the better plan is to step back and work out how we can reduce the duplicate across the FooTableByBar and FooTableByBarAndBaz classes. I think we should be able to achieve this using templates with relatively little work, while making the classes much more maintainable.

sathiyan-sivathas:
I've made some more changes, could you re-review, please.

PR #578, opened 2016-10-28
[Reviewer: Graeme] Move sas avp to common

eleanor-merry:
This pulls some useful diameter functions into common code


include/sasevent.h
```diff
@@ -151,6 +151,8 @@ namespace SASEvent {
   const int LOAD_MONITOR_ACCEPTED_REQUEST = COMMON_BASE + 0x0500;
   const int LOAD_MONITOR_REJECTED_REQUEST = COMMON_BASE + 0x0501;
 
+  const int DIAMETER_MSG_MISSING_AVP = COMMON_BASE + 0x0600;
```
graemerobertson: Shouldn't this go with the other Diameter events?

PR #579, opened 2016-10-31
[Reviewer: Ellie] Cassandra Connection pool changes

sebrexmetaswitch:
If the cassandra process goes down and comes back up, connections to that
process are no longer valid and should be torn down and recreated.

So this commit adds a member variable to the connection pool. If true, when a
connection is released but marked not to be returned to the pool, then all
existing connections in the pool for the same target are destroyed too.

This is false by default, and only true for the CassandraConnectionPool. This
means that when one bad Cassandra connection is found, the others are
proactively destroyed.

**Testing done**
- UTs (see the cpp-common-test branch)
- live tests

### DIFF
--- a/include/connection_pool.h
+++ b/include/connection_pool.h
@@ -282,7 +282,8 @@ void ConnectionPool<T>::release_connection(ConnectionInfo<T>* conn_info_ptr,
   {
     if (_free_on_error)
     {
-      // Need to destroy all connections for the same target
+      // Need to destroy all connections for the same target that are currently
+      // in the pool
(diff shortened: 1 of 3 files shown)


include/connection_pool.h
```diff
@@ -275,6 +280,27 @@ void ConnectionPool<T>::release_connection(ConnectionInfo<T>* conn_info_ptr,
   }
   else
   {
+    if (_free_on_error)
```
eleanor-merry: Can you comment on the difference between these two destroy actions (one destroys the connections in the pool, one destroys the single connection that isn't in the pool)

PR #581, opened 2016-11-15
Https support

dimbleby:
Add support for querying https servers.

There are two commits here:

- the first is John's original, in which the handling of http queries is untouched, and is what we're currently actually running with
- but the asymmetry between http and https here bothered me, so the second re-unifies them
  -  not sure what the original code was getting at when it claimed that the `CURL_RESOLVE` implementation is 'incomplete' - maybe I've missed something?

I think that the general principle here should be uncontroversial; I'd be happy if you'd take either the one-commit or two-commit version of the fix.

If you take the two-commit version there are some (very minor) UT changes needed to accompany it in sprout and memento (requests now go to hostname rather than to IP address, which a handful of scripts notice).  I have those ready to go but won't submit pull requests until we agree that this is the right plan.

### DIFF
--- a/include/httpconnection.h
+++ b/include/httpconnection.h
@@ -80,13 +80,15 @@ public:
                  SNMP::IPCountTable* stat_table,
                  LoadMonitor* load_monitor,
                  SASEvent::HttpLogLevel,
-                 BaseCommunicationMonitor* comm_monitor);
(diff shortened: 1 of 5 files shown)


dimbleby:
Darn, the merge looks a little awkward... I'll take a look.

dimbleby:
Gonna close this one in favour of one starting from master.

PR #582, opened 2016-11-15
Support querying https URLs

dimbleby:
A rework of #581, but starting at `master`.

As in that one, there are a handful of trivial UT fixes to be made in sprout and memento, that would need to be coordinated with this.

### DIFF
--- a/src/httpclient.cpp
+++ b/src/httpclient.cpp
@@ -369,17 +369,18 @@ HTTPCode HttpClient::send_request(RequestType request_type,
   corr_marker.add_var_param(uuid_str);
   SAS::report_marker(corr_marker, SAS::Marker::Scope::Trace, false);
 
+  std::string scheme;
   std::string server;
   std::string path;
-  if (!Utils::parse_http_url(url, server, path))
+  if (!Utils::parse_http_url(url, scheme, server, path))
   {
     TRC_ERROR("%s could not be parsed as a URL : fatal",
               url.c_str());
     return HTTP_BAD_REQUEST;
   }
 
-  std::string host = host_from_server(server);
-  int port = port_from_server(server);
+  std::string host = host_from_server(scheme, server);
+  int port = port_from_server(scheme, server);
 
   // Resolve the host.
   BaseAddrIterator* target_it = _resolver->resolve_iter(host, port, trail);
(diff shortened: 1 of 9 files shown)


dimbleby:
per email - function added here should be uncontroversial, though obvs the code wants usual standard of review

per #581 John's original fix treated the http and https paths quite separately - presumably with the aim of doing what had to be done for https while leaving http alone - more details in that pull.  Can go back to that if you want but I've been a bit bolder here and unified.

src/httpclient.cpp
```diff
@@ -427,27 +428,22 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
-    {
-      ip_url = "http://[" + std::string(remote_ip) + "]:" + std::to_string(target.port) + path;
-    }
-    else
-    {
-      ip_url = "http://" + std::string(remote_ip) + ":" + std::to_string(target.port) + path;
-    }
+    std::string resolve_addr = host + ":" + std::to_string(port) + ":" + remote_ip;
+    struct curl_slist *host_resolve_add_addr = curl_slist_append(NULL, resolve_addr.c_str());
+    TRC_DEBUG("Set CURLOPT_RESOLVE: %s", resolve_addr.c_str());
```
mirw: Why do we use CURLOPT_RESOLVE when the comment above (in the old version) says the implementation is incomplete.  Was that comment out-of-date?

src/httpclient.cpp
```diff
@@ -427,27 +428,22 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
-    std::string ip_url;
-    if (target.address.af == AF_INET6)
-    {
-      ip_url = "http://[" + std::string(remote_ip) + "]:" + std::to_string(target.port) + path;
-    }
-    else
-    {
-      ip_url = "http://" + std::string(remote_ip) + ":" + std::to_string(target.port) + path;
-    }
+    std::string resolve_addr = host + ":" + std::to_string(port) + ":" + remote_ip;
```
mirw: Does this work correctly for an IPv6 remote_ip?  (Have you tested it?)

src/httpclient.cpp
```diff
@@ -899,23 +905,25 @@ void HttpClient::host_port_from_server(const std::string& server, std::string& h
   else
   {
     host = server_copy;
-    port = 0;
+    port = (scheme == "https") ? DEFAULT_HTTPS_PORT : DEFAULT_HTTP_PORT;
```
mirw: Why have we changed from returning 0 to returning `DEFAULT_*_PORT`?  Would curl not have defaulted based on the scheme?

mirw:
Generally looks good - a few comments inline.

src/httpclient.cpp
dimbleby: Documentation says yes - [here](https://curl.haxx.se/libcurl/c/CURLOPT_RESOLVE.html):

> If libcurl is built to support IPv6, ADDRESS can of course be either IPv4 or IPv6 style addressing. 

But I haven't tested it.

src/httpclient.cpp
dimbleby: John used CURLOPT_RESOLVE for https, presumably because he actually needed to (for else it would surely have been easier not to bother).  I didn't actually check what went wrong if we treated https in the same way as the old code.

I don't know what the comment was getting at when it claimed that the CURLOPT_RESOLVE implementation was incomplete.  So far as I can see it does the job just fine.

dimbleby:
replies inline

src/httpclient.cpp
mirw: OK.

src/httpclient.cpp
mirw: OK, but I think we need to test this - in particular with HTTP, where presumably it's only been tested with HTTPS so far?

mirw:
Further replies inline.

src/httpclient.cpp
dimbleby: (oops, missed this one before reassigning)

The code on our test system is like that in this pull request, but perhaps it never has any reason to do any http stuff.

Again, we don't use the real curl in unit test.  

Do you have an IPv6 system available where we can test live and kill two birds with one stone?

src/httpclient.cpp
mirw: I think we'd need to test live.

src/httpclient.cpp
mirw: We worked out why we needed to use CURLOPT_RESOLVE: because checking the HTTPS certificate requires curl to know the target domain name.

We don't have an IPv6 system set up, although you can do this quite easily with HEAT: https://github.com/Metaswitch/clearwater-heat/tree/ipv6.

dimbleby:
It has occurred to me that I can simply configure my app server to make a request of an http URL and check that this works; and that I can hack the code to use an IPv6 address and check that this also works.  eg just use something like httpbin.org 

(Of course I won't get the right answers if I point at the wrong place, but I can verify that the HTTP requests behave as expected).

I think this will be less trouble for me - shout if you're going to find that unconvincing for some reason.

dimbleby:
Half right - I can test http (and have done so successfully).  But I'm living on EC2 and IPv6 from there seems to be more complicated.

dimbleby:
@rkd-msw tells me that he'll be doing some IPv6 testing in two or three weeks and so that will be an opportune moment to take a look at this too.   So that's the current plan.

mirw:
OK, do you want to merge, then?

dimbleby:
I'll follow up shortly with pull requests for those teensy unit test fixes in other projects

PR #583, opened 2016-11-17
Add default handler to HttpStack and blacklist hosts returning 503 with retry-after header

JamesSilk:
Use built in libevhtp function to allow a default handler to be set on a HttpStack, rather than adding a handler matching on a wildcard regex.

Also modify HttpClient so that hosts which return a 503 response with a Retry-After header are blacklisted. We also need a change to fackecurl to test this; allowing custom headers to be returned.

I've tested these changes on live homestead and sprout nodes. Both functions work as expected. UTs are in a separate pull request.


src/httpclient.cpp
```diff
@@ -501,15 +509,17 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
+                                                       response_headers->end());
+
+      if (((!(http_rc >= 400)) || ((http_rc == 503) && retry_after_present)) &&
           (rc != CURLE_REMOTE_FILE_NOT_FOUND) &&
           (rc != CURLE_REMOTE_ACCESS_DENIED))
       {
         // The CURL connection should not be returned to the pool
         conn_handle.set_return_to_pool(false);
-
         _resolver->blacklist(target);
```
mirw: It looks like we're blacklisting for the default blacklist interval independent of the value of Retry-After.  Is that right?

mirw:
Looks good.  1 comment.

src/httpclient.cpp
JamesSilk: Yep, I should have mentioned that in my original comment. The Retry-After header can take a value of either an integer (number of seconds to wait, which would be fine) or a HTTP Time which we'd have to parse and then deal with any errors parsing and make sure the amount of time to wait from 'now' is positive. Given the nature of the failures that will cause this kind of blacklisting I don't think there's any advantage in the server giving the client an amount of time to wait. But the code could probably do with a comment to summarise that.

mirw: As just discussed, I think it's not hard to parse the relative time and use it, and I think it means we're less likely to hit nasties in future.  (I imagine our server will always return the blacklist value anyway, but we shouldn't be baking that into our client code.)

That said, I don't think we should worry about parsing absolute time values - it's a lot of effort and I don't think I've ever seen anyone use this function over HTTP.

src/httpclient.cpp
```diff
@@ -511,17 +511,40 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
           (rc != CURLE_REMOTE_FILE_NOT_FOUND) &&
           (rc != CURLE_REMOTE_ACCESS_DENIED))
       {
         // The CURL connection should not be returned to the pool
         conn_handle.set_return_to_pool(false);
         _resolver->blacklist(target);
       }
+      else if (((http_rc == 503) && retry_after_present) &&
+               (rc != CURLE_REMOTE_FILE_NOT_FOUND) &&
+               (rc != CURLE_REMOTE_ACCESS_DENIED))
```
mirw: Can `rc` be `CURLE_REMOTE_FILE_NOT_FOUND` or `CURLE_REMOTE_ACCESS_DENIED` if `http_rc` is 503?  It seems not.  (So can we remove these tests from this if expression?)

src/httpclient.cpp
```diff
@@ -511,17 +511,40 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
+      {
+        // Get the value from the Retry-After header. If not an integer just
+        // use the default blacklist call. Note this means that HTTP dates are
+        // ignored.
+        std::string retry_after_val = retry_after_header->second;
+        int retry_after = atoi(retry_after_val.c_str());
```
mirw: I think we should trace (maybe SAS log?) if we fail to parse a Retry-After header.  I think there's an interesting question as to whether we should still blacklist in this case.

Maybe the best logic would be:
```
if !(http_rc >= 400) && ...
  // (As above)
  blacklist for default time
  don't return to pool
else if http_rc == 503:
  check for retry_after_header, and try to parse (moving code from above inside the 503-specific block)
  if it exists and parsing succeeds:
    blacklist for specified time
    don't return to pool
  else:
    report target as successful
else
  report target as successful
```

JamesSilk: Yep, agree with that logic and definitely trace to add. I'll Have a think about SAS logs.

JamesSilk:
OK so I've added in the SAS logging following the existing pattern and the logic has been adjusted to match the previous comment. How does this look?

src/httpclient.cpp
```diff
@@ -927,6 +928,19 @@ void HttpClient::sas_log_curl_error(SAS::TrailId trail,
   }
 }
 
+void HttpClient::sas_log_bad_retry_after_value(SAS::TrailId trail,
+                                               const std::string value,
+                                               uint32_t instance_id)
+{
+  if (_sas_log_level != SASEvent::HttpLogLevel::NONE)
+  {
+    int event_id = (SASEvent::HTTP_BAD_RETRY_AFTER_VALUE);
```
mirw: Do we need to do the same `_sas_log_level` check here as we do in `sas_log_curl_error`?  It feels like we might.

JamesSilk:
New SAS log at detail level added and SAS resource bundle updated too.

mirw:
Looks good.

PR #585, opened 2016-11-21
Don't free curl_slist until curl is done with it

dimbleby:


mirw:
OK, please let me know when you'd like me to review.

mirw:
Please can you let me know what testing you've done?  Thanks!

(I've not reviewed yet, but will do after my next meeting.)

dimbleby:
Fired it up live on the test robocall server, which makes https queries to our reputation provider, and verified that our live test suite for that service works as expected (implying that it made the requests and got the expected answers).

src/httpclient.cpp
```diff
@@ -443,11 +443,19 @@ HTTPCode HttpClient::send_request(RequestType request_type,
                           buf,
                           sizeof(buf));
 
+    // Each time round this loop we add an entry to curl's DNS cache, and also
+    // leave a note reminding ourselves to remove the entry next time.
```
mirw: Please can you provide a bit more background on this in the comment (as it's all quite complicated)?  In particular, I think the key points are that

*   setting CURLOPT_RESOLVE adds/removes entries to/from the DNS cache
    *   to add you add a list containing the new entry you want
    *   to remove you add a list containing the entry you want to remove prefixed by a "-"
*   CURLOPT_RESOLVE is only processed when you actually do a curl request

...so whenever we add an entry to CURLOPT_RESOLVE, we need to prepopulate a list with the corresponding removal and store that in the private data.

PR #586, opened 2016-11-22
Don't resolve something that's already an IP address

dimbleby:
Hopefully one last wrinkle in this set of curl changes.

I don't expect this to make any difference to external behaviour, but it's cleaner not to attempt to resolve something that's already an IP address.

### DIFF
--- a/src/httpclient.cpp
+++ b/src/httpclient.cpp
@@ -384,6 +384,8 @@ HTTPCode HttpClient::send_request(RequestType request_type,
 
   // Resolve the host.
   BaseAddrIterator* target_it = _resolver->resolve_iter(host, port, trail);
+  IP46Address address;
+  bool host_is_ip = BaseResolver::parse_ip_target(host, address);
 
   // Track the number of HTTP 503 and 504 responses and the number of timeouts
   // or I/O errors.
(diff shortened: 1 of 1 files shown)


src/httpclient.cpp
```diff
@@ -384,6 +384,8 @@ HTTPCode HttpClient::send_request(RequestType request_type,
 
   // Resolve the host.
   BaseAddrIterator* target_it = _resolver->resolve_iter(host, port, trail);
+  IP46Address address;
+  bool host_is_ip = BaseResolver::parse_ip_target(host, address);
```
mirw: Can we improve the comment?  It says "Resolve the host", and then parses an IP address, but doesn't use it.  (Maybe it makes sense to rename `address` to `dummy_address` or similar (assuming it's not supposed to be used)?

src/httpclient.cpp
```diff
@@ -455,18 +457,24 @@ HTTPCode HttpClient::send_request(RequestType request_type,
     // remove that entry, and then _next_ time round we use that as well as
     // adding the entry that we do want.
     //
-    // This is the point at which we retrieve that curl_slist (which is always
-    // present, except on the very first time through this loop for this
-    // connection).
+    // This is the point at which we retrieve that curl_slist, if present.
```
mirw: Why might it not be present?

PR #587, opened 2016-11-29
[Reviewer: Seb] Change timeouts

eleanor-merry:
This increases the Cassandra timeouts. 

Tested live.

Fix for issue https://github.com/Metaswitch/homestead/issues/387


sebrexmetaswitch:
Looks good.

As discussed, I think we're happy that 150ms is low enough (we need to ensure there is sufficient time for a request to timeout and still let us try the next node), but that we probably can't go any higher, so this looks right.

PR #588, opened 2016-11-30
[Reviewer: SR2] Add extra reserved blocks for alarms and logs

alango:
Adding extra reserved blocks for alarms and logs.


sebrexmetaswitch:
Yep, looks fine

PR #589, opened 2016-12-01
Add SAS event bases for new components

JamesSilk:


JamesSilk:
Seb, as our contact for this week I've assigned this pull request to you. Should be uncontroversial and I know you're really busy this week, there is no rush for this to get merged.

include/sasevent.h
```diff
@@ -88,6 +88,9 @@ namespace SASEvent {
   const int MANGELWURZEL_BASE = 0x870000;
   const int CEDAR_BASE = 0x880000;
   const int HOUDINI_BASE = 0x890000;
+  const int TIMBERLAKE_BASE = 0x900000;
```
eleanor-merry: Not 8A, 8B, 8C..?

include/sasevent.h
```diff
@@ -42,7 +42,7 @@
 
 namespace SASEvent {
 
-  const std::string CURRENT_RESOURCE_BUNDLE_DATESTAMP = "20161118";
+  const std::string CURRENT_RESOURCE_BUNDLE_DATESTAMP = "20161128";
```
eleanor-merry: You'll need to update this to the current date when you check this in (and update the resource bundle as well)

include/sasevent.h
JamesSilk: Good point. I clearly wasn't thinking in hex!

JamesSilk:
Thanks for the quick turn around!

PR #590, opened 2016-12-05
Remove timberlake from SAS events header file

JamesSilk:
Remove SAS base for timberlake from sasevents.h.


sebrexmetaswitch:
Looks good (assuming that there aren't any other changes required to the other 2 because no events have been created for them yet, so changing the range here for them is fine)

JamesSilk:
Yeah, I've talked to the guys working on Bifrost and Weatherwax and they're fine with this change.

PR #591, opened 2016-12-06
Soften some errors to warnings

dimbleby:
Specifically, failure to contact external HTTP or DNS servers

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -843,7 +843,7 @@ void DnsCachedResolver::dns_response(const std::string& domain,
     // extended for 30 seconds at a time. Note that this only kicks in
     // on DNS server failure, and if a record is deliberately deleted,
     // that will return NXDOMAIN and not be cached.
-    TRC_ERROR("Failed to retrieve record for %s: %s", domain.c_str(), ares_strerror(status));
+    TRC_WARNING("Failed to retrieve record for %s: %s", domain.c_str(), ares_strerror(status));
 
     if (status == ARES_ENOTFOUND)
     {
--- a/src/httpclient.cpp
+++ b/src/httpclient.cpp
@@ -531,8 +531,8 @@ HTTPCode HttpClient::send_request(RequestType request_type,
     }
     else
     {
-      TRC_ERROR("%s failed at server %s : %s (%d) : fatal",
-                url.c_str(), remote_ip, curl_easy_strerror(rc), rc);
+      TRC_WARNING("%s failed at server %s : %s (%d) : fatal",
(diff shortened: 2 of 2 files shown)


dimbleby:
Hopefully the last in this series of turning-errors-into-warnings.

As usual, background is that we monitor our logs for errors, and this is less useful if they can be provoked by network glitches or third-party oddities.

Arguably this case is a bit less clear-cut than some of the other similar ones I've submitted - we actually would like to know about it eg if some third-party server is permanently failed.  But that's probably better detected by larger-scope monitoring of service state rather than by individual error logs for each failed query - and certainly what we actually experience currently is just the occasional log that's more annoying than useful.

sebrexmetaswitch:
I've picked up looking at this pull request too.

I think I'm happy with the change to the dnscachedresolver - if we fail to contact the DNS server, we can normally still proceed (as we'll just keep using the cached record) so I'm OK to move that to a warning.

I don't agree with the change to httpclient though. If we fail to contact an HTTP server, then that's an error. We can't continue processing without that, and we'll return an error code, so we should be logging an error.

So I'm not happy for this to be merged as-is.

dimbleby:
I think you're wrong - I'll write down why, and if you don't find it convincing please give me a call and let's discuss...

A popular convention, and one which I'd strongly encourage in Clearwater, is that ERROR logs should require operator intervention.  (eg see all of the leading answers [here](http://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels)).

(Eg we monitor /var/log/sprout and trigger email alerts when we see ERRORs, in the expectation that they'll be telling us something important).

Now if you only use an HTTP client to talk to other 'internal' elements (eg sprout -> homestead) then it may well make sense to say that a failure is an ERROR.

However, we have a use case where we talk to an external server and if that goes down, I don't want to be woken up to be told about it - rather we write our code to do something sensible on failure.  It's an HTTP error, but not an ERROR.

So what I'm saying is failure in an HTTP client may or may not be an ERROR - it should be up to the caller to decide and make the log, if appropriate.  The HTTP client itself should just return an error code.

sebrexmetaswitch:
Fine. I think I'm still not 100% convinced, but I don't really mind too much. (From my point of view, whether or not you want to get woken up to be told about an HTTP error is irrelevant. If that happens because you're watching /var/log/sprout for error logs, then your process is wrong. And I don't think anyone's agreed to restraining the kind of things that are logged at error level, so this may fix the immediate issue for you now but I suspect that you'll hit it again with something else in future).

But we have both warning and error logs enabled by default, so I don't really mind, so go ahead and merge it :)

dimbleby:
thanks!

PR #592, opened 2016-12-06
[Reviewer: Seb] Make cassandra error more precise

eleanor-merry:
Make the SAS log for when Cassandra fails to read at consistency level TWO more specific (rather than having a generic SAS log for consistency errors).

Fixes https://github.com/Metaswitch/homestead/issues/293

Tested live


sebrexmetaswitch:
LGTM

PR #594, opened 2016-12-08
[Reviewer: Ellie] Better SAS logs for memcached SET errors

sebrexmetaswitch:
We can get a string from the memcached rc, so use it in the SAS log

**testing**
 - live tested

### DIFF
--- a/src/memcachedstore.cpp
+++ b/src/memcachedstore.cpp
@@ -724,8 +724,7 @@ Store::Status TopologyAwareMemcachedStore::set_data(const std::string& table,
                                                     int expiry,
                                                     SAS::TrailId trail)
 {
-  Store::Status status = Store::Status::OK;
-  std::string error_string = "";
+  Store::Status status = Store::Status::ERROR;
 
   TRC_DEBUG("Writing %d bytes to table %s key %s, CAS = %ld, expiry = %d",
             data.length(), table.c_str(), key.c_str(), cas, expiry);
@@ -834,10 +833,10 @@ Store::Status TopologyAwareMemcachedStore::set_data(const std::string& table,
 
       if (!memcached_success(rc))
       {
-        error_string = memcached_strerror(replicas[replica_idx], rc);
+        // Make a log,then move on to the next replica
         TRC_DEBUG("memcached_cas command failed, rc = %d (%s)\n%s",
                   rc,
-                  error_string.c_str(),
+                  memcached_strerror(replicas[replica_idx], rc),
                   memcached_last_error_message(replicas[replica_idx]));
       }
     }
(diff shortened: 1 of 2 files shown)


src/memcachedstore.cpp
```diff
@@ -821,6 +822,7 @@ Store::Status TopologyAwareMemcachedStore::set_data(const std::string& table,
     {
       // This is an update to an existing record, so use memcached_cas
       // to make sure it is atomic.
+       /*
```
eleanor-merry: Uh - I don't think you meant to comment this out

src/memcachedstore.cpp
```diff
@@ -1334,6 +1339,7 @@ Store::Status TopologyNeutralMemcachedStore::set_data(const std::string& table,
     {
       SAS::Event err(trail, SASEvent::MEMCACHED_SET_FAILED, 0);
       err.add_var_param(fqkey);
+      err.add_var_param(memcached_strerror(NULL, rc));
```
eleanor-merry: A TRC_DEBUG would be useful here as well.

eleanor-merry: Does NULL work here?

src/memcachedstore.cpp
```diff
@@ -830,12 +832,14 @@ Store::Status TopologyAwareMemcachedStore::set_data(const std::string& table,
                             memcached_expiration,
                             flags,
                             cas);
-
+*/
+      rc = MEMCACHED_TIMEOUT;
       if (!memcached_success(rc))
       {
+        error_string = memcached_strerror(replicas[replica_idx], rc);
```
eleanor-merry: This section was quite hard to review, and I think it's because the existing code is in a weird order. You're going to look at improving this.

src/memcachedstore.cpp
sebrexmetaswitch: Nope, didn't mean to...

src/memcachedstore.cpp
sebrexmetaswitch: Yeah, we're just converting from a number to a string. I had a look at memcached and the first parameter is unused (no idea why it even exists)

sebrexmetaswitch: There's one below already

eleanor-merry: Not with the error string,

sebrexmetaswitch:
I've made those markups now, including making the similar change for the other memcached failure cases (get and delete).

Can you review again please?

src/memcachedstore.cpp
```diff
@@ -884,39 +880,52 @@ Store::Status TopologyAwareMemcachedStore::set_data(const std::string& table,
...
     if (trail != 0)
     {
-      SAS::Event err(trail, SASEvent::MEMCACHED_SET_FAILED, 0);
+      SAS::Event err(trail, SASEvent::MEMCACHED_SET_CONTENTION, 0);
       err.add_var_param(fqkey);
       SAS::report_event(err);
     }
 
-    update_vbucket_comm_state(vbucket, FAILED);
+    // Commms are woking if there was data contention
```
eleanor-merry: *working

PR #596, opened 2016-12-19
Bifrost httpclient

oa-metaswitch:
This pull request consists of two modifications to the http client and stack. They are:

1) Adding an optional socket-callback function, used to get the http client to use clearwater-socket-factory.
2) Modifying the http stack to add the content of X-Real-Ip and X-Real-Port to SAS logs in stead of the address of the socket. This is needed in cases where we have a reverse proxy obscuring an external UE's IP address.


mirw:
I haven't reviewed the code in detail, but, at a high level,

* why are we modifying SasLogger rather than subclassing it to handle X-Real-Ip/X-Real-Port?
* should we be overriding the socket creation mechanism by subclassing (as we have to do that anyway) rather than adding a callback function?

oa-metaswitch:
Good points. I think subclassing is a better solution for X-Real-Ip, since that issue really only affects http-stacks where the client isn't also logging to SAS. I'm more ambivalent when it comes to the socket creation mechanism, though I don't have a strong opinion either way.

If I do subclass these, do you want them in cpp-common?

mirw:
Yes, I think it would be useful to have the subclasses in cpp-common - it sounds like they should be generic enough.

oa-metaswitch:
Subclasses have been submitted. Ellie: care to take a look? :)

src/httpclient.cpp
```diff
@@ -1059,3 +1066,42 @@ int HttpClient::Recorder::record_data(curl_infotype type,
...
+  curl_easy_setopt(curl, CURLOPT_OPENSOCKETDATA, socket);
+  curl_easy_setopt(curl, CURLOPT_SOCKOPTFUNCTION, sockopt_callback);
+
+  return (void*) socket;
+}
+
+void CallbackHttpClient::cleanup_host_context(void* host_context)
+{
+  // Delete the socket created above after the connection has been completed.
+  delete (int*)host_context;
```
mirw: Don't you need to close this socket?

...or actually maybe you only need to close the socket if we **haven't** already given it to curl?

src/httpstack.cpp
```diff
@@ -464,6 +465,33 @@ bool HttpStack::Request::get_remote_ip_port(std::string& ip, unsigned short& por
...
+
+  if (real_ip != "")
+  {
+    rc = true;
+    ip.assign(real_ip);
+    std::string port_s = header("X-Real-Port");
+
+    if (port_s != "")
+    {
+      port = (short)std::stoi(port_s);
```
oa-metaswitch: Yup, I'll keep it for consistency.

src/httpstack.cpp
```diff
@@ -464,6 +465,33 @@ bool HttpStack::Request::get_remote_ip_port(std::string& ip, unsigned short& por
...
+    }
+    else
+    {
+      // This should only happen if the reverse proxy has been mis-configured.
+      TRC_WARNING("Recieved a proxied request with missing X-Real-Port-header.");
```
oa-metaswitch: That's true. What's your preferred solution? I don't want to swamp the logs, but at the same time this feels like it should get logged somewhere.

src/httpstack.cpp
```diff
@@ -464,6 +465,33 @@ bool HttpStack::Request::get_remote_ip_port(std::string& ip, unsigned short& por
...
+bool HttpStack::Request::get_x_real_ip(std::string& ip, unsigned short& port)
+{
+  bool rc = false;
+  std::string real_ip = header("X-Real-Ip");
+  TRC_DEBUG("Real IP: %s", real_ip.c_str());
+
+  if (real_ip != "")
+  {
+    rc = true;
+    ip.assign(real_ip);
```
oa-metaswitch: True. I was following the pattern of get_ip_port below, however, they're using libevent and I'm not.

oa-metaswitch:
Matt: I've implemented the changes we discussed yesterday. Care to take a look?

src/httpstack.cpp
```diff
@@ -37,8 +37,11 @@
 #include "httpstack.h"
 #include <cstring>
 #include <sys/stat.h>
+#include <climits>
+#include <algorithm>
 #include "log.h"
 
+
```
mirw: Superfluous blank line.

mirw:
Fine.

PR #598, opened 2017-01-10
Made the error cases in get_x_real_ip_port clearer

oa-metaswitch:
Not really a bug fix, but these modifications make it a lot clearer what happens when proxied_sas_logger receives a request with an x-real-ip header, but missing an x-real-port header. 

The previous implementation relied on the default value of shorts being 0, which works, but hides what the error handling is doing.

### DIFF
--- a/src/httpstack.cpp
+++ b/src/httpstack.cpp
@@ -479,6 +479,8 @@ bool HttpStack::Request::get_x_real_ip_port(std::string& ip, unsigned short& por
     ip = real_ip;
     std::string port_s = header("X-Real-Port");
     int port_i;
+    // We have a real IP, if we cannot get a real port we fail softly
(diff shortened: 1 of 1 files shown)


src/httpstack.cpp
```diff
@@ -488,16 +490,12 @@ bool HttpStack::Request::get_x_real_ip_port(std::string& ip, unsigned short& por
       }
       catch (...)
       {
-        port_i = 0;
+        port_i = -1;
       }
-      if (port_i > 0 && port_i < USHRT_MAX)
+      if (port_i >= 0 && port_i < USHRT_MAX)
```
eleanor-merry: I'd put this if statement in the try (to avoid setting the port_i to -1).

Issue #394, opened 2015-10-29
Need to commonise SNMP tables 


johadalin:
With added work introducing the continuous_increment_table, there is not even more reason to commonise the SNMP tables: the continuous increment table, event accumulator table and continuous accumulator table.
They share a substantial amount of internal code, and so creating a new base class for them should work, but there may be conflict in their construction methods, due to the need to separate the SNMP namespace from the rest of the code.

rkd-msw:
I did a bit of thinking about this last night. I wonder if we could restructure the SNMP tables so that you have:

- "datastore" classes which know nothing about SNMP, but provide methods to let you put in data, are responsible for turning the data into a tabular-but-not-SNMP-specific format (including managing current/previous time periods), and can be queried (e.g. what's the row at this index?)
- an SNMP table class (probably just one) which is constructed with an OID, list of index types, list of column types, and a datastore pointer - this is responsible for making the netsnmp calls to set it all up properly, and then querying the datastore to answer any SNMP requests
- (we do away with the separate Row class, since I think this is extra flexibility and complexity we don't use or really need)

I think this is nicer in a couple of ways:

- creating a new table type just requires subclassing Datastore, whereas today you need a Row subclass, a Table subclass, and some sort of non-SNMP-referencing interface class
- you have fewer classes to reason about - no parallel hierarchies of Table and Row
- because the Datastore classes don't know about SNMP, you don't need as many contortions to keep the SNMP header files from polluting things

@mirw , thoughts as lead architect? (Not urgent.)

mirw:
Sorry for the delay.  Sounds sensible.

johadalin:
As described in the PR and Issue linked here, I believe that this is becoming a more important issue. I think it actually sits between high and medium priority, as i think we want to plan some work for the next major release to revamp all of this code, rather than treating this as a normal bug, but i think that we shouldn't just let this slip.

PR #599, opened 2017-01-13
[Reviewer: Matt] Thread safer stats

johadalin:
All the areas we discussed, and rested with locks, are now using atomic functions to ensure we set the right value, or re-do the calculation with a new starting point.

In the continuous increment table, to do this, I have moved where we set the new value for `current_value` up to before all of the average, sum etc. calculations, which occur at the top of `accumulate_internal`. This means that they will be calculated using the new `current_value` rather than the old `current_value`. 
I'm not sure if this is a big issue for the statistics, are you able to comment at all on that?

In testing, this now appears to resolve the three Chronos stats issues:
https://github.com/Metaswitch/chronos/issues/328
https://github.com/Metaswitch/chronos/issues/329
https://github.com/Metaswitch/chronos/issues/330

### DIFF
--- a/src/snmp_continuous_increment_table.cpp
+++ b/src/snmp_continuous_increment_table.cpp
@@ -118,8 +118,6 @@ private:
     // using the new current value.
     do
     {
-      new_value = 0;
-
       if (increment_total)
       {
         new_value = current_value + value_delta;
@@ -130,15 +128,20 @@ private:
         // and decrement, or leave value as default of 0.
         if (value_delta < current_value)
         {
-         new_value = current_value - value_delta;
+          new_value = current_value - value_delta;
+        }
+        else
+        {
+          new_value = 0;
         }
       }
-    }while(!current_data->current_value.compare_exchange_weak(current_value, new_value));
+    } while (!current_data->current_value.compare_exchange_weak(current_value, new_value));
 
-    accumulate_internal(current_data, new_value, now);
+    accumulate_internal(current_data, current_value, new_value, now);
   }
 
   void accumulate_internal(ContinuousStatistics* current_data,
+                           uint64_t current_value,
                            uint64_t sample,
                            const struct timespec& now)
   {
@@ -150,7 +153,6 @@ private:
     // current value held, that can be used if the period ends.
     uint64_t time_since_last_update = ((now.tv_sec * 1000) + (now.tv_nsec / 1000000))
                                      - (current_data->time_last_update_ms.load());
-    uint32_t current_value = current_data->current_value.load();
 
     current_data->time_last_update_ms = (now.tv_sec * 1000) + (now.tv_nsec / 1000000);
     current_data->sum += current_value * time_since_last_update;
@@ -200,8 +202,6 @@ ColumnData ContinuousAccumulatorRow::get_columns()
     lwm = 0;
   }
   uint_fast32_t hwm = accumulated->hwm.load();
-  uint_fast64_t time_last_update_ms = accumulated->time_last_update_ms.load();
-  uint_fast64_t time_period_start_ms = accumulated->time_period_start_ms.load();
   uint_fast64_t sum = accumulated->sum.load();
   uint_fast64_t sqsum = accumulated->sqsum.load();
 
@@ -212,24 +212,48 @@ ColumnData ContinuousAccumulatorRow::get_columns()
   // to stop us from updating periods that are now out of date.
   uint64_t time_now_ms = (now.tv_sec * 1000) + (now.tv_nsec / 1000000);
 
-  // As a time period might not begin on a boundary, we must synchronize
-  // the end of the period with a boundary which is why the following line
-  // requires so many interval_ms!
-  uint64_t time_period_end_ms = ((time_period_start_ms + interval_ms) / interval_ms) * interval_ms;
-  uint64_t time_comes_first_ms = std::min(time_period_end_ms, time_now_ms);
+
+  // Get last update from the data structure, and set this update point
+  // making sure we don't conflict with other threads
+  uint64_t time_comes_first_ms;
+  uint_fast64_t time_period_start_ms;
+  uint64_t time_period_end_ms;
+  uint_fast64_t time_last_update_ms = accumulated->time_last_update_ms.load();
+  do
+  {
+    // Make sure that we have the correct period start time, in case we loop.
+    time_period_start_ms = accumulated->time_period_start_ms.load();
+
+    // As a time period might not begin on a boundary, we must synchronize
+    // the end of the period with a boundary which is why the following line
+    // requires so many interval_ms!
+    time_period_end_ms = ((time_period_start_ms + interval_ms) / interval_ms) * interval_ms;
+
+    // Choose the earlier of these as the new update point
+    time_comes_first_ms = std::min(time_period_end_ms, time_now_ms);
+  } while (!accumulated->time_last_update_ms.compare_exchange_weak(time_last_update_ms, time_comes_first_ms));
 
   uint64_t time_since_last_update_ms = time_comes_first_ms - time_last_update_ms;
-  accumulated->time_last_update_ms.store(time_comes_first_ms);
   uint64_t period_count = (time_comes_first_ms - time_period_start_ms);
 
   if (period_count > 0)
   {
     // Calculate the average and the variance from the stored average/time of last updated
-    // and sum-of-squares.
-    accumulated->sum+=(time_since_last_update_ms * current_value);
-    accumulated->sqsum+=(time_since_last_update_ms * current_value * current_value);
-    avg = sum / period_count;
-    variance = ((sqsum * period_count) - (sum * sum)) / (period_count * period_count);
+    // and sum-of-squares, and reinsert into the data safely.
+    uint64_t new_sum;
+    do
+    {
+      new_sum = sum + (time_since_last_update_ms * current_value);
+    } while (!accumulated->sum.compare_exchange_weak(sum, new_sum));
+
+    uint64_t new_sqsum;
+    do
+    {
+      new_sqsum = sqsum + (time_since_last_update_ms * current_value * current_value);
+    } while (!accumulated->sum.compare_exchange_weak(sqsum, new_sqsum));
+
+    avg = new_sum / period_count;
+    variance = ((new_sqsum * period_count) - (new_sum * new_sum)) / (period_count * period_count);
   }
 
   // Construct and return a ColumnData with the appropriate values
--- a/src/snmp_continuous_increment_table.cpp
+++ b/src/snmp_continuous_increment_table.cpp
@@ -40,7 +40,6 @@
 #include "snmp_internal/snmp_time_period_table.h"
 #include "snmp_continuous_increment_table.h"
 #include "limits.h"
-std::mutex _mutex;
 
 std::mutex _cont_inc_mutex;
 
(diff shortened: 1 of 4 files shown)


src/snmp_continuous_increment_table.cpp
```diff
@@ -104,36 +104,42 @@ class ContinuousIncrementTableImpl: public ManagedTable<ContinuousAccumulatorRow
...
+        new_value = current_value + value_delta;
       }
       else
       {
-        total -= value;
+        // Check to ensure the value to accumulate will not be negative,
+        // and decrement, or leave value as default of 0.
+        if (value_delta < current_value)
+        {
+         new_value = current_value - value_delta;
```
mirw: Indentation.

src/snmp_continuous_increment_table.cpp
```diff
@@ -104,36 +104,42 @@ class ContinuousIncrementTableImpl: public ManagedTable<ContinuousAccumulatorRow
...
+        }
       }
-    }
-    accumulate_internal(current_data, total, now);
+    }while(!current_data->current_value.compare_exchange_weak(current_value, new_value));
```
mirw: We normally leave a space between `}` and `while`.

src/snmp_continuous_increment_table.cpp
```diff
@@ -104,36 +104,42 @@ class ContinuousIncrementTableImpl: public ManagedTable<ContinuousAccumulatorRow
...
+    do
     {
-      total += value;
-    }
-    else
-    {
-      // Check to ensure the value to accumulate will not be negative,
-      // and then set to 0 or decrement appropriately.
-      if (total < value)
+      new_value = 0;
```
mirw: I'd prefer to put this in the `else` branch of the `if (value_data < current_value)` block below - that means that we're setting new_value to 0 when appropriate, rather than leaving it as the "default" (which it isn't really).

src/snmp_continuous_increment_table.cpp
mirw: Updating current_data->current_value at this point means that accumulate_internal's logic is wrong - it is supposed to add on the amount of time spent with the old value before taking the new value.  (We're trying to track the "area under the graph" of sample value against time.)

src/snmp_continuous_increment_table.cpp
```diff
@@ -221,10 +226,8 @@ ColumnData ContinuousAccumulatorRow::get_columns()
...
     // Calculate the average and the variance from the stored average/time of last updated
     // and sum-of-squares.
-    sum += time_since_last_update_ms * current_value;
-    accumulated->sum.store(sum);
-    sqsum += time_since_last_update_ms * current_value * current_value;
-    accumulated->sqsum.store(sqsum);
+    accumulated->sum+=(time_since_last_update_ms * current_value);
+    accumulated->sqsum+=(time_since_last_update_ms * current_value * current_value);
     avg = sum / period_count;
     variance = ((sqsum * period_count) - (sum * sum)) / (period_count * period_count);
```
mirw: Surely `avg` and `variance` are now wrong, as they're calculated based on the old values, not the new ones.  You should remove the `sum` and `sqsum` variables above if you're not going to keep them up to date.

src/snmp_continuous_increment_table.cpp
```diff
@@ -221,10 +226,8 @@ ColumnData ContinuousAccumulatorRow::get_columns()
   {
```
mirw: The `accumulated->time_last_update_ms.store` line just above looks vary dangerous.  We might not need to fix this right now, but please check what the implications will be, and if you don't fix it now, make sure it's tracked by an issue.

src/timer_counter.cpp
```diff
@@ -117,32 +117,34 @@ void TimerCounter::refresh_statistics(SNMP::ContinuousStatistics* data, timespec
...
   uint64_t current_value = data->current_value.load();
-  uint64_t sum = data->sum.load();
-  uint64_t sqsum = data->sqsum.load();
 
-  sum += current_value * time_since_last_update;
-  sqsum += current_value * current_value * time_since_last_update;
-
-  data->sum.store(sum);
-  data->sqsum.store(sqsum);
+  data->sum+=(current_value * time_since_last_update);
```
mirw: We normally leave spaces around `+=` (and I wouldn't bother with brackets here either).

src/snmp_continuous_increment_table.cpp
```diff
@@ -221,10 +226,8 @@ ColumnData ContinuousAccumulatorRow::get_columns()
...
     // Calculate the average and the variance from the stored average/time of last updated
     // and sum-of-squares.
-    sum += time_since_last_update_ms * current_value;
-    accumulated->sum.store(sum);
-    sqsum += time_since_last_update_ms * current_value * current_value;
-    accumulated->sqsum.store(sqsum);
+    accumulated->sum+=(time_since_last_update_ms * current_value);
+    accumulated->sqsum+=(time_since_last_update_ms * current_value * current_value);
```
mirw: We normally leave spaces around += (and I wouldn't bother with brackets here either).

src/snmp_infinite_scalar_table.cpp
```diff
@@ -73,7 +74,7 @@ namespace SNMP
   protected:
     static const uint32_t max_row = 1;
     static const uint32_t max_column = 2;
-    std::map<std::string, uint32_t> _scalar_counters;
+    std::map<std::string, std::atomic<std::uint32_t>> _scalar_counters;
```
mirw: Although the value of this map is not thread-safe, maps themselves are not.  Unfortunately, I think this might be a case where we genuinely need a lock.  Please check with @AlexHockey in case there's a cunning new C++ solution, though.

src/timer_counter.cpp
```diff
@@ -117,32 +117,34 @@ void TimerCounter::refresh_statistics(SNMP::ContinuousStatistics* data, timespec
...
+  data->sqsum+=(current_value * current_value * time_since_last_update);
   data->time_last_update_ms.store(time_comes_first_ms);
```
mirw: This feels risky.  Is it safe?

src/timer_counter.cpp
```diff
@@ -117,32 +117,34 @@ void TimerCounter::refresh_statistics(SNMP::ContinuousStatistics* data, timespec
...
-  // If value_delta would cause new_value to underflow, leave new_value as 0.
-  if ((value_delta > 0) || ((uint64_t)abs(value_delta) < current_value))
+  // Attempt to calculate the new value, and set it to the atomic beneath.
+  // If the atomic value has changed in the mean time, repeat the calculations
+  // using the new current value.
+  do
   {
-    new_value = current_value + value_delta;
-  }
+    new_value = 0;
```
mirw: As above, it's better to set this explicitly rather than leaving it to default, as then people extending the code will have to think about whether they need to set it or not.

src/timer_counter.cpp
```diff
@@ -117,32 +117,34 @@ void TimerCounter::refresh_statistics(SNMP::ContinuousStatistics* data, timespec
...
+
+    // Ensure value_delta is less than or equal to the current value if negative.
+    // If value_delta would cause new_value to underflow, leave new_value as 0.
+    if ((value_delta > 0) || ((uint64_t)abs(value_delta) <= current_value))
```
mirw: Why do we do `abs()` here?  Can't we just write `-current_value <= value_delta`?

src/timer_counter.cpp
```diff
@@ -117,32 +117,34 @@ void TimerCounter::refresh_statistics(SNMP::ContinuousStatistics* data, timespec
...
+    {
+      new_value = current_value + value_delta;
+    }
+
+  }while(!data->current_value.compare_exchange_weak(current_value, new_value));
```
mirw: Formatting as above.

src/timer_counter.cpp
johadalin:
```
../modules/cpp-common/src/timer_counter.cpp:140:49: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if ((value_delta > 0) || (-current_value <= value_delta))
                                                 ^
```
Because values are stored as uint64_ts, it doens't like the comparison.

src/timer_counter.cpp
johadalin: Not sure i follow why this might be risky

src/snmp_continuous_increment_table.cpp
johadalin: Yes, this is now broken. I'll make sure that it isn't.

Is it worth breaking all of the stats out into separate do..while loops, as with above, so that we can safely increment the values?

src/snmp_continuous_increment_table.cpp
johadalin: This is what i was unsure of how much it might matter. We're only changing values by small amounts, so i think it's not huge, but i'll modify it so that the old `current_value` is correctly used for calculating the other stats.

src/snmp_continuous_increment_table.cpp
johadalin: I think this comment has somehow got to the wrong place. i assume you mean line 222.
I don't think the implications of this are too bad, as any time we're likely to be caught in threading issues, the difference between times can't be very great, and will only lead to minor discrepancies in stats that will be smoothed out over time.
I think that these tables need reviewing and re-working in future work, to commonise, simplify, and make sure that they are genuinely safe and accurate, so happy to raise an issue (and possibly story) to cover that.

mirw: Yes, I definitely added this just below line 222 - I think Github has got confused.

What *will* happen with the current code?  I can believe that it may be acceptable, but I'd like to check that rather than just saying "it probably won't matter".

src/snmp_continuous_increment_table.cpp
mirw: I'm not quite sure what the do...while loops would do - let's discuss.

src/timer_counter.cpp
mirw: We're storing time_last-update_ms without first checking whether it's been changed by another thread.  What would happen if we got that wrong?

src/timer_counter.cpp
mirw: Fine, so you can just cast, can't you?  `abs()` seems unnecessary.

johadalin:
There are still a few points in the code where there is the potential for thread conflicts, e.g. at https://github.com/Metaswitch/cpp-common/blob/9681ab734f9910683d60c7dc637e37fe490dc3c4/src/snmp_continuous_increment_table.cpp#L154, we are getting and setting the time_last_update_ms in a non-safe manner. Moreover, this check should really happen at the same time that we get-set the current_value change, so that we are actually calculating the new values from the correct point in time.

A while back we raised https://github.com/Metaswitch/cpp-common/issues/394 to track the need to commonise these snmp tables, and since then yet more un-commonised tables have been created (notably, the continuous_accumulator_by_scope table). I think we need to do some higher level re-thinking and re-working of all of these tables so as to make them simpler to maintain, and to make sure that they are genuinely thread-safe, and not going to lead to inconsistencies in the values they calculate. 

I am going to raise a new issue, to cover the further section in the continuous_increment_table described above. However, as the deltas between samples are always going to be 1 (currently chronos only increments/decrements by 1 for this table), and the time differences we would be dealing with in the race-condition circumstances are going to be very narrow, the impact that these inaccuracies could have will be minor. As such, i don't think this work is critical.

I will also then link that issue through to #394, and raise the priority on that. I believe that we should work on that once we have resolved the current critical issues, rather than treating it as low priority. If we do not commonise the tables at this point, but continue patching threading issues, the tables will diverge to a point that it will be very difficult to re-integrate and maintain them.

PR #600, opened 2017-01-13
[Reviewer: AJH] Rename path to url_tail

sathiyan-sivathas:
This path variable can contain the path, query and fragment parts, so it is more correct to call it the url_tail.


AlexHockey:
:+1:

PR #605, opened 2017-01-18
Add a mock HTTP client. 

AlexHockey:
I've written some UTs in sprout that use this class so I'm confident it works.


sathiyan-sivathas:
:+1:

PR #606, opened 2017-01-20
[Reviewer: Adam] Set persistent file somewhere that the cpp process can write to

eleanor-merry:
Write the persistent SNMP file to somewhere Sprout can write to. We don't actually use this - but we get errors logs on shutdown for the current location. Fixes https://github.com/Metaswitch/sprout/issues/1475. Tested live


johadalin:
:+1:

Issue #609, opened 2017-01-25
Memcached not replicating data in a GR system


eleanor-merry:
#### Symptoms
On a GR system with 50ms latency between the two sites (i.e a 100ms round-trip time) I noticed that registrations weren't being replicated across to the remote site - the initial get from the remote site always failed with `error 47 (SERVER HAS FAILED AND IS DISABLED UNTIL TIMED RETRY)`. 

#### Impact
Any call between subscribers that had registered in different sites failed. No protection against site failure.

#### Release and environment 
From at least Vulpix

eleanor-merry:
The issue is that we had a hard coded limit of 50ms on any memcached connection - the fix is to increase this value. We've gone for a large value here so that we won't realistically hit timeouts (and we rely on the existing timeout code to handle retries, blacklisting, etc..)

eleanor-merry:
@rkd-msw can I pass this to you to see if you hit this again when you do more testing of GR?

rkd-msw:
Yep, will do.

rkd-msw:
@eleanor-merry - are there any other fixes needed to make GR work, or do you expect everything will now be fine tomorrow?

eleanor-merry:
Somewhere in between - I don't know of any other fixes that need to go in now.

rkd-msw:
It's currently working - calls are getting through with 100ms latency (200ms round-trip time) but I saw some memcached timeouts when I started testing (with either 50ms or 100ms latency - unfortnately I don't know which).

Specifically, I've seen calls fail with 480 errors, with the SAS logs "Sending get request to memcached for key reg\\sip:50232007@gr-stress.clearwater.test" (at 12:08:32.159) and "Memcached get request failed for key reg\\sip:50232007@gr-stress.clearwater.test

Error: A TIMEOUT OCCURRED" (at 12:08:32.360).

I don't understand why these requests are timing out - but that's partly because the SAS logs don't contain memcached IP addresses so I can't tell where it's reading from. I guess that we haven't found the data locally so we're hitting https://github.com/Metaswitch/sprout/blob/ee05b98c9f533fc8f34c19d2ccb122f5533bed51/src/scscfsproutlet.cpp#L234 and reading cross-site - but that shouldn't time out and the data should have been replicated to this site....

I've attached:

* SAS trace of a failed call
* SAS trace of the registration corresponding to the user we failed to find on the failed call
* a diags package from the system that handled the REGISTER (in site1)
[...]

PR #610, opened 2017-01-25
[Reviewer: Seb] Increase timeout

eleanor-merry:
Fix for #609 - increases the memcached timeout as discussed.


mirw:
Please can you record in the issue why this is the right value?  "As discussed" won't help when we come to this in a few years time.  Thanks!

PR #611, opened 2017-01-26
[Reviewer: Adam] Add a timeout to cw_stat

eleanor-merry:
Add a timeout to the cw_stat call (as otherwise it blocks forever). Part of the fix for issue https://github.com/Metaswitch/astaire/issues/76


johadalin:
LGTM

PR #612, opened 2017-01-31
Add a functor thread pool

AlexHockey:
Contains a variety of thread pool enhancements:

* A thread pool variant that can take and run any callable object. 
* Allows work items to be added to the pool by move rather than by copy. 
* Adds a `run_once` method to allow the thread pool to be driven in UTs.


sathiyan-sivathas:
👍

PR #614, opened 2017-02-03
[Reviewer: Seb] Change to have different timeouts for local and remote sites

eleanor-merry:
Change to memcached timeouts so that local and remote sites have different timeouts, rather than not having an effective timeout.


sebrexmetaswitch:
Looks good

PR #615, opened 2017-02-03
[Reviewer: Graeme] Be tolerant to not finding per-message data (e.g. in error scenarios)

mirw:
Graeme,

Please can you review this fix to https://github.com/Metaswitch/homestead/issues/392?  Basically, the pmd structure can be NULL, so we should avoid dereferencing it if so.  We also TRC_WARNING so that we can track this - it should happen very rarely - if it starts happening frequently, we may have uncorrelated SAS trails.

UTs pass.  I'm just kicking the tyres live.

The other tidy-ups are to misleading/imprecise trace statements and superfluous blank lines.

Matt


mirw:
Wrong branch - closing.

PR #620, opened 2017-02-07
Optionally obscure Http bodies in SAS logs

alango:
Makes changes to the HttpStack and the HttpClient so that the bodies of Http requests and responses can be optionally obscured in SAS:

- ProxiedSasLogger (which is currently only used by Bifrost) is replaced with ProxiedPrivateSasLogger which doesn't log any bodies to SAS.

- A 'should_omit_body' variable is added to HttpClient which defaults to false. If enabled, then http bodies are again obscured when being reported to SAS.

### DIFF
--- a/src/httpclient.cpp
+++ b/src/httpclient.cpp
@@ -894,13 +894,14 @@ void HttpClient::sas_add_ip_addrs_and_ports(SAS::Event& event,
   sas_add_port(event, curl, CURLINFO_LOCAL_PORT);
 }
 
-void HttpClient::sas_log_http_req(SAS::TrailId trail,
-                                  CURL* curl,
-                                  const std::string& method_str,
-                                  const std::string& url,
-                                  const std::string& request_bytes,
-                                  SAS::Timestamp timestamp,
-                                  uint32_t instance_id)
+void HttpClient::log_req_event(SAS::TrailId trail,
+                               CURL* curl,
+                               const std::string& method_str,
+                               const std::string& url,
+                               const std::string& request_bytes,
+                               SAS::Timestamp timestamp,
+                               uint32_t instance_id,
+                               bool omit_body)
 {
   if (_sas_log_level != SASEvent::HttpLogLevel::NONE)
   {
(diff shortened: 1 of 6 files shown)


src/httpstack.cpp
```diff
@@ -727,3 +750,19 @@ void HttpStack::ProxiedSasLogger::add_ip_addrs_and_ports(SAS::Event& event, Requ
     // LCOV_EXCL_STOP
   }
 }
+
+void HttpStack::ProxiedPrivateSasLogger::sas_log_rx_http_req(SAS::TrailId trail,
+                                                      HttpStack::Request& req,
```
eleanor-merry: Alignment

src/httpstack.cpp
```diff
@@ -552,7 +553,29 @@ void HttpStack::SasLogger::log_req_event(SAS::TrailId trail,
...
+    {
+      // We are omitting the body but there wasn't one in the messaage. Just log
+      // the header.
+      event.add_compressed_param(req.get_rx_header(), &SASEvent::PROFILE_HTTP);
+    }
+    else
+    {
+      // There was a body that we need to omit. Add a fake body to the header
+      // explaining that the body was intentionally not logged.
+      event.add_compressed_param(req.get_rx_header() + "<Body present but not logged>",
```
eleanor-merry: Does this need the `\r\n\r\n`? Can this string be pulled out to a constant?

src/httpstack.cpp
```diff
@@ -552,7 +553,29 @@ void HttpStack::SasLogger::log_req_event(SAS::TrailId trail,
...
   event.add_var_param(req.full_path());
-  event.add_compressed_param(req.get_rx_message(), &SASEvent::PROFILE_HTTP);
+
+  if (!omit_body)
+  {
+    event.add_compressed_param(req.get_rx_message(), &SASEvent::PROFILE_HTTP);
+  }
+  else
+  {
+    // LCOV_EXCL_START
```
eleanor-merry: Why isn't this covered by the UTs?

src/httpclient.cpp
```diff
@@ -894,6 +896,25 @@ void HttpClient::sas_add_ip_addrs_and_ports(SAS::Event& event,
   sas_add_port(event, curl, CURLINFO_LOCAL_PORT);
 }
 
+std::string HttpClient::get_obscured_message_to_log(const std::string& message)
+{
+  std::string message_to_log;
+  std::size_t body_pos = message.find("\r\n\r\n");
```
eleanor-merry: This feels pretty fragile. Can `\r\n\r\n` be missing from the message? Can this be pulled into a constant as well?

mirw:
Are we right to be making this part of HttpClient itself, rather than a function of one of the SasLoggers?  It feels like HttpClient is becoming a bit of a monolith.

alango:
HttpClient doesn't use the SasLogger classes at the moment, making it do so seems like it would be a much bigger change?

src/httpclient.cpp
alango: It will always appear after the headers in an HTTP message, even if there is no body afterwards.

mirw:
A bigger change, but a much more maintainable one!  My nervousness is that if (for example) we wanted to change HttpClient to not log the URI, or not log the headers, we'd end up plumbing yet more Booleans through.  At some stage, we're going to need to do the refactor - the question is how much tech debt we can afford to accumulate before then.

alango:
I've opened a PR to add UTs for this change at https://github.com/Metaswitch/cpp-common-test/pull/41 which I've assigned to @eleanor-merry

alango:
@mirw Realised we never actually resolved this, would you prefer it then if I didn't merge this? Seeing as we only need these changes for the non-Clearwater components at the moment, we could just use our own branch of cpp-common for now.

mirw:
It's fine.  We just need to bear this in mind next time we make a change.

PR #622, opened 2017-02-20
Add a serializable bloom filter implementation

AlexHockey:
Sathiyan, please can you review this bloom filter code? Hopefully self explanatory but let me know if not. siphash.cpp is taken from the reference implementation of SipHash, so it's not worth reviewing that file!

Tested in UT.

### DIFF
--- a/include/bloom_filter.h
+++ b/include/bloom_filter.h
@@ -40,6 +40,8 @@
 #include "rapidjson/writer.h"
 #include "rapidjson/document.h"
 
+#include <vector>
+
 class BloomFilter
 {
 public:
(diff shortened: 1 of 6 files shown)


include/bloom_filter.h
```diff
@@ -0,0 +1,155 @@
+/**
+ * @file bloom_filter.h
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2016  Metaswitch Networks Ltd
```
sathiyan-sivathas: Come on, it's already February 😛

eleanor-merry: +1

PR #623, opened 2017-02-21
Separate PrivateSasLogger from ProxiedPrivateSasLogger

alango:
As per my email, this is to separate the part of the SAS logger which obscures the bodies of requests from the part that sits behind a reverse proxy.


eleanor-merry:
LGTM

PR #624, opened 2017-02-21
Fix bug that returned ttl of 0

KristaHyer:
This relates to issue #265.

Ttl was originally set to 0, so when the minimum ttl was found it was always 0.
To fix - first time ttl is set no minimum is found, therefore zero is always overwritten.

Have written some tests to confirm this works - all tests pass.


src/diameterresolver.cpp
```diff
@@ -179,7 +185,8 @@ void DiameterResolver::resolve(const std::string& realm,
     }
     else
     {
-      a_resolve(host, _address_family, DEFAULT_PORT, DEFAULT_TRANSPORT, max_targets, targets, ttl, 0);
+      a_resolve(host, _address_family, DEFAULT_PORT, DEFAULT_TRANSPORT, max_targets, targets, new_ttl, 0);
+      ttl = set_ttl? std::min(ttl, new_ttl) : new_ttl;
```
eleanor-merry: space before the ? please

eleanor-merry: Worth adding the set_ttl here as well (in case we add some more branches later on, and don't notice that this branch is missing the set_ttl).

PR #627, opened 2017-03-07
[Reviewer: Adam] Return HTTP 503 on curl timeout

sebrexmetaswitch:
When we get a curl timeout on an HTTP request, we should report the HTTP return code as 503.

This allows the upstream node to handle the timeout correctly.

This is a fix for issue https://github.com/Metaswitch/sprout/issues/1623

#### Testing
 - live tested by manually dropping packets to Homestead, and verifying that Sprout correctly interprets that as an HTTP 503
 - verified live tests still pass on a deployment with this fix


johadalin:
Yep. Looks good

PR #628, opened 2017-03-08
[Reviewer: Krista] Add wildcard matching functions

eleanor-merry:
This adds some utility functions to help with wildcard parsing

Tested in UTs and live

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -601,6 +601,14 @@ namespace Utils
 
   // Does the passed in address have brackets?
   bool is_bracketed_address(const std::string& address);
+
+  // Checks if a string represents a wildcard URI.
+  bool is_wildcard_uri(std::string wildcard);
+
+  // Checks if two URIs match, including wildcard checking.
+  bool uris_user_match(const std::string& matcher,
+                       const std::string& matchee);
+
 } // namespace Utils
 
 #endif /* UTILS_H_ */
--- a/include/utils.h
+++ b/include/utils.h
@@ -603,11 +603,11 @@ namespace Utils
   bool is_bracketed_address(const std::string& address);
 
   // Checks if a string represents a wildcard URI.
-  bool is_wildcard_uri(std::string wildcard);
+  bool is_wildcard_uri(const std::string& possible_wildcard);
 
   // Checks if two URIs match, including wildcard checking.
-  bool uris_user_match(const std::string& matcher,
-                       const std::string& matchee);
+  bool check_users_equivalent(const std::string& wildcard_user,
+                              const std::string& specific_user);
 
 } // namespace Utils
 
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -713,7 +713,7 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
   }
 }
 
-bool Utils::is_wildcard_uri(std::string wildcard)
+bool Utils::is_wildcard_uri(const std::string& possible_wildcard)
 {
   // This function counts how many !s there are in the URI string
   // before the @ - note, this only checks whether the URI string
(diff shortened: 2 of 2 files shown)


src/utils.cpp
```diff
@@ -711,3 +712,105 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
                          IPAddressType::INVALID;
   }
 }
+
+bool Utils::is_wildcard_uri(std::string wildcard)
```
KristaHyer: I'd rename what you pass in - "wildcard" makes it sound like it's definitely a wildcard.

src/utils.cpp
```diff
@@ -711,3 +712,105 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
...
+                     '!') >= 2);
+}
+
+bool Utils::uris_user_match(const std::string& matcher,
+                            const std::string& matchee)
+{
+  // Quick check if the identities are already the same without looking at the
+  // wildcards. Also check if either string is empty (where we've caught the
+  // case where they're both empty at the start) as this definitely won't match,
+  // and checking now simplifies the error checking logic later.
```
KristaHyer: Make this comment clearer.

src/utils.cpp
```diff
@@ -711,3 +712,105 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
...
+  }
+
+  // Check the start and end parts
+  std::string matcher_start_str = matcher_user[0].substr(0, matcher_start);
+  std::string matcher_end_str = matcher_user[0].substr(matcher_end + 1);
+  std::string matchee_start_str = matchee_user[0].substr(0, matcher_start);
+  std::string matchee_end_str = matchee_user[0].substr(matchee_start_str.size());
+
+  if ((matchee_start_str != matcher_start_str) ||
+      (matchee_end_str.substr(matchee_end_str.size() - matcher_end_str.size()) !=
```
KristaHyer: I'd pull this logic out of the "if" statement to make the logic clearer, and also so that nobody can jump in at line 789 and use "matchee_end_str" before it's been trimmed down to be just the end.

src/utils.cpp
```diff
@@ -711,3 +712,105 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
...
+  {
+    // Either the start or end of the wildcard didn't directly match the
+    // matchee string
+    return false;
+  }
+
+  // Finally, check against the regex.
+  std::string matcher_part = matcher_user[0].substr(matcher_start + 1,
+                                                     (matcher_end - matcher_start - 1));
+  std::string matchee_part = matchee_end_str.substr(0, (matchee_end_str.size() -
```
KristaHyer: Once you've trimmed it down fully above, the logic to find the middle in the matcher and matchee can be the same as well

include/utils.h
```diff
@@ -601,6 +601,14 @@ namespace Utils
 
   // Does the passed in address have brackets?
   bool is_bracketed_address(const std::string& address);
+
+  // Checks if a string represents a wildcard URI.
+  bool is_wildcard_uri(std::string wildcard);
```
graemerobertson: `const std::string&`?

src/utils.cpp
```diff
@@ -711,3 +712,105 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
...
+  // heavyweight.
+  std::vector<std::string> wildcard_parts;
+  Utils::split_string(wildcard, '@', wildcard_parts, 0, false);
+  return (std::count(wildcard_parts[0].begin(),
+                     wildcard_parts[0].end(),
+                     '!') >= 2);
+}
+
+bool Utils::uris_user_match(const std::string& matcher,
+                            const std::string& matchee)
```
graemerobertson: WTF is a matchee?!

In all seriousness, I can't think of a good names for this function and parameters, but I think I would have gone for something like

`bool Utils::check_users_equivalent(const std::string& specific_user, const std::string& canonical_user)`

Either way, I think you should comment on the significance of matcher and matchee because I'm struggling to get my head around it and matchee isn't a real word.

src/utils.cpp
```diff
@@ -711,3 +712,105 @@ Utils::IPAddressType Utils::parse_ip_address(std::string address)
...
+    return false;
+  }
+
+  // Finally, check against the regex.
+  std::string matcher_part = matcher_user[0].substr(matcher_start + 1,
+                                                     (matcher_end - matcher_start - 1));
+  std::string matchee_part = matchee_end_str.substr(0, (matchee_end_str.size() -
+                                                    matcher_end_str.size()));
+  boost::regex matcher_regex = boost::regex(matcher_part,
+                                             boost::regex_constants::no_except);
```
graemerobertson: LAIGNMENT

PR #629, opened 2017-03-10
[Reviewer:HJS] Update sas_event.h following SAS changes

jack-atack:
Update resource bundle identifier, and add new Event Space

### DIFF
--- a/include/sasevent.h
+++ b/include/sasevent.h
@@ -41,7 +41,7 @@
 #include "sas.h"
 
 namespace SASEvent {
-  const std::string CURRENT_RESOURCE_BUNDLE_DATESTAMP = "20170203";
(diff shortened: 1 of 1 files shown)


include/sasevent.h
```diff
@@ -89,6 +89,7 @@ namespace SASEvent {
   const int HOUDINI_BASE = 0x890000;
   const int BIFROST_BASE = 0x8A0000;
   const int WEATHERWAX_BASE = 0x8B0000;
+  const int RPE_BASE = 0x8C0000;  // This is also defined in the python RPE source code
```
huws: I don't think the comment is necessary.  Actually, I don't think the line is necessary - we never use it from the C++ code, and its not like someone could define overlapping events - they'd spot the clash in the SAS resource bundle.

PR #633, opened 2017-03-21
[Reviewer: Adam] Commonise the XML functions

eleanor-merry:
This pulls out the generic XML functions from Sprout/Homestead - the code hasn't changed.


rkd-msw:
I'd be useful to get this merged early today if possible - I'm trying to do some stress testing of wildcards and I'm hitting a crash at https://github.com/Metaswitch/homestead/blob/dev/src/xmlutils.cpp#L234, which I think this fixes.

PR #634, opened 2017-03-22
[Reviewer: HJS] SAS datestamp update 20170322

GAB-MS:
Update SAS datestamp for resource bundle update.

Master only.

### DIFF
--- a/include/sasevent.h
+++ b/include/sasevent.h
@@ -44,7 +44,7 @@ namespace SASEvent {
   // On updating CURRENT_RESOURCE_BUNDLE, you must also update
   // SAS_RESOURCE_BUNDLE in the /python_src/metaswitch/rpe/rpesasconstants.py
   // file in the houdini repository.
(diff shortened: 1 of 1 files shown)


include/sasevent.h
```diff
@@ -44,7 +44,7 @@ namespace SASEvent {
   // On updating CURRENT_RESOURCE_BUNDLE, you must also update
   // SAS_RESOURCE_BUNDLE in the /python_src/metaswitch/rpe/rpesasconstants.py
   // file in the houdini repository.
-  const std::string CURRENT_RESOURCE_BUNDLE_DATESTAMP = "20170310";
+  const std::string CURRENT_RESOURCE_BUNDLE_DATESTAMP = "20170322";
```
huws: The comment above this line says that you need to update a value in the RPE, too.

GAB-MS: That's true - good spot. I'll add that update to the other PR.

PR #639, opened 2017-04-24
[Reviewer: KH1] Constants for call barring

sathiyan-sivathas:
Add some constants for the call barring field.


include/xml_utils.h
```diff
@@ -133,6 +134,7 @@ namespace RegDataXMLUtils
   const char* const STATE_NOT_REGISTERED = "NOT_REGISTERED";
   const char* const STATE_REGISTERED = "REGISTERED";
   const char* const STATE_UNREGISTERED = "UNREGISTERED";
+  const char* const STATE_BARRED = "1";
```
eleanor-merry: Worth having state_unbarred as well?

sathiyan-sivathas: There will be once Krista merges her PR.

PR #640, opened 2017-04-26
Add new barring constant

KristaHyer:
To neaten up the logic when handling RTRs, I've added this new constant.
(It allows me to set this as default.)


include/xml_utils.h
```diff
@@ -135,6 +135,7 @@ namespace RegDataXMLUtils
   const char* const STATE_REGISTERED = "REGISTERED";
   const char* const STATE_UNREGISTERED = "UNREGISTERED";
   const char* const STATE_BARRED = "1";
+  const char* const STATE_UNBARRED = "0";
```
sathiyan-sivathas: If I was being really picky I'd say 0 should come before 1

KristaHyer: Now that you've pointed it out I see it too.... Fixing it now

PR #643, opened 2017-05-10
Enhance the local store to allow failures when getting data

AlexHockey:
Some enhancements to the local store class we us in UTs to allow it to fail get requests. Needed for https://github.com/Metaswitch/sprout/pull/1823

### DIFF
--- a/src/localstore.cpp
+++ b/src/localstore.cpp
@@ -51,7 +51,8 @@ LocalStore::LocalStore() :
   _data_contention_flag(false),
   _db_lock(PTHREAD_MUTEX_INITIALIZER),
   _db(),
-  _force_error_flag(false),
+  _force_error_on_set_flag(false),
+  _force_error_on_get_flag(false),
   _old_db()
 {
   TRC_DEBUG("Created local store");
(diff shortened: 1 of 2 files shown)


src/localstore.cpp
```diff
@@ -161,10 +178,10 @@ Store::Status LocalStore::set_data(const std::string& table,
 
   // This is for the purpose of testing data SETs failing.  If the flag is set
   // to true, then we'll just return an error.
-  if (_force_error_flag == true)
+  if (_force_error_on_set_flag == true)
```
eleanor-merry: Not just if (_force_error_on_set_flag)?

src/localstore.cpp
```diff
@@ -98,6 +106,15 @@ Store::Status LocalStore::get_data(const std::string& table,
   TRC_DEBUG("get_data table=%s key=%s", table.c_str(), key.c_str());
   Store::Status status = Store::Status::NOT_FOUND;
 
+  // This is for the purpose of testing data GETs failing.  If the flag is set
+  // to true, then we'll just return an error.
+  if (_force_error_on_get_flag == true)
```
eleanor-merry: Not just if (_force_error_on_get_flag)?

PR #648, opened 2017-05-24
Rough pass to rework multi state alarms

eleanor-merry:
Alex, you've agreed to quickly look over these changes before I change the rest of the code to use it,

The changes:
* Merge the Alarm and MultiStateAlarm classes so that there's one Alarm class. Code that uses alarms must now call set_major(), set_minor(), etc.. rather than calling set().
* Adds an adaptor class (in this case just for the communication monitor) to translate for code that doesn't know what the right severity to raise is in a particular situation.

I've not added comments/thought about naming yet.

### DIFF
--- a/include/alarm.h
+++ b/include/alarm.h
@@ -122,40 +122,35 @@ private:
   std::string _identifier;
 };
 
-/// @class BaseAlarm
-/// 
-/// Superclass for alarms that allows us to split different types of
-/// alarms into subclasses. Those alarms which only have one possible raised
-/// state will be constructed by subclass Alarm. Those alarms which have two or
-/// more possible raised states will be constructed by subclass
-/// MultiStateAlarm.
-
-class BaseAlarm
+class Alarm
 {
 public:
+  Alarm(AlarmManager* alarm_manager,
+        const std::string& issuer,
+        const int index);
+
+  virtual ~Alarm();
+
+  /// Returns the index of this alarm.
+  int index() const {return _index;}
+
+  /// Queues a request to generate an alarm state change corresponding to the
+  /// requested severity
+  virtual void set(AlarmDef::Severity severity);
+
   /// Queues a request to generate an alarm state change corresponding to the
   /// CLEARED severity.
   virtual void clear();
 
-  /// Uses the _last_state_raised member variable to re-raise the latest state
-  /// of the alarm.
-  void reraise_last_state();
-  
-  /// Returns the current state of the alarm as one of UNKNOWN, CLEARED, or ALARMED.
-  virtual AlarmState::AlarmCondition get_alarm_state();
+  /// Helper functions that set an alarm to a known state.
+  void set_indeterminate();
+  void set_warning();
+  void set_minor();
+  void set_major();
+  void set_critical();
 
-  // If an alarm is currently in a different state to the one we wish to raise
-  // the alarm in, we raise the alarm and update _last_state_raised.
-  void switch_to_state(AlarmState* new_state);
-
-protected:
-  BaseAlarm(AlarmManager* alarm_manager,
-            const std::string& issuer,
-            const int index);
-  ~BaseAlarm();
-
+private:
   const int _index;
-  AlarmState _clear_state;
 
   // When an alarm is issued we change the _last_state_raised member variable of
   // the alarm. We must not allow another thread to raise the same alarm at a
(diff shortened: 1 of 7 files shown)


include/alarm.h
```diff
@@ -246,64 +247,50 @@ class AlarmManager
 
 /// @class Alarm
 ///
-/// Encapsulates an alarm's only active state with its associated alarm clear state.
 /// Used to manage the reporting of a fault condition, and subsequent clear
 /// of said condition.
-
-class Alarm: public BaseAlarm
+class Alarm : public BaseAlarm
```
AlexHockey: Do we still need `BaseAlarm`?

include/alarm.h
```diff
@@ -246,64 +247,50 @@ class AlarmManager
...
 ///
-/// Encapsulates an alarm's two or more active states with its associated clear
-/// state. Used to manage the reporting of a fault condition, and subsequent
-/// clear of said condition. We further subclass this on a per-alarm basis, to
-/// give an alarm visibility of the protected raising functions cordesponding to that
-/// alarm's possible raised states. This would stop a user raising alarm at a
-/// state which does not exist for that alarm.
-
-class MultiStateAlarm: public BaseAlarm
+class AlarmAdaptor
```
AlexHockey: Is there much point in this class? If the only purpose is to own the alarm and free it on destruction, maybe a smart pointer would be better?

include/communicationmonitor.h
```diff
@@ -15,6 +15,39 @@
...
+    AlarmAdaptor(alarm),
+    _on_some_errors(on_some_errors),
+    _on_all_errors(on_all_errors)
+  {}
+
+  virtual ~CMAlarmAdaptor() {}
+
+  void only_errors()
+  {
+    _alarm->set(_on_all_errors);
```
AlexHockey: `set` is declared as a private method. It looks like you're intending to make it public (I don't disagree with that position, I just want to check that's what you're intending).

eleanor-merry:
Ok, this is ready for full review now please. There's two PRs that use this new code at https://github.com/Metaswitch/chronos/pull/352 and https://github.com/Metaswitch/ralf/pull/275 (but I won't send you the rest).

test_utils/mockalarm.h
```diff
@@ -19,24 +19,11 @@ class MockAlarm : public Alarm
...
-{
-public:
-  MockMultiStateAlarm(AlarmManager* alarm_manager) :
-    MultiStateAlarm(alarm_manager, "sprout", 0) {}
-
-  MOCK_METHOD0(set_indeterminate, void());
-  MOCK_METHOD0(set_warning, void());
-  MOCK_METHOD0(set_minor, void());
-  MOCK_METHOD0(set_major, void());
-  MOCK_METHOD0(set_critical, void());
```
AlexHockey: Worth keeping these given they're present on the interface?

src/communicationmonitor.cpp
```diff
@@ -128,17 +128,17 @@ void CommunicationMonitor::track_communication_changes(unsigned long now_ms)
...
             case SOME_ERRORS:
               CL_CM_CONNECTION_PARTIAL_ERROR.log(_sender.c_str(),
                                                  _receiver.c_str());
-              _alarm->clear();
+              _alarm_adaptor->some_comm_errors();
               break;
 
             case ONLY_ERRORS: // No change in state. Ensure alarm is raised.
-              _alarm->set();
+              _alarm_adaptor->only_comm_errors();
```
AlexHockey: Not your code, but these switch statements confused me for a bit. Would it be clearer to just have one switch, and only generate the ENT log if the old and new states are different.

src/alarm.cpp
```diff
@@ -101,72 +115,69 @@ void BaseAlarm::reraise_last_state()
...
+    case AlarmDef::CRITICAL:
+      set_critical();
+      break;
+
+    // LCOV_EXCL_START
+    case AlarmDef::UNDEFINED_SEVERITY:
+    default:
+      // We should never hit this case
+      break;
+    // LCOV_EXCL_STOP
```
AlexHockey: I can't put my finger on why but this feels backwards to me (though TBH so does the fact that we have separate fields for each alarm state). I don't care too much but I thought I'd mention it in case you were thinking the same thing!

eleanor-merry:
Closing - we think this needs better auto-gen of the alarms.

PR #649, opened 2017-05-26
[Reviewer: Adam] Cassandra fixes

sebrexmetaswitch:
This is a fix to alleviate issue: https://github.com/Metaswitch/clearwater-issues/issues/95

For our cassandra read requests, we try Consistency Level TWO first, and then Consistency Level ONE if that fails.

Previously, we would do this for each Cassandra node that we tried to contact. This meant that, in the scenario where we have 2 Cassandra nodes, A and B, and A is being too slow to respond to our request, we would have done:

 - TWO request to A, which times out
 - ONE request to A, which times out
 - TWO request to B, which times out
 - ONE request to B, which succeeds

Although this eventually succeeds, we've taken a long time to establish that the TWO requests didn't work. In the case of Homestead talking to Cassandra, Sprout would have already timed out the request.

Therefore, if we fail a TWO read, we want all subsequent attempts at the read to be a ONE read.

We do this by

* creating a new `HAOperation` class, which handles performing the HA gets:
* the `HA` macro is moved to be used by this class to define its `ha_*` methods
* the class records whether it has already tried an HA get, and if so it skips straight to the ONE read

since the same Operation is retried with both cassandra Clients, this means that the third read in the example above would be a ONE read

This changes the interface for the ha_ gets, so requires corresponding changes to other repos (homestead and memento)

Also, remove a trailing coverage exclusion that's not necessary

## Testing
* UTs and live tests pass.
* I ran stress through a 2-2-2 deployment in EC2, as I was trying to repro the issue on Vellum before this change. I wasn't able to repro it, and hit other problems first, but I did run the same level of stress through a system with these changes and check that all calls succeeded

### DIFF
--- a/src/cassandra_store.cpp
+++ b/src/cassandra_store.cpp
@@ -391,7 +391,7 @@ bool Store::perform_op(Operation* op,
     cass_result = OK;
     attempt_count++;
 
-    // Only retry on connection errors
+    // Both Cassandra timeouts and connection errors should result in a retry
     retry = false;
 
     // Get a client to execute the operation.
@@ -430,6 +430,21 @@ bool Store::perform_op(Operation* op,
       TRC_DEBUG("Connection error");
       retry = true;
     }
+    catch(TimedOutException& te)
+    {
+      // A timeout should trigger a retry, but should not blacklist the node we
+      // connected to (as we succeeded in connecting to it)
+      cass_result = TIMEOUT;
+      cass_error_text = (boost::format("Exception: %s")
+                         % te.what()).str();
+
+      // SAS log the timeout.
+      SAS::Event event(trail, SASEvent::CASS_TIMEOUT, 0);
+      SAS::report_event(event);
+
+      TRC_DEBUG("Timeout");
+      retry = true;
+    }
     catch(InvalidRequestException& ire)
     {
       cass_result = INVALID_REQUEST;
@@ -460,9 +475,8 @@ bool Store::perform_op(Operation* op,
       cass_error_text = "Unknown error";
     }
 
-    // Since we only retry on connection errors, the value of retry is used to
-    // determine whether we connected to the target successfully
-    if (retry)
+    // Only blacklist the target if there was an error connecting to it
+    if (cass_result == CONNECTION_ERROR)
     {
       _resolver->blacklist(target);
     }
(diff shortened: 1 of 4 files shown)


sebrexmetaswitch:
Note - I've not yet done the memento changes, I'll do them before merging this (but the homestead ones are done)

johadalin:
Code changes look all good. 
Did you do any manual testing to ensure that the retry behaviour works as expected? I.e. turning one cassandra down, and trying to force a request through it? 
If not, we should discuss with the test team to see if they are able to verify it's all working as expected.

sebrexmetaswitch:
Yes, I thought I'd written that up too, but looks like I missed it off.

I tested that, with one Cassandra down, the first request (to the down Cassandra) was a level TWO read, and then second request went to the live Cassandra and was a level ONE read.

sebrexmetaswitch:
Adam - that last commit adds the behaviour we spoke about earlier, namely that we should retry a Cassandra operation that hits a Cassandra timeout error. I've also made corresponding changes in homestead and the sas resource bundle (as there's a new SAS log added for this error case).

(We want to do this because we want to ensure that a consistency ONE request that times out gets retried, as that's what we would have done before these changes were made)

src/cassandra_store.cpp
```diff
@@ -430,6 +430,21 @@ bool Store::perform_op(Operation* op,
       TRC_DEBUG("Connection error");
       retry = true;
     }
+    catch(TimedOutException& te)
+    {
+      // A timeout should trigger a retry, but should not blacklist the node we
+      // connected to (as we succeeded in connecting to it)
```
johadalin: If someone is not familiar with the split in cassandra between the routing and storage connections, this comment could be confusing. Can you clarify that it signifies we connected to the routing node, but then hit a timeout?

src/cassandra_store.cpp
```diff
@@ -430,6 +430,21 @@ bool Store::perform_op(Operation* op,
...
+      cass_result = TIMEOUT;
+      cass_error_text = (boost::format("Exception: %s")
+                         % te.what()).str();
+
+      // SAS log the timeout.
+      SAS::Event event(trail, SASEvent::CASS_TIMEOUT, 0);
+      SAS::report_event(event);
+
+      TRC_DEBUG("Timeout");
```
johadalin: Similarly here, it's not clear without context what the difference is between the 'Connection error' log above, and 'timeout' here. Think they should be improved to be a little clearer on what the exception that triggered them actually is, and where the timeout occurred.

PR #650, opened 2017-06-01
Extend mockhttpconnection to include send_get method.

james-jra:
Extend mockhttpconnection.h to include a mocked-out send_get method.

Also change argument name in existing mocked-out send_post method to reflect the interface as defined in include/httpconnection.h


johadalin:
Also, not sure who i should be passing this back to, between @dimbleby and @james-jra . Saw in the other PR that you guys are working together on lots of this. Passed back to Dave as he assigned this to me, but who knows

dimbleby:
thanks - this is from testing the robocall-blocking sproutlet

PR #654, opened 2017-06-19
[Reviewer: Alex] Make CM logs more specific

eleanor-merry:
The communication monitor had a very generic set of ENT logs that weren't very specific. This PR makes users of the monitor have to provide the ENT logs, so they can now be specific.

Tested in UT

### DIFF
--- a/include/communicationmonitor.h
+++ b/include/communicationmonitor.h
@@ -14,6 +14,24 @@
 
 #include "alarm.h"
 #include "base_communication_monitor.h"
+#include "pdlog.h"
+
+/// Structure that holds the different ENT logs that get raised by the
+// Communication Monitor
+struct CMPDLogs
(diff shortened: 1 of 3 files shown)


include/communicationmonitor.h
```diff
@@ -44,12 +63,14 @@ class CommunicationMonitor : public BaseCommunicationMonitor
   unsigned long current_time_ms();
 
   Alarm* _alarm;
+  CMPDLogs _cm_pd_logs;
   std::string _sender;
```
AlexHockey: Do we still need `_sender` and `_receiver`? They only thing there are used for now is producing a debug log so it doesn't seem worthwhile to store them.

AlexHockey:
Looks good. One minor comment.

PR #656, opened 2017-07-17
[Reviewer: AJH] Add 20s HTTP request read timeout

GAB-MS:
@AlexHockey please let me know if you're not the right person to take this.

This addresses a potential vulnerability in libevhtp. After initialising the stack, we set a 20s read timeout. Closure of the vulnerability has been tested by running nmap: after applying this fix, the issue is not flagged.

### DIFF
--- a/src/httpstack.cpp
+++ b/src/httpstack.cpp
@@ -123,10 +123,8 @@ void HttpStack::initialize()
     _evhtp = evhtp_new(_evbase, NULL);
 
     // Set a read timeout of 20s to mitigate the Slowloris vulnerability.
-    struct timeval* recv_timeo = new timeval();
-    recv_timeo->tv_sec = 20;
-    evhtp_set_timeouts(_evhtp, recv_timeo, NULL);
-    delete recv_timeo;
+    struct timeval recv_timeo = { .tv_sec = 20, .tv_usec = 0 };
+    evhtp_set_timeouts(_evhtp, &timeval, NULL);
   }
 }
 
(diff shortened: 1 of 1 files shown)


GAB-MS:
20 seconds was a value I chose after finding a few relevant sites on mitigating slowloris. One test suggested that against a similarly-vulnerable Apache web server, setting a read timeout below 30s on the HTTP request body prevented a single attack agent from blocking the server from fulfilling requests.

This change sets the timeout on the buffer event to 20s, which was enough to stop nmap complaining about the vulnerability, but shouldn't cause any issues for legitimate requests unless we're hitting major network problems.

I'll expand the comment to explain as such, and send this back to you to make sure you're happy with this explanation.

AlexHockey:
When I saw the PR initially I thought "why not set this to something much smaller like 1s?". It would be great if your comment talked about that as well :-)

GAB-MS:
I'm concerned about cross-site latency. I believe that at least Chronos needs to be able to communicate across GR sites, and uses this stack. I'd be rather uneasy about 1s of latency as a result, in case there are issues leading to increased latency between sites.

There's scope for it to be reduced further but, given that 20s is enough to close the vulnerability as far as nmap is concerned, I'm unconvinced that it's worth doing so. But I'm also not sure how tolerant we are of latency in general; as long as we're above that threshold, we'd be okay. Thoughts?

AlexHockey:
That's fine. I'm not saying that you're choice is wrong. I'm just asking you to put your justification in a comment so we don;t lose your reasoning :-).

PR #657, opened 2017-07-20
Replace a duff engineering log with a SAS log

AlexHockey:
The log at the end of `HttpClient::send_request` is all kinds of wrong. 

* It contains little useful information about what was going on that might have caused the problem.
* It can be generated when there wasn't a cURL error.
* It can be generated when cURL wasn't even invoked (e.g. when DNS failed to resolve).
* It shouldn't even be a log - it should be a SAS event instead. 

This PR fixes these issues by removing the log and replacing it with a SAS log covering the scenario when the hostname does not resolve.


AlexHockey:
cpp-common-test runs cleanly. I tested live by misconfiguring the homestead hostname and checking the trace in SAS.

PR #658, opened 2017-07-24
Add support for string and time indexed event stats

MatMeredith:
Add support for string and time indexed event stats.   We're going to need these in order to expose per AS latency stats in the S-CSCF.

Tested by https://github.com/Metaswitch/clearwater-fv-test/pull/56

### DIFF
--- a/src/event_statistic_accumulator.cpp
+++ b/src/event_statistic_accumulator.cpp
@@ -54,7 +54,7 @@ void EventStatisticAccumulator::accumulate(uint32_t sample)
   }
 }
 
-inline void EventStatisticAccumulator::get_stats(EventStatistics &stats)
+void EventStatisticAccumulator::get_stats(EventStatistics &stats)
 {
   stats.count = _count;
 
--- a/src/event_statistic_accumulator.cpp
+++ b/src/event_statistic_accumulator.cpp
@@ -1,0 +1,91 @@
+/**
+ * @file event_statistic_accumulator.cpp
+ *
+ * Copyright (C) Metaswitch Networks 2017
+ * If license terms are provided to you in a COPYING file in the root directory
+ * of the source code repository by which you are accessing this code, then
+ * the license outlined in that COPYING file applies to your use.
+ * Otherwise no rights are granted except for those provided to you by
+ * Metaswitch Networks in a separate written agreement.
+ */
+
+#include <vector>
+#include <map>
+#include <string>
+#include <atomic>
+
+#include "logger.h"
+#include "event_statistic_accumulator.h"
+#include "limits.h"
+
+namespace SNMP
+{
+
+EventStatisticAccumulator::EventStatisticAccumulator()
+{
+  reset(0);
+}
+
+void EventStatisticAccumulator::accumulate(uint32_t sample)
+{
+  _count++;
+
+  // Just keep a running total as we go along, so we can calculate the average and variance on
+  // request
+  _sum += sample;
+  _sqsum += (sample * sample);
+
+  // Update the low- and high-water marks.  In each case, we get the current
+  // value, decide whether a change is required and then atomically swap it
+  // if so, repeating if it was changed in the meantime.  Note that
+  // compare_exchange_weak loads the current value into the expected value
+  // parameter (lwm or hwm below) if the compare fails.
+  uint_fast64_t lwm = _lwm.load();
+  while ((sample < lwm) &&
+         (!_lwm.compare_exchange_weak(lwm, sample)))
+  {
+    // Do nothing.
+  }
+  uint_fast64_t hwm = _hwm.load();
+  while ((sample > hwm) &&
+         (!_hwm.compare_exchange_weak(hwm, sample)))
+  {
+    // Do nothing.
+  }
+}
+
+inline void EventStatisticAccumulator::get_stats(EventStatistics &stats)
+{
+  stats.count = _count;
+
+  if (_count > 0)
+  {
+    // Get the values that we are working with atomically.
+    uint_fast64_t sum = _sum;
+    uint_fast64_t sumsq = _sqsum;
+
+    // Calculate the average and the variance from the stored sum and sum-of-squares.
+    stats.mean = sum/stats.count;
+    stats.variance = ((sumsq * stats.count) - (sum * sum)) / (stats.count * stats.count);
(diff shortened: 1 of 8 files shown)


sebrexmetaswitch:
One thought on this (Adam's reviewing it, but I'm using it for the Sprout AS stats implementation):

Currently, we index the stats by period, and then by string. This means we see stats like:

```
 scopePrevious5SecondPeriod."appserver.domain"                           0                            0                       0                       0                         0
scopePrevious5SecondPeriod."appserver2.domain"                           0                            0                       0                       0                         0
  scopeCurrent5MinutePeriod."appserver.domain"                           0                            0                       0                       0                         0
 scopeCurrent5MinutePeriod."appserver2.domain"                           0                            0                       0                       0                         0
 scopePrevious5MinutePeriod."appserver.domain"                         150                         2500                     200                     100                         2
scopePrevious5MinutePeriod."appserver2.domain"                         300                        10000                     400                     200                         2

```

I'm not sure how easy it would be to do, and if it's hard then it may not be worth it, but would it be nicer to index them by string and then by time period?

(I also don't know if it makes much difference, but I can imagine it would be easier to use the stats if they were indexed by App Server first, then by time period)

src/snmp_time_and_string_based_event_table.cpp
```diff
@@ -0,0 +1,181 @@
...
+  {
+    pthread_rwlock_destroy(&_table_lock);
+
+    //Spin through the maps of 5s and 5m views and delete them.
+    for (auto& kv : _five_second) {  delete kv.second; }
+    for (auto& kv : _five_minute) {  delete kv.second; }
+  }
+private:
+
+  TimeAndStringBasedEventRow* new_row(int indexes) { return NULL;};
```
johadalin: `indicies`, surely :P

MatMeredith:
@sebrexmetaswitch It's a good question.   
- It would be a fair amount of effort to change it I think -- certainly not something we can do this sprint.   On the other hand, once we've shipped an updated MIB, we can't change it around retrospectively.   
- I don't know which way round is better.   My initial instinct is that the other way round would be better, but...  the way it is does mean if what you care about generally is e.g. the previous 5s stats you can say "show me the stats for all app servers for the previous 5s".   And maybe that's actually more useful?   How often are you going to want an overview of all the servers vs wanting to see stats for one specific one?   And if you do want to see stats for a specific one you can of course query it -- you just have to do so separately for each time period.
- Finally (and most importantly!) what we've currently implemented is (I hope!) consistent with all other stats in Clearwater that are indexed by timescope and something else.  See e.g. https://github.com/Metaswitch/clearwater-snmp-handlers/blob/master/mib-generator/CLEARWATER-MIB-COMMON#L1737-L1738 

OK?

src/event_statistic_accumulator.cpp
```diff
@@ -0,0 +1,97 @@
...
+  // structure.
+  stats.count = _count;
+
+  if (_count > 0)
+  {
+    // Get local copies of the values that we are working with atomically.
+    // Note that e.g. _sum is a std::atomic and so assignment is guaranteed
+    // atomic.
+    uint_fast64_t sum = _sum;
+    uint_fast64_t sumsq = _sqsum;
```
MatMeredith: Reason?  Copied from elsewhere!  Good point though ;-)

src/snmp_time_and_string_based_event_table.cpp
MatMeredith: https://english.stackexchange.com/questions/61080/plural-of-index-indexes-or-indices/3126
```
mgm@ip-10-0-0-85:~/clearwater-fv-test/modules/cpp-common$ grep -r indexes *
include/snmp_internal/snmp_table.h:    snmp_free_varbind(_table->indexes_template);
include/snmp_internal/snmp_table.h:    snmp_free_varbind(_table_info->indexes);
include/snmp_internal/snmp_counts_by_other_type_table.h:  RowType* new_row(int indexes) { return NULL; };
include/memcachedstoreview.h:  /// Converts a vector of server indexes into a vector of server names.
include/memcachedstoreview.h:  /// @param ids          - A vector of replica indexes.
src/snmp_continuous_accumulator_by_scope_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_time_and_string_based_event_table.cpp:  TimeAndStringBasedEventRow* new_row(int indexes) { return NULL;};
src/snmp_scalar_by_scope_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_event_accumulator_by_scope_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_continuous_increment_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_counter_by_scope_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_event_accumulator_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_counter_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_success_fail_count_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/memcachedstoreview.cpp:      std::vector<int> server_indexes = ring.get_nodes(ii, replicas);
src/memcachedstoreview.cpp:      for (size_t jj = 0; jj < server_indexes.size(); jj++)
src/memcachedstoreview.cpp:        int idx = server_indexes[jj];
src/snmp_ip_row.cpp:  // Set the IPAddrType and IPAddr as indexes
src/snmp_ip_row.cpp:  // Set the IPAddrType and IPAddr as indexes
src/snmp_continuous_accumulator_table.cpp:  // Map row indexes to the view of the underlying data they should expose
src/snmp_cx_counter_table.cpp:  CxCounterRow* new_row(int indexes) { return NULL;};
mgm@ip-10-0-0-85:~/clearwater-fv-test/modules/cpp-common$ grep -r indicies *
mgm@ip-10-0-0-85:~/clearwater-fv-test/modules/cpp-common$
```

MatMeredith: I will however fix up my incorrect "indices"!

MatMeredith:
@johadalin Thanks.   Re. testing.   As I understand it stats code is deliberately tested in clearwater-fv-test rather than in cpp-common UT because of its dependency on snmp.   I don't plan to do any differently here...

PR #659, opened 2017-07-25
Fix bugs in SAS logging when HTTP DNS name does not resolve to anything

AlexHockey:
Two changes:

* We increment attempts whenever we try to contact an  HTTP server. 
* Unescape the URL before logging to SAS. 

Tested as follows:

* REGISTER with the homestead hostname configured correctly. 

image

* REGISTER with the hostname mis-configured (so it does not resolve). 

image

* REGISTER with homestead not running (hopefully proves the change from for loop -> while loop hasn't broken the world). 

image

Unit test code incoming.

### DIFF
--- a/src/httpclient.cpp
+++ b/src/httpclient.cpp
@@ -390,11 +390,11 @@ HTTPCode HttpClient::send_request(RequestType request_type,
   //
   // Note that we need to accurately track how many attempts we have actually
   // made, even if we break out of the loop early (so we generate accurate
-  // logs). For this reason we increment the counter as soon as we  start a loop
-  // and assume that if we will try a host on each iteration. This is not
-  // perfect, but it's better than incrementing the counter mid-way through the
-  // loop (when actually trying the host) and risking not incrementing the
-  // counter for some reason, which would give an infinite loop.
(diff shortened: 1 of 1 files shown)


src/httpclient.cpp
```diff
@@ -387,11 +387,19 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
   // targets are exhausted. If only one target is available, it should be tried
   // twice.
-  int attempts;
-  for (attempts = 0;
-       target_it->next(target) || attempts == 1;
-       ++attempts)
+  //
+  // Note that we need to accurately track how many attempts we have actually
+  // made, even if we break out of the loop early (so we generate accurate
+  // logs). For this reason we increment the counter as soon as we  start a loop
```
eleanor-merry: extra space between we and start

src/httpclient.cpp
```diff
@@ -387,11 +387,19 @@ HTTPCode HttpClient::send_request(RequestType request_type,
...
+  // and assume that if we will try a host on each iteration. This is not
```
eleanor-merry: I don't think the if should be there.

PR #661, opened 2017-08-03
[Reviewer: HJS] 'Allowed host state' interface change and implementation for DNS resolvers

GAB-MS:
Goes along with Metaswitch/sprout#1891. I will be opening the cpp-common-test PR shortly, which includes the unit tests for this change.

Interesting things to note:
* The constants defining the allowed host states had to be defined as `static`. Furthermore, I chose to define them as `constexpr` instead of `const` - this implies both `inline` and `const` for the static variable, so they are evaluated at compile time. I saw no downside in doing this, but please shout if you disagree.
* As discussed in the Sprout PR - I felt that it was appropriate to add a helper function (`get_allowed_host_states`) which takes the `allowed_host_state` bit and sets two bool references to indicate which host states were allowed, instead of performing a bitwise check every single time we want to evaluate this in the various resolve functions. This was also a convenient place to add trace for these bitwise checks.
* In both `srv_resolve` and `LazyAddrIterator::take`, I declare both bools in one line before passing to the above-mentioned function. Shout if you disagree and think they should be on separate lines.
* As with the `SIPResolver` SAS change, I'm adding the allowed host states as extra static parameters on the SAS events here, where the events should change their text based on these parameters. However, for these events specifically, if we wanted to avoid a change to their interface and their definition, we could instead just modify the strings we pass in to something appropriate - thoughts?
* For A-record resolution, I was unsure whether we should always be returning a graylisted record, or whether we should only do so if set to allow whitelisted records. I implemented the latter, but shout if this is wrong.
* The weird if/else statement structure starting at line 1018 in `baseresolver.cpp` is necessary to be able to log cases where we're ignoring a server. It's also necessary to split the if/else for the host state check away from checking whether whitelisting is allowed, otherwise we'd move on to the second branch in the case where the host is whitelisted and whitelisted results aren't allowed - we would have to check the host state again to ensure that it's not white, alongside checking whether blacklisted results are allowed.

### DIFF
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -64,6 +64,11 @@ public:
   // it is desirable not to expose
   friend class LazyAddrIterator;
 
+  // Constants indicating the allowed host state values.
+  const int WHITELISTED = 0x01;
+  const int BLACKLISTED = 0x02;
+  const int ALL_LISTS   = WHITELISTED | BLACKLISTED;
+
 protected:
   void create_naptr_cache(std::map<std::string, int> naptr_services);
   void create_srv_cache();
@@ -85,7 +90,8 @@ protected:
                    int retries,
                    std::vector<AddrInfo>& targets,
                    int& ttl,
-                   SAS::TrailId trail);
+                   SAS::TrailId trail,
+                   int allowed_host_state=ALL_LISTS);
 
   /// Does an A/AAAA record resolution for the specified name, selecting
   /// appropriate targets.
@@ -96,7 +102,8 @@ protected:
                  int retries,
                  std::vector<AddrInfo>& targets,
                  int& ttl,
-                 SAS::TrailId trail);
+                 SAS::TrailId trail,
+                 int allowed_host_state=ALL_LISTS);
 
   /// Does an A/AAAA record resolution for the specified name, and returns an
   /// Iterator that lazily selects appropriate targets.
@@ -105,7 +112,8 @@ protected:
                                            int port,
                                            int transport,
                                            int& ttl,
-                                           SAS::TrailId trail);
+                                           SAS::TrailId trail,
+                                           int allowed_host_state=ALL_LISTS);
 
   /// Converts a DNS A or AAAA record to an IP46Address structure.
   IP46Address to_ip46(const DnsRRecord* rr);
@@ -329,7 +337,8 @@ public:
                    BaseResolver* resolver,
                    int port,
                    int transport,
-                   SAS::TrailId trail);
+                   SAS::TrailId trail,
+                   int allowed_host_state);
   virtual ~LazyAddrIterator() {}
 
   /// Returns a vector containing at most num_requested_targets AddrInfo targets,
@@ -348,6 +357,9 @@ private:
 
   // A pointer to the BaseResolver that created this iterator
   BaseResolver* _resolver;
+
+  // The allowed state of hosts returned by this iterator
+  int _allowed_host_state;
 
   std::string _hostname;
   SAS::TrailId _trail;
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -113,7 +113,7 @@ protected:
                                            int transport,
                                            int& ttl,
                                            SAS::TrailId trail,
-                                           int allowed_host_state=ALL_LISTS);
+                                           int allowed_host_state);
 
   /// Converts a DNS A or AAAA record to an IP46Address structure.
   IP46Address to_ip46(const DnsRRecord* rr);
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -65,9 +65,9 @@ public:
   friend class LazyAddrIterator;
 
   // Constants indicating the allowed host state values.
-  static constexpr int WHITELISTED = 0x01;
-  static constexpr int BLACKLISTED = 0x02;
-  static constexpr int ALL_LISTS   = WHITELISTED | BLACKLISTED;
+  static const int WHITELISTED = 0x01;
+  static const int BLACKLISTED = 0x02;
+  static const int ALL_LISTS   = WHITELISTED | BLACKLISTED;
 
 protected:
   void create_naptr_cache(std::map<std::string, int> naptr_services);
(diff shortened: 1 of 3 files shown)


include/baseresolver.h
```diff
@@ -64,6 +64,11 @@ class BaseResolver
   // it is desirable not to expose
   friend class LazyAddrIterator;
 
+  // Constants indicating the allowed host state values.
+  static constexpr int WHITELISTED = 0x01;
+  static constexpr int BLACKLISTED = 0x02;
+  static constexpr int ALL_LISTS   = WHITELISTED | BLACKLISTED;
```
huws: I do not believe that `constexpr` is having any effect here (beyond implicitly making the variables `const`).  We're already initialising these constants from literal values, so the compiler knows that these are constant expressions.  That said, I don't think this does any harm (other than making the reader look-up what `constexpr` means...).

include/baseresolver.h
```diff
@@ -77,6 +82,13 @@ class BaseResolver
   void destroy_srv_cache();
   void destroy_blacklist();
 
+  // Helper function that takes the bit representing the allowed host states,
+  // and sets two bool references passed into the function to indicate which
+  // host states are permitted.
+  void get_allowed_host_states(const int allowed_host_state,
```
huws: FYI - this `const` does nothing.  I'm assuming that you've put it there just to match the one in the method definition (which does do something).

src/baseresolver.cpp
```diff
@@ -100,6 +100,33 @@ void BaseResolver::destroy_blacklist()
...
+  {
+    TRC_DEBUG("Blacklisted addresses allowed");
+    blacklisted_allowed = true;
+  }
+  else
+  {
+    TRC_DEBUG("Blacklisted addresses not allowed and will be ignored");
+    blacklisted_allowed = false;
+  }
+}
```
huws: As I mentioned in the sprout review, I think this method is entirely unnecessary.  It is altogether too verbose.  For example:
 - The bit-wise operators return integer values that can be implicitly cast to Booleans.  Checking each flag should be one line.
 - We don't need per-branch logging of such a simple operation as checking a bitwise flag.  We might want to log the value of the bit-field when it is received by a top-level method, but we don't need to log details of how we've done a bit-wise comparison.

So I propose that you replace each call to this method with something like:
```cpp
bool allow_white = allowed_host_state & BaseResolver::WHITELISTED;
bool allow_black = allowed_host_state & BaseResolver::BLACKLISTED;
```

This is shorter than the existing code to declare the two bools and then pass them into this method.

src/baseresolver.cpp
```diff
@@ -121,11 +148,17 @@ void BaseResolver::srv_resolve(const std::string& srv_name,
...
                                int& ttl,
-                               SAS::TrailId trail)
+                               SAS::TrailId trail,
+                               int allowed_host_state)
 {
   // Accumulate blacklisted targets in case they are needed.
   std::vector<AddrInfo> blacklisted_targets;
 
+  // Initialise variables for allowed host states.
+  bool whitelisted_allowed, blacklisted_allowed;
```
huws: Avoid this.  There are cases where it does dangerous things.  Rather than remembering what those cases are and how to identify them, we just avoid the trap.  If someone were to change the type of these variables, they could make an error which causes part of that type to refer only to the first variable.

src/baseresolver.cpp
```diff
@@ -204,14 +237,18 @@ void BaseResolver::srv_resolve(const std::string& srv_name,
...
+            // Address isn't blacklisted and whitelisted addresses are allowed,
+            // so copy across to the active list.
             active.push_back(ai.address);
           }
-          else
+          else if ( addr_blacklisted && blacklisted_allowed )
           {
-            // Address is blacklisted, so copy to blacklisted list.
+            // Address is blacklisted and blacklisted addresses are allowed,
+            // so copy to blacklisted list.
```
huws: I think you should unconditionally add the addresses to the black/white lists, and then later decide whether to add them to the list of addresses that we'll actually attempt.  This allows us to log all the addresses concerned, which will help greatly with debugging.

src/baseresolver.cpp
```diff
@@ -913,6 +956,11 @@ LazyAddrIterator::LazyAddrIterator(DnsResult& dns_result,
 
 std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
 {
+  // Initialise variables for allowed host states.
+  bool whitelisted_allowed, blacklisted_allowed;
```
huws: Ditto

src/baseresolver.cpp
```diff
@@ -967,24 +1015,39 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
     AddrInfo result = _unused_results.back();
     _unused_results.pop_back();
 
-    if (_resolver->host_state(result) == BaseResolver::Host::State::WHITE)
+    if ( _resolver->host_state(result) == BaseResolver::Host::State::WHITE )
```
huws: You've done this in a couple of places.  Whitespace is not required immediately inside the brackets for an `if` test.

src/baseresolver.cpp
```diff
@@ -967,24 +1015,39 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
-      // Update logging.
-      found_blacklisted_str += result.address_and_port_to_string() + ";";
+        // Update logging.
+        found_blacklisted_str += result.address_and_port_to_string() + ";";
+        TRC_DEBUG("Found an unhealthy server");
+      }
+      else
+      {
+        TRC_DEBUG("Blacklisted hosts not allowed, ignoring blacklisted server");
+      }
```
huws: Again - I think you should build up the lists of address types, and then separately construct the set of addresses to try, based on the allowed types.

include/baseresolver.h
GAB-MS: I'm removing the method anyway, as we don't need it - but I wanted to check, this really doesn't do anything? The rest of the functions in the file replicate const in the header and source files, so I followed that trend. I'm kind of surprised if it's not needed in one of the places, given the potential for confusion?

include/baseresolver.h
GAB-MS: I'll change them to `const`, it seems like the right thing to do if they'll do the same thing and if readers will be more familiar with `const`.

GAB-MS:
Passing this PR, along with the others, to @AlexHockey for external review.

src/baseresolver.cpp
```diff
@@ -966,25 +995,37 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+      if (blacklisted_allowed)
+      {
+        // Add the record to the list of unhealthy targets.
+        _unhealthy_results.push_back(result);
 
-      // Update logging.
-      found_blacklisted_str += result.address_and_port_to_string() + ";";
+        // Update logging.
+        found_blacklisted_str += target;
+        TRC_DEBUG("Found an unhealthy server");
```
sathiyan-sivathas: Like the trace statement for whitelisted targets, it would be useful if this trace statement said how many unhealthy servers we have at this point.

PR #662, opened 2017-08-04
[Reviewer: Seb] Add parse_multi_site_stores_arg

richardwhiuk:
Seb,

We've got a whole bunch of code duplicated between Sprout and Ralf for no reason. This fixes that by moving the code to cpp-common.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -27,6 +27,7 @@
 #include <signal.h>
 #include <sys/stat.h>
 #include <syslog.h>
+#include <boost/algorithm/string.hpp>
 #include <boost/regex.hpp>
 
 #include "utils.h"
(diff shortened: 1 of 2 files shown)


src/utils.cpp
```diff
@@ -478,6 +479,42 @@ bool Utils::split_site_store(const std::string& site_store,
...
+                                 store_location,
+                                 remote_stores_locations))
+    {
+      TRC_ERROR("Invalid format of %s program argument", store_name);
+      return false;
+    }
+
+    if (store_location == "")
+    {
+      // If we've failed to find a local site registration store then Sprout has
```
sebrexmetaswitch: This is now common, so we shouldn't refer to Sprout

src/utils.cpp
```diff
@@ -478,6 +479,42 @@ bool Utils::split_site_store(const std::string& site_store,
...
+      // been misconfigured.
+      TRC_ERROR("No local site %s specified", store_name);
+      return false;
+    }
+    else
+    {
+      TRC_INFO("Using memcached %s", store_name);
```
sebrexmetaswitch: We only use this for memcached at the moment, but it seems like it's pretty generic to me, so we could potentially change this log line to something more generic?

src/utils.cpp
```diff
@@ -478,6 +479,42 @@ bool Utils::split_site_store(const std::string& site_store,
   }
 }
 
+bool Utils::parse_multi_site_stores_arg(const std::vector<std::string>& stores_arg,
```
sebrexmetaswitch: Can we have a comment like the one above for `Utils::split_site_store`?

PR #663, opened 2017-08-04
[Reviewer AJH] Add timeout override to HttpClient

steelyvoid:
Alex, are you OK to review this?  Its part of the fix for this "send provisioning traffic over the signaling interface" enhancement we discussed the other day.

The change I had to make to cpp-common here was to add a CURL timeout override to HttpClient.  I've tested this live and checked that the timeout is applied if that new parameter is passed.

I've also updated cpp-common-test (PR coming up)

### DIFF
--- a/include/http_connection_pool.h
+++ b/include/http_connection_pool.h
@@ -50,7 +50,8 @@ class HttpConnectionPool : public ConnectionPool<CURL*>
 {
 public:
   HttpConnectionPool(LoadMonitor* load_monitor,
-                     SNMP::IPCountTable* stat_table);
+                     SNMP::IPCountTable* stat_table,
+                     long timeout_ms = -1);
 
   ~HttpConnectionPool()
   {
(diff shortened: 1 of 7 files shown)


steelyvoid:
Also related to https://github.com/Metaswitch/bifrost/pull/71

AlexHockey:
At present the HTTP request timeout is controlled by the `HttpConnectionPool` class (https://github.com/Metaswitch/cpp-common/blob/master/src/http_connection_pool.cpp#L42). So I think you should either:

* Plumb the override value into the connection pool, or
* Have the HTTP client put the timeout back to what is was before returning the connection to the pool.

Given the timeout is at the scope as the entire client, and this is the same scope as the connection pool, I would probably go for the former.

AlexHockey:
:+1:

AlexHockey:
The latest changes look good too.

PR #664, opened 2017-08-09
[Reviewer: EM] Add method to calculate diameter timeout from target latency

sathiyan-sivathas:


sathiyan-sivathas:
I thought of that, but utils.cpp isn't covered, so I might get away it 😉

eleanor-merry:
☹️

PR #667, opened 2017-08-22
[Reviewer: Seb] Pass allowed_hosts through to httpconnection

johadalin:
This PR enables the use of allowed_host_state in baseresolver from the HttpConnection class. 
It is basically just adding the `allowed_host_state` variable to the methods that run through, and defaulting it to `ALL_LISTS`. There are then a few changes to test fixtures to get things working.

I've tested this running it in cpp-common-test, and that now all works (with associated PR to be linked below shortly)
I am also doing a search for anywhere that uses the fakehttp/mockhttp classes in their tests, and will link test fixes below, so that hopefully merging this doesn't kill builds.

### DIFF
--- a/include/a_record_resolver.h
+++ b/include/a_record_resolver.h
@@ -30,13 +30,15 @@ public:
                        int port,
                        int max_targets,
                        std::vector<AddrInfo>& targets,
-                       SAS::TrailId trail);
+                       SAS::TrailId trail,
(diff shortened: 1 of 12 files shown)


johadalin:
Couple of changes in latest commits in https://github.com/Metaswitch/sprout/pull/1901

johadalin:
Decided to take a different approach here, in the linked PR above, so closing this down.

PR #668, opened 2017-08-22
[Reviewer: Seb] Cass to memcached

sebrexmetaswitch:
This PR is for the cpp-common changes to support https://github.com/Metaswitch/homestead/pull/465


src/localstore.cpp
```diff
@@ -179,19 +202,20 @@ Store::Status LocalStore::set_data(const std::string& table,
...
+      // CAS matches, or record has expired and CAS is zero), or we aren't
+      // checking CAS values so update the record.
 
       // This writes data this is one update out-of-date to _old_db. This is for
       // the purposes of simulating data contention in Unit Testing.
       _old_db[fqkey] = r;
 
       r.data = data;
-      r.cas = ++cas;
+      r.cas = check_cas ? ++cas : (r.cas + 1);
```
sebrexmetaswitch: We're in an if block that's already checking that `check_cas` is true, so this doesn't feel necessary

src/memcachedstore.cpp
```diff
@@ -491,7 +467,91 @@ Store::Status TopologyNeutralMemcachedStore::set_data(const std::string& table,
...
+  Store::Status status = Store::Status::OK;
+  std::vector<AddrInfo> targets;
+  memcached_return_t rc;
+
+  // Memcached uses a flexible mechanism for specifying expiration.
+  // - 0 indicates never expire.
+  // - <= MEMCACHED_EXPIRATION_MAXDELTA indicates a relative (delta) time.
+  // - > MEMCACHED_EXPIRATION_MAXDELTA indicates an absolute time.
+  // Absolute time is the only way to force immediate expiry.  Unfortunately,
+  // it's not reliable - see https://github.com/Metaswitch/cpp-common/issues/160
```
sebrexmetaswitch: I know this comment was just copied, but that link is dead (this is why we don't paste links in code people!) so maybe just remove it?

src/localstore.cpp
richardwhiuk: Oops, the if block above is correct. I'll fix this.

PR #669, opened 2017-08-23
allowed_host_state plumbed through for http

johadalin:
The approach i took in https://github.com/Metaswitch/cpp-common/pull/667 was going to make life really painful all over, so its been thrown out, and this PR replaces it. 

Instead of just plumbing allowed_host_state through everything with a default value, i've added new methods for the functions that take the `allowed_host_state` parameter, and have the existing methods call across to it with the default value set.
This should mean we  won't hit as many issues in the code already using these functions, as the interface remains unchanged.

Tested, and both sprout (building from https://github.com/Metaswitch/sprout/pull/1901) and cpp-common-test build and pass all UTs on this.


johadalin:
Built sprout using these changes, and live tests pass fine.

PR #672, opened 2017-08-25
[Reviewer: SS5, AJH] Graylisting functionality added to SIP

Neel-Nanda:
Graylisting added to SIP in the BaseResolver class

Unit tests written and working with full coverage


src/baseresolver.cpp
```diff
@@ -227,40 +227,47 @@ BaseAddrIterator* BaseResolver::srv_resolve_iter(const std::string& srv_name,
...
           std::string target = "[" + ai.address_and_port_to_string() + "] ";
 
-          if (addr_state == Host::State::GRAY_NOT_PROBING && search_for_gray)
+          // If whitelisting is allowed, the first graylisted target reached
+          // will be selected for probing and put at the start of the list of
+          // targets - it is treated like it is whitelisted. However, if only
+          // blacklisting is allowed then it is not probed and treated as though
+          // it is blacklisted. So it is possible that one address will be a
+          // target for this request regardless of whether blacklisting or
+          // whitelisting are allowed.
```
sathiyan-sivathas: "whitelisting/blacklisting is allowed" isn't the right phrase here. You should say "If only whitelisted targets are requested..."

src/baseresolver.cpp
```diff
@@ -358,17 +367,19 @@ BaseAddrIterator* BaseResolver::srv_resolve_iter(const std::string& srv_name,
...
-      if (targets.size()>0)
+      if (targets.size() > 0)
       {
-        // We only want to probe a graylisted target if it is at the top
-        // priority level, and we ignore priority levels where all targets are
-        // blacklisted or graylisted and already being probed.
+        // We only want to probe a graylisted address if it is at the top
+        // priority level. We verify that at least one target was found, since
+        // if the top priority level had all addresses blacklisted we ignore it
+        // and treat the second highest level as though it was top.
```
sathiyan-sivathas: I think you should phrase this differently. Rather than saying that we treat the second priority level as the highest priority, it's better to say that we keep looking for graylisted targets at lower priority levels if we haven't found any whitelisted targets at higher priority levels. An example would be useful here e.g. if there are no whitelisted targets at priority level 1, we will keep looking for graylisted targets at level 2.

src/baseresolver.cpp
```diff
@@ -887,6 +906,21 @@ BaseResolver::Host::State BaseResolver::host_state(const AddrInfo& ai, time_t cu
...
+bool BaseResolver::blacklisted(const AddrInfo& ai)
+{
+  pthread_mutex_lock(&_hosts_lock);
+
+  Host::State state = host_state(ai);
+
+  pthread_mutex_unlock(&_hosts_lock);
+
+  return (state != Host::State::WHITE);
+
```
sathiyan-sivathas: Don't need this blank line.

src/baseresolver.cpp
```diff
@@ -1120,8 +1152,32 @@ pthread_mutex_lock(&(_resolver->_hosts_lock));
...
+      if ((_allowed_host_state == BaseResolver::WHITELISTED) ||
+          (_allowed_host_state == BaseResolver::BLACKLISTED))
+      {
+        // The search was restricted to either just blacklisted or just
+        // whitelisted addresses - there were none with the specified state.
+        bool blacklisted = (_allowed_host_state == BaseResolver::BLACKLISTED);
+        SAS::Event event(_trail, SASEvent::BASERESOLVE_NO_ALLOWED_RECORDS, 0);
+        event.add_var_param(_hostname);
+        event.add_static_param(blacklisted);
+        SAS::report_event(event);
```
sathiyan-sivathas: You've copy/pasted this code into two places. You should put it into a subfunction and pass it the information it requires.

src/baseresolver.cpp
```diff
@@ -887,6 +906,21 @@ BaseResolver::Host::State BaseResolver::host_state(const AddrInfo& ai, time_t cu
   return state;
 }
 
+/// Return true if the host state of ai is gray or black. Gray is considered
+/// blacklisted, even if ai is graylisted and the request where this function
+/// was called is currently probing ai
```
sathiyan-sivathas: "even if ai is graylisted and the calling code is currently probing it."

sathiyan-sivathas: Can you also say the given AddrInfo instead of ai like you've done in the header.

src/baseresolver.cpp
```diff
@@ -890,9 +890,13 @@ BaseResolver::Host::State BaseResolver::host_state(const AddrInfo& ai, time_t cu
   return state;
 }
 
+// LCOV_EXCL_START
+
+/// This is excluded from coverage because it is tested in SIPResolver.
```
sathiyan-sivathas: Can you put this on the same line as the `LCOV_EXCL_START` and be more specific than "this" e.g.

```
// LCOV_EXCL_START - This function is excluded from coverage because it is tested in the SIPResolver.
```

src/baseresolver.cpp
```diff
@@ -890,9 +890,13 @@ BaseResolver::Host::State BaseResolver::host_state(const AddrInfo& ai, time_t cu
...
+
```
sathiyan-sivathas: Can you delete this newline.

src/baseresolver.cpp
```diff
@@ -904,6 +908,8 @@ bool BaseResolver::blacklisted(const AddrInfo& ai)
   return (state != Host::State::WHITE);
 }
```
sathiyan-sivathas: And this one.

AlexHockey:
Closing. Code has been merged under https://github.com/Metaswitch/cpp-common/pull/694

PR #674, opened 2017-09-01
Allow RTRs to be searchable in SAS

metaswitch-colin-haywood:
[Reviewer AJH] pjutils function refactored into cpp-common

Companion PR to https://github.com/Metaswitch/homestead/pull/470

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -19,6 +19,7 @@
 #include <list>
 #include <queue>
 #include <string>
+#include <boost/regex.hpp>
 #include <arpa/inet.h>
 
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


src/utils.cpp
```diff
@@ -202,6 +203,54 @@ std::string Utils::xml_escape(const std::string& s)
...
+  {
+    if ((user[i] == '+') ||
+        (user[i] == '-') ||
+        (user[i] == '.') ||
+        (user[i] == '(') ||
+        (user[i] == ')') ||
+        (user[i] == '[') ||
+        (user[i] == ']') ||
+        ((user[i] >= '0') &&
+         (user[i] <= '9')))
```
AlexHockey: A very minor thing - it might be more efficient to put the `0 <= user[i] <= 9` test as the first condition. Most tel URIs will just contains numbers, and this prevent having to evaluate the 7 other conditions. Normally I wouldn't bother with such a small optimization but as it is so low cost we might as well make it.

metaswitch-colin-haywood:
Closing as this was already committed from a different branch.

PR #675, opened 2017-09-01
[Reviewer=CJR, KH1] Add ability to customize the server details we send to SAS

jack-atack:
This PR give us the ability to log a different server address to the SAS call flow, than the actual address in the HTTP request.  An example use case is when sending requests to localhost, where we may wish to instead log the public/private IP of the node in SAS

I don't think that the arguments I have added to the constructor necessarily belong at the end, but I have added them at the end to ensure this is a non-breaking change for existing uses of these classes.

I have tested this on my deployment and run the live test suite, and `make full_test` passes in Sprout

### DIFF
--- a/include/httpclient.h
+++ b/include/httpclient.h
@@ -66,7 +66,9 @@ public:
              BaseCommunicationMonitor* comm_monitor,
              bool should_omit_body = false,
              bool remote_connection = false,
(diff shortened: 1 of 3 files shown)


GAB-MS:
Jack's out today, but we want this in today, so I'm going to merge it in for him.

PR #676, opened 2017-09-01
[Reviewer: AJH] Disable Nagle's algorithm in memcached

sathiyan-sivathas:
Now that I'm adding connection pooling to astaire, I've moved the code to disable Nagle's algorithm into the memcached connection pool code. I think this is ok because of the last part of the comment which says "libmemcached's buffers are large enough that it will never send small message fragments, and this store's implementation means we very rarely pipeline requests".

I've also tidied up some other code:
- I've moved `parse_ip_target()` from the Baseresolver to Utils since that's where all the other address parsing utility functions live. It also means I don't need to include baseresolver.h to use this function.
- We have 3 implementations of a function to turn an AddrInfo into a string. I've consolidated all this in the AddrInfo class and will be deleting the other functions.

### DIFF
--- a/src/memcached_connection_pool.cpp
+++ b/src/memcached_connection_pool.cpp
@@ -20,6 +20,28 @@ memcached_st* MemcachedConnectionPool::create_connection(AddrInfo target)
   memcached_behavior_set(conn,
                          MEMCACHED_BEHAVIOR_CONNECT_TIMEOUT,
                          _max_connect_latency_ms);
+
+  // Disable Nagle's algorithm
+  // (https://en.wikipedia.org/wiki/Nagle%27s_algorithm). If we leave it on
+  // there can be up to 500ms delay between this code sending an
+  // asynchronous SET and it actually being sent on the wire, e.g.
+  //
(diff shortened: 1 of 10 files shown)


src/memcached_connection_pool.cpp
```diff
@@ -20,6 +20,28 @@ memcached_st* MemcachedConnectionPool::create_connection(AddrInfo target)
...
+  // * Ask libmemcached to do a 2nd async SET.
+  // * Up to 500ms passes.
+  // * TCP stack receives ACK to 1st SET (may be delayed because the server
+  //   does not send a protocol level response to the async SET).
+  // * 2nd async SET sent on the wire (up to 500ms late).
+  //
+  // This delay can open up window conditions in failure scenarios. In
+  // addition there is not much point in using Nagle. libmemcached's buffers
+  // are large enough that it will never send small message fragments, and
+  // this store's implementation means we very rarely pipeline requests.
```
AlexHockey: Referring to "this store" doesn't make sense now that the code is in the connection pool.

PR #679, opened 2017-09-12
Add resource bundle comment

huws:
Add a comment to the resource bundle datestamp line, making it clear what needs to be updated if you ever do need to change this value by hand.  This should help to avoid issues where we ship releases in which different components depend on different SAS resource bundles.


huws:
Holding off on merging this trivial change until the code is unfrozen.

PR #685, opened 2017-09-18
[Reviewer AJH] (DO NOT MERGE!) Lazy Targeting Functionality Added to SIP

Neel-Nanda:
DO NOT MERGE! Created for code review purposes

Lazy Targeting added to SIP in the BaseResolver class
Changed TTLCache to use shared pointers

Passes all current UTs, no new UTs written yet. Incomplete coverage.

### DIFF
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -127,27 +127,6 @@ protected:
   /// Converts a DNS A or AAAA record to an IP46Address structure.
   IP46Address to_ip46(const DnsRRecord* rr);
 
-  /// Helper function to create SAS logs if no targets were resolved. Says if
-  /// this was because only whitelisted or blacklisted targets were requested,
-  /// or if there were no records at all for that address
-  void no_targets_resolved_logging(const std::string name,
-                                   SAS::TrailId trail,
-                                   bool whitelisted_allowed,
-                                   bool blacklisted_allowed);
-
-  /// Helper function to allow LazySRVResovleIter to call A Record DNS
-  /// Resolution when preparing a priority level
-  void BaseResolver::dns_query(std::vector<std::string>& domains,
-                               int dnstype,
-                               std::vector<DnsResult>& results,
-                               SAS::TrailId trail);
-
-  /// Helper function for LazySRVResolveIter. Returns a pointer to an SRV
-  /// Priority List for the given SRV name.
-  std::shared_ptr<SRVPriorityList> BaseResolver::get_srv_list(const std::string& srv_name,
-                                                              int &ttl,
-                                                              SAS::TrailId trail);
-
   /// Holds the results of applying NAPTR replacement on a target domain name.
   struct NAPTRReplacement
   {
@@ -310,6 +289,27 @@ protected:
   /// Indicates that the calling thread is selected to probe the given AddrInfo.
   /// _hosts_lock must be held when calling this method.
   void select_for_probing(const AddrInfo& ai);
+
+  /// Helper function to create SAS logs if no targets were resolved. Says if
+  /// this was because only whitelisted or blacklisted targets were requested,
+  /// or if there were no records at all for that address
+  void no_targets_resolved_logging(const std::string name,
+                                   SAS::TrailId trail,
+                                   bool whitelisted_allowed,
+                                   bool blacklisted_allowed);
+
+  /// Helper function to allow LazySRVResovleIter to call A Record DNS
+  /// Resolution when preparing a priority level
+  void dns_query(std::vector<std::string>& domains,
+                               int dnstype,
+                               std::vector<DnsResult>& results,
+                               SAS::TrailId trail);
+
+  /// Helper function for LazySRVResolveIter. Returns a pointer to an SRV
+  /// Priority List for the given SRV name.
+  std::shared_ptr<SRVPriorityList> get_srv_list(const std::string& srv_name,
+                                                              int &ttl,
+                                                              SAS::TrailId trail);
 
   int _default_blacklist_duration;
   int _default_graylist_duration;
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -298,18 +298,19 @@ protected:
                                    bool whitelisted_allowed,
                                    bool blacklisted_allowed);
 
-  /// Helper function to allow LazySRVResovleIter to call A Record DNS
-  /// Resolution when preparing a priority level
+  /// Allows DNS Resolution to be called with a pointer to the Base Resolver.
+  /// This just returns the (possibly cached) result of a DNS Query, so any
+  /// post-processing of the BaseResolver is returned.
   void dns_query(std::vector<std::string>& domains,
-                               int dnstype,
-                               std::vector<DnsResult>& results,
-                               SAS::TrailId trail);
+                 int dnstype,
+                 std::vector<DnsResult>& results,
+                 SAS::TrailId trail);
 
-  /// Helper function for LazySRVResolveIter. Returns a pointer to an SRV
+  /// Helper function to perform SRV Record DNS Resolution via the SRV Cache and returns a pointer to an SRV
   /// Priority List for the given SRV name.
   std::shared_ptr<SRVPriorityList> get_srv_list(const std::string& srv_name,
-                                                              int &ttl,
-                                                              SAS::TrailId trail);
+                                                int &ttl,
+                                                SAS::TrailId trail);
 
   int _default_blacklist_duration;
   int _default_graylist_duration;
@@ -355,7 +356,7 @@ private:
 };
 
 // AddrInfo iterator that uses the blacklist system of a BaseResolver to lazily
-// select targets from A Record Resolution.
+// select targets using A Record Resolution.
 class LazyAResolveIter : public BaseAddrIterator
 {
 public:
@@ -396,7 +397,7 @@ private:
 };
 
 // AddrInfo iterator that uses the blacklist system of a BaseResolver to lazily
-// select targets from SRV Record Resolution.
+// select targets using SRV Record Resolution.
 class LazySRVResolveIter : public BaseAddrIterator
 {
 public:
@@ -429,7 +430,7 @@ private:
   /// _search_for_gray is true, then the entire priority level will be searched
   /// for an unprobed graylisted address. If one is found then _gray_found will
   /// be set to true and it will be stored in _unprobed_gray_target and not in
-  /// either 2D vector.
+  /// the *_addresses_by_srv vectors.
   ///
   /// Returns true if the priority level was prepared successfully, false if
   /// there were no priority levels left to prepare
@@ -443,8 +444,8 @@ private:
   /// adds black and gray addresses to targets and nothing to
   /// _unhealthy_targets.
   ///
-  /// Two strings are passed in to log the whitelisted and unhealthy targets
-  /// found, so that take can log a SAS event with them.
+  /// Two strings are passed out to log the whitelisted and unhealthy targets
+  /// found, so that the take method can log a SAS event with them.
   ///
   /// Returns the number of targets that are still to be found, or 0 if
   /// num_targets_to_find were all found. If all targets were found, the search
@@ -466,13 +467,12 @@ private:
 
   int _af;
   int _transport;
+  std::string _srv_name;
 
   // The smallest time to live found so far, from all DNS Resolutions. Updated
   // by prepare_priority_level if the DNS resolution finds a smaller time to
   // live.
   int _ttl;
-
-  std::string _srv_name;
 
   // Pointer to a map from priority levels to a vector of SRVs for that priority
   // level.
@@ -499,12 +499,8 @@ private:
   // Unhealthy is used as a catch-all term for blacklisted and graylisted
   // addresses, since graylisted addresses which are not being probed by this
   // request are treated the same as blacklisted addresses.
-  std::vector<std::vector<IP46Address> > _whitelisted_addresses_by_srv;
-  std::vector<std::vector<IP46Address> > _unhealthy_addresses_by_srv;
-
-  // Where prepare_priority_level stores the SRVs from the current priority
-  // level after putting them in a random order with a weighted selector.
-  std::vector<const BaseResolver::SRV*> _srvs;
+  std::vector<std::vector<AddrInfo> > _whitelisted_addresses_by_srv;
+  std::vector<std::vector<AddrInfo> > _unhealthy_addresses_by_srv;
 
   // Vector where get_from_priority_level stores the black or graylisted targets
   // it finds when both whitelisted and blacklisted targets are desired.  Uses
@@ -513,18 +509,29 @@ private:
   // requested then this vector is left empty.
   std::vector<AddrInfo> _unhealthy_targets;
 
-  // The index of the SRV get_from_priority_level is currently looking at, so
-  // that if the number of targets required is found before the entire priority
-  // level is searched, get_from_priority_level can pause its search of the SRV
-  // Records and resume from the same place if take is called again.
-  int _current_srv;
-
   // Boolean to track whether there's any addresses left in the current priority
   // level. Initialised to false at the start and whenever a new priority level
   // is prepared, and is then set to true once an SRV is found which will
   // have addresses left the next time get_from_priority_level scans through
   // each SRV at the current priority level.
   bool _more_in_priority_level;
+
+  // Boolean to track whether every address in the current priority level, ie.
+  // the one last prepared, has been looked at by get_from_priority_level. If it
+  // is true and more targets are needed, the take method will call
+  // prepare_priority_level. Initialised to true, since before
+  // prepare_priority_level is ever called there is no current priority level.
+  bool _finished_priority_level;
+
+  // The following 3 data members track the current position in various vectors
+  // and maps of the iterator, allowing the lazy iterator to pause algorithms
+  // once enough targets have been found, and resume the next time it's called.
+
+  // The index of the SRV get_from_priority_level is currently looking at, so
+  // that if the number of targets required is found before the entire priority
+  // level is searched, get_from_priority_level can pause its search of the SRV
+  // Records and resume from the same place if take is called again.
+  int _current_srv;
 
   // The index of the unhealthy target to return next. Ensures that in
   // subsequent calls to take the same unhealthy target is not returned twice.
--- a/include/baseresolver.h
+++ b/include/baseresolver.h
@@ -135,6 +135,19 @@ protected:
                                    bool whitelisted_allowed,
                                    bool blacklisted_allowed);
 
+  /// Helper function to allow LazySRVResovleIter to call A Record DNS
+  /// Resolution when preparing a priority level
+  void BaseResolver::dns_query(std::vector<std::string>& domains,
+                               int dnstype,
+                               std::vector<DnsResult>& results,
+                               SAS::TrailId trail);
+
+  /// Helper function for LazySRVResolveIter. Returns a pointer to an SRV
+  /// Priority List for the given SRV name.
+  std::shared_ptr<SRVPriorityList> BaseResolver::get_srv_list(const std::string& srv_name,
+                                                              int &ttl,
+                                                              SAS::TrailId trail);
+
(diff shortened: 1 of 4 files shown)


src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+    TRC_VERBOSE("Do A record look-ups for %ld SRVs", a_targets.size());
+    _resolver->dns_query(a_targets,
+                         (_af == AF_INET) ? ns_t_a : ns_t_aaaa,
+                         a_results,
+                         _trail);
+
+    // Clear the two vectors where the results are stored, to store the results
+    // from this priority level
+    _whitelisted_addresses_by_srv.clear();
+    _unhealthy_addresses_by_srv.clear();
```
AlexHockey: It would be easier to read if this function cleared out all the members it needs to reuse in one place at the start of the function.

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+
+    for (size_t ii = 0; ii < _srvs.size(); ++ii)
+    {
+      DnsResult& a_result = a_results[ii];
+      TRC_DEBUG("SRV %s:%d returned %ld IP addresses",
+                _srvs[ii]->target.c_str(),
+                _srvs[ii]->port,
+                a_result.records().size());
+      std::vector<IP46Address> whitelisted_addresses;
```
AlexHockey: Could this be a reference into `_whitelisted_addresses_by_srv`. That would probably be more efficient and matches what you are trying to do more closely.

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+        BaseResolver::Host::State addr_state = _resolver->host_state(ai);
+        std::string target = "[" + ai.address_and_port_to_string() + "] ";
+
+        // If whitelisted targets are requested, the first unprobed graylisted
+        // target reached will be selected for probing and put at the start of
+        // the list of targets, it is treated like it is whitelisted. However,
+        // if only blacklisted targets are requested then this function will not
+        // search for a graylisted target to probe, all graylisted addresses
+        // will be treated as though they were blacklisted.
+        if (addr_state == BaseResolver::Host::State::GRAY_NOT_PROBING && _search_for_gray && _whitelisted_allowed)
```
AlexHockey: Please put brackets around compound subclauses in conditionals. e.g. `if ((a == b) || c)`

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+          // which we'll select targets.
+          if (_whitelisted_allowed)
+          {
+            whitelisted_addresses.push_back(ai.address);
+          }
+        }
+        else
+        {
+          // Address is blacklisted or graylisted and not being probed by this
+          // request.  If we're allowed to return blacklisted targets, also add
```
AlexHockey: Why do you say "also" here?

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+                                                int num_targets_to_find,
+                                                int num_requested_targets,
+                                                std::string& whitelisted_targets_str,
+                                                std::string& unhealthy_targets_str)
+{
+  // Select the appropriate number of targets by looping through the SRV records
+  // taking one address each time until either we have enough for the number of
+  // retries allowed, or we have no more addresses.
+  //
+  // This takes one target from each site at this priority level and then
```
AlexHockey: Stop using site!!!!!! 😠

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+    ai.transport = _transport;
+
+    // This lock must be held to safely call the host_state method of
+    // BaseResolver.
+    pthread_mutex_lock(&(_resolver->_hosts_lock));
+
+    // The for loop is only initialised if on the previous iteration it went
+    // through all SRVs. This lets get_from_priority_level pause the for loop if
+    // it finds enough targets and resume when it is next called.
+    if ((size_t) _current_srv == _srvs.size())
```
AlexHockey: No space between a cast's type and the value being cast.

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+  {
+    more = false;
+    AddrInfo ai;
+    ai.transport = _transport;
+
+    // This lock must be held to safely call the host_state method of
+    // BaseResolver.
+    pthread_mutex_lock(&(_resolver->_hosts_lock));
+
+    // The for loop is only initialised if on the previous iteration it went
```
AlexHockey: You're wording here is confusing. I think you mean to say "If we're at the end of the SRV records, start from the beginning"

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+    TRC_DEBUG("All priority levels have been prepared and searched for targets of the desired host state.");
+    // Returns false to show that there was not a valid priority level to
+    // prepare.
+    return false;
+  }
+}
+
+int LazySRVResolveIter::get_from_priority_level(std::vector<AddrInfo> &targets,
+                                                int num_targets_to_find,
+                                                int num_requested_targets,
```
AlexHockey: Mark this as const (as if it changes, you're logs end up being wrong)

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+        // there are insufficiently many whitelisted addresses over all priority
+        // levels for take to return.
+        ai.address = _unhealthy_addresses_by_srv[_current_srv].back();
+        _unhealthy_addresses_by_srv[_current_srv].pop_back();
+        _unhealthy_targets.push_back(ai);
+      }
+
+      more = more ||
+             ((!_whitelisted_addresses_by_srv[_current_srv].empty()) ||
+             (!_unhealthy_addresses_by_srv[_current_srv].empty()));
```
AlexHockey: You do `_whitelisted_addresses_by_srv[_current_srv]` and `_unhealthy_addresses_by_srv[_current_srv]`. Is it worth having turning these into local variables initialized at the start of the loop body?

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+    }
+
+    // host_state no longer needs to be called, so this lock can be relaxed.
```
AlexHockey: Relaxed??? I think you mean released 😉

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+  std::string whitelisted_targets_str;
+  std::string unhealthy_targets_str;
+  std::string graylisted_targets_str;
+
+  // Tracks the number of targets left to be added.
+  int num_targets_to_find = num_requested_targets;
+
+  if (_srv_list != nullptr)
+  {
+    // If srv_list is null, DNS resolution found no SRVs, so an empty vector
```
AlexHockey: Don;t talk about the list being NULL in the branch where is in non-NULL.

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+    // should be returned.
+    while (num_targets_to_find > 0)
+    {
+      // If this is the first request, or the previous priority level has been
+      // finished, then both vectors of addresses_by_srv will be empty and
+      // get_from_priority_level will do nothing and return the same number of
+      // targets. Otherwise, this resumes the search of the current priority
+      // level from a previous call of take.
+      num_targets_to_find = get_from_priority_level(targets,
```
AlexHockey: Do you not need to call parse priority level first??

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+
+      if (targets.size() > 0)
+      {
+        // If a graylisted target is probed by this request, it must be the
+        // first target to guarantee that it is probed. It is important to
+        // respect priority levels, so if targets have been found at a higher
+        // priority level this request should not search for gray targets to
+        // probe.
+        _search_for_gray=false;
+      }
```
AlexHockey: I think `get_from_priority_level` should be handling this.

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+        {
+          // A graylisted target will only be returned if this is the first time
+          // take was called for this iterator. It is guaranteed that this
+          // target will be the first target returned.
+          targets.push_back(_unprobed_gray_target);
+          _resolver->select_for_probing(_unprobed_gray_target);
+          _gray_found = false;
+          --num_targets_to_find;
+          TRC_DEBUG("Added a graylisted server for probing to targets, now have 1 of %d", num_requested_targets);
+        }
```
AlexHockey: I would expect this to be handled by `get_from_priority_level`. It certainly seems wrong to be doing this after preparing the next priority level!

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+      }
+    }
+
+    if (add_unhealthy && _whitelisted_allowed && _blacklisted_allowed)
```
AlexHockey: It would be better to set `add_unhealthy` correctly based on the value of the other flags. This test looks like you have made the wrong decision earlier on, and then changed your mind.

src/baseresolver.cpp
```diff
@@ -1168,11 +894,467 @@ std::vector<AddrInfo> LazyAddrIterator::take(int num_requested_targets)
...
+  if (_trail != 0)
+  {
+    SAS::Event event(_trail, SASEvent::BASERESOLVE_SRV_RESULT, 0);
+    event.add_var_param(_srv_name);
+    event.add_var_param(graylisted_targets_str);
+    event.add_static_param(_whitelisted_allowed);
+    event.add_var_param(whitelisted_targets_str);
+    event.add_static_param(_blacklisted_allowed);
+    event.add_var_param(unhealthy_targets_str);
+    SAS::report_event(event);
```
AlexHockey: Please keep the static params and the var params together.

include/baseresolver.h
```diff
@@ -298,6 +290,27 @@ class BaseResolver
...
+  void dns_query(std::vector<std::string>& domains,
+                               int dnstype,
+                               std::vector<DnsResult>& results,
+                               SAS::TrailId trail);
+
+  /// Helper function for LazySRVResolveIter. Returns a pointer to an SRV
+  /// Priority List for the given SRV name.
+  std::shared_ptr<SRVPriorityList> get_srv_list(const std::string& srv_name,
+                                                              int &ttl,
+                                                              SAS::TrailId trail);
```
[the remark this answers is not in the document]
Neel-Nanda: My logic was that it seemed neater to get the SRV list from the Lazy Iterator, as this meant fewer parameters needed to be passed in. Deleting this function and passing in the list of SRVs and time to live to the Lazy Iterator when creating it would work equally well, if you think that would be better?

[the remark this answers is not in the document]
AlexHockey: I asked my question because this doesn't match the design you wrote up. There is a mental cost to adding new methods and we need to weigh this up against the cost of making the constructor have more parameters. I don't care too much, but you could have saved some time by flagging in the PR that you deviated from the design.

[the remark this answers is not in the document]
AlexHockey: I've thought about this a bit more. I think what I don't like is that you've got two methods on `BaseResolver` than can both be used to perform SRV lookups but that work in different ways. Maybe the solution is to comment `dns_query` to say that this just returns the (possibly cached) result of a DNS query, and bypasses any post processing the BaseResolver has done.

include/baseresolver.h
```diff
@@ -298,18 +298,19 @@ class BaseResolver
                                    bool whitelisted_allowed,
                                    bool blacklisted_allowed);
 
-  /// Helper function to allow LazySRVResovleIter to call A Record DNS
-  /// Resolution when preparing a priority level
+  /// Allows DNS Resolution to be called with a pointer to the Base Resolver.
+  /// This just returns the (possibly cached) result of a DNS Query, so any
+  /// post-processing of the BaseResolver is returned.
```
AlexHockey: *not* returned right?

include/baseresolver.h
```diff
@@ -298,18 +298,19 @@ class BaseResolver
...
-                               int dnstype,
-                               std::vector<DnsResult>& results,
-                               SAS::TrailId trail);
+                 int dnstype,
+                 std::vector<DnsResult>& results,
+                 SAS::TrailId trail);
 
-  /// Helper function for LazySRVResolveIter. Returns a pointer to an SRV
+  /// Helper function to perform SRV Record DNS Resolution via the SRV Cache and returns a pointer to an SRV
   /// Priority List for the given SRV name.
```
AlexHockey: line looks too long.

include/baseresolver.h
```diff
@@ -513,12 +509,34 @@ class LazySRVResolveIter : public BaseAddrIterator
   // requested then this vector is left empty.
   std::vector<AddrInfo> _unhealthy_targets;
 
+  // Boolean to track whether there's any addresses left in the current priority
+  // level. Initialised to false at the start and whenever a new priority level
+  // is prepared, and is then set to true once an SRV is found which will
+  // have addresses left the next time get_from_priority_level scans through
+  // each SRV at the current priority level.
+  bool _more_in_priority_level;
```
AlexHockey: Can't you work this out based on whether any of the `*_addresses_by_srv` vectors are still populated?

include/baseresolver.h
```diff
@@ -513,12 +509,34 @@ class LazySRVResolveIter : public BaseAddrIterator
...
+
+  // Boolean to track whether every address in the current priority level, ie.
+  // the one last prepared, has been looked at by get_from_priority_level. If it
+  // is true and more targets are needed, the take method will call
+  // prepare_priority_level. Initialised to true, since before
+  // prepare_priority_level is ever called there is no current priority level.
+  bool _finished_priority_level;
```
AlexHockey: Why is this not the same as the previous member?

src/baseresolver.cpp
```diff
@@ -812,19 +815,23 @@ std::vector<AddrInfo> LazyAResolveIter::take(int num_requested_targets)
...
-  // required number is reached, or the unused results are exhausted.
-  while ((_unused_results.size() > 0) && (targets.size() < (size_t)num_requested_targets))
+  // required number of targets is reached, or the unused results are exhausted.
+  // If only blacklisted targets are allowed to be returned, the first condition
+  // is instead until the required number of unhealthy results is reached.
+  while ((_unused_results.size() > 0) &&
+         ((whitelisted_allowed && (targets.size() < (size_t)num_requested_targets)) ||
+          (blacklisted_allowed &&
+           !whitelisted_allowed &&
+           (_unhealthy_results.size() < (size_t) num_requested_targets))))
```
AlexHockey: This loop condition is really hard to understand. Wouldn't it be easier to just have the code below that's dealing with blacklisted targets add directly to `targets` if only blacklisted targets are allowed?

src/baseresolver.cpp
```diff
@@ -974,14 +978,36 @@ std::vector<AddrInfo> LazySRVResolveIter::take(int num_requested_targets)
...
+        // priority levels have already been searched for targets of the desired
+        // state.
+        if (!prepare_priority_level())
+        {
+          if (_whitelisted_allowed && _blacklisted_allowed)
+          {
+            // All whitelisted targets have been returned but not enough targets
+            // have been found, and both types of host state were requested, so
+            // unhealthy targets should be returned.
+            add_unhealthy = true;
```
AlexHockey: Better to do `add_unhealthy = (_whitelisted_allowed && _blacklisted_allowed);`

src/baseresolver.cpp
```diff
@@ -990,61 +1016,30 @@ std::vector<AddrInfo> LazySRVResolveIter::take(int num_requested_targets)
...
 
-    if (add_unhealthy && _whitelisted_allowed && _blacklisted_allowed)
+    if (add_unhealthy)
     {
       // This adds unhealthy targets until either no unhealthy targets are left,
       // or the number of requested targets has been reached. If only whitelisted
       // or blacklisted addresses were requested then _unhealthy_targets is empty,
       // addresses of the desired host state were added straight to targets.
-      size_t to_copy = std::min((size_t) num_targets_to_find, _unhealthy_targets.size());
+      size_t to_copy = std::min((size_t) num_targets_to_find, _unhealthy_targets.size()-_unhealthy_target_pos);
```
AlexHockey: Spaces around operators please.

src/baseresolver.cpp
```diff
@@ -990,61 +1016,30 @@ std::vector<AddrInfo> LazySRVResolveIter::take(int num_requested_targets)
...
 
       TRC_VERBOSE("Adding %ld unhealthy servers", to_copy);
-      for (size_t ii = 0; ii < to_copy; ++ii)
+
+      int max_unhealthy_index = to_copy + _unhealthy_target_pos;
+      // The for loop is initialised in the class constructor, as the for loop
+      // needs to be paused and resumed in subsequent calls to take, so that the
+      // same unhealthy targets are not returned several times.
+      for (; _unhealthy_target_pos < max_unhealthy_index; ++_unhealthy_target_pos)
```
AlexHockey: This all seems a bit convoluted. You already know how many things to copy. Just have a simple for loop (`for(int i = 0; i < to_copy; ++i`) and adjust the indices in the loop body.

src/baseresolver.cpp
```diff
@@ -1238,51 +1242,72 @@ bool LazySRVResolveIter::prepare_priority_level()
...
-                                                int num_requested_targets,
+                                                const int num_requested_targets,
                                                 std::string& whitelisted_targets_str,
                                                 std::string& unhealthy_targets_str)
 {
+  // An AddrInfo object to temporarily store the current AddrInfo object being
+  // looked at from a *_addresses_by_srv vector.
+  AddrInfo ai;
+
+  if (_gray_found && (num_targets_to_find > 0))
```
AlexHockey: Why does there being targets to find imply that the iterator has not yet returned a target? Does this suggest your unit testing is lacking?

src/baseresolver.cpp
```diff
@@ -1330,31 +1355,58 @@ int LazySRVResolveIter::get_from_priority_level(std::vector<AddrInfo> &targets,
...
+      _more_in_priority_level = _more_in_priority_level ||
+                                ((!whitelisted_addresses.empty()) ||
+                                 (!unhealthy_addresses.empty()));
     }
 
-    // host_state no longer needs to be called, so this lock can be relaxed.
+    // host_state no longer needs to be called, so this lock can be
+    // released.
     pthread_mutex_unlock(&(_resolver->_hosts_lock));
+    TRC_DEBUG("@@@ END OF WHILE! %d %d", num_targets_to_find, (int) _more_in_priority_level);
```
AlexHockey: Log should be removed.

src/baseresolver.cpp
AlexHockey: Ah OK, you only search for a gray once in prepare priority level. Your comment below needs correcting then.

AlexHockey:
Closing code has been merged under https://github.com/Metaswitch/cpp-common/pull/694

PR #686, opened 2017-09-19
[Reviewer: Richard] Don't do unnecessary work behind the connection pool lock

sebrexmetaswitch:
Creating and destroying connections takes an unknown amount of time, so we must
never do them behind the connection pool's lock

Specifically, memcached connections require actually connecting to the memcached server when created. When talking to a remote site, this may take up to 200ms, so we really can't do this behind the lock.

This is a fix for: https://github.com/Metaswitch/clearwater-issues/issues/2413

### DIFF
--- a/include/connection_pool.h
+++ b/include/connection_pool.h
@@ -206,7 +206,7 @@ ConnectionHandle<T> ConnectionPool<T>::get_connection(AddrInfo target)
             target.address.to_string().c_str(),
             target.port);
 
-  ConnectionInfo<T>* conn_info_ptr;
+  ConnectionInfo<T>* conn_info_ptr = nullptr;
 
   pthread_mutex_lock(&_conn_pool_lock);
 
--- a/include/connection_pool.h
+++ b/include/connection_pool.h
@@ -217,17 +217,15 @@ ConnectionHandle<T> ConnectionPool<T>::get_connection(AddrInfo target)
     // If there is a connection in the pool for the given AddrInfo, retrieve it
     conn_info_ptr = slot_it->second.front();
     slot_it->second.pop_front();
-    pthread_mutex_unlock(&_conn_pool_lock);
     TRC_DEBUG("Found existing connection %p in pool", conn_info_ptr);
   }
-  else
+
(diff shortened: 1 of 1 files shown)


include/connection_pool.h
```diff
@@ -216,19 +217,22 @@ ConnectionHandle<T> ConnectionPool<T>::get_connection(AddrInfo target)
...
+    // We don't need to create the connection behind the lock, so release the
+    // lock first
+    pthread_mutex_unlock(&_conn_pool_lock);
+
     TRC_DEBUG("No existing connection in pool, create one");
     conn_info_ptr = new ConnectionInfo<T>(create_connection(target), target);
     TRC_DEBUG("Created new connection %p", conn_info_ptr);
   }
 
-  pthread_mutex_unlock(&_conn_pool_lock);
```
richardwhiuk: I don't like this change, because the scoping no longer matches the locking, which makes introducing bugs much easier.

I think it'd be better to do:

```
ConnectionInfo<T>* conn_info_ptr = nullptr;

....

pthread_mutex_lock(&_conn_pool_lock);
...
pthread_mutex_unlock(&_conn_pool_lock);
...

if (!conn_info_ptr)
{
     conn_info_ptr = new ConnectionInfo<T>(create_connection(target), target);
     TRC_DEBUG("Created new connection %p", conn_info_ptr);
}
```

include/connection_pool.h
```diff
@@ -290,6 +306,7 @@ template<typename T>
 void ConnectionPool<T>::free_old_connection()
 {
   time_t current_time = time(NULL);
+  ConnectionInfo<T>* conn_to_destroy = NULL;
```
richardwhiuk: I'd prefer `= nullptr`.

include/connection_pool.h
sebrexmetaswitch: Yeah, I was considering doing this anyway, so I'll make that change

include/connection_pool.h
sebrexmetaswitch: I just used `NULL` because that's what we use elsewhere in the file

PR #689, opened 2017-09-22
[Reviewer: Seb] Allow get/set of data without SAS log of the body

richardwhiuk:
Seb,

Currently we always SAS log the data we write to memcached. I've added the ability so we can choose not to do this on a case by case basis.

I've added an extra method (rather than adding an extra default parameter) to avoid having to fix all of the mocks.


include/sasevent.h
```diff
@@ -127,6 +127,9 @@ namespace SASEvent {
   const int MEMCACHED_NO_HOSTS = COMMON_BASE + 0x00010D;
   const int MEMCACHED_TRY_HOST = COMMON_BASE + 0x00010E;
   const int MEMCACHED_SET_WITHOUT_CAS_START = COMMON_BASE + 0x00010F;
+  const int MEMCACHED_GET_WITHOUT_DATA_SUCCESS = COMMON_BASE + 0x000110;
+  const int MEMCACHED_SET_WITHOUT_DATA_START = COMMON_BASE + 0x000111;
+  const int MEMCACHED_SET_WITHOUT_DATA_OR_CAS_START = COMMON_BASE + 0x000112;
```
sebrexmetaswitch: I'm not in love with the naming here. `SET_WITHOUT_DATA` is a bit misleading. It probably doesn't matter though

PR #691, opened 2017-10-09
[Reviewer: Mat] Only add a Content-Type header if we have a body

richardwhiuk:
Mat,

This resolves a bug we've seen whereby we always add a Content-Type header.

This was originally the case (see https://github.com/Metaswitch/cpp-common/blob/11222e20d4fe50adb17b6fb3db05a3937f01a800/src/httpconnection.cpp#L495-L499 ) but got lost in some refactoring.

### DIFF
--- a/include/httpclient.h
+++ b/include/httpclient.h
@@ -125,7 +125,7 @@ public:
   ///                           data
   /// @param response           Location to store retrieved data
   /// @param trail              SAS trail to use
-  /// @param body               Body to send on the request
+  /// @param body               JSON body to send on the request
   /// @param username           Username to assert if assertUser is true, else
   ///                           ignored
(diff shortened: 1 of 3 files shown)


MatMeredith:
My issue with this is that I don't like the API.   I call a generic send_put function, I provide a body and I provide any extra_headers I want.   Our generic build_request code then assumes that my body is JSON and adds an application/json header.   What happens when my body is actually e.g. XML?

Ideally, we would either have the caller specify the body type, or we would rename the functions to be e.g. send_json_put?   At the very least we should explicitly document that the body parameter on send_put and build_request is assumed to be JSON? (and check that this is always true everywhere they are called?)

Thoughts?

richardwhiuk:
I've added some comments to clarify it's a JSON body.

PR #693, opened 2017-10-10
Overload Integration

tomwatson1024:
Integrates multiple changes made to improve overload behavior:
- Removes the 2 second limit on recalculation of the load monitor bucket fill rate;
- Refactors the eventq class to allow different queue types.

These changes pass all but one of the existing UTs - details of the failed UT are given below (@eleanor-merry)

```
[ RUN      ] LoadMonitorTest.RequestComplete
ut/load_monitor_test.cpp:103: Failure
Value of: initial_rate
  Actual: 10
Expected: _load_monitor->bucket.rate
Which is: 19.1173
ut/load_monitor_test.cpp:123: Failure
Value of: changed_rate
  Actual: 19.1173
Expected: _load_monitor->bucket.rate
Which is: 15.9311
[  FAILED  ] LoadMonitorTest.RequestComplete (0 ms)
```


src/load_monitor.cpp
```diff
@@ -185,96 +185,93 @@ void LoadMonitor::request_complete(int latency)
...
-      {
-        // Our latency is below the threshold, so increasing our permitted request rate would be
-        // sensible. Before doing that, we check that we're using a significant proportion of our
-        // current rate - if we're allowing 100 requests/sec, and we get 1 request/sec because it's
-        // a quiet period, then it's going to be handled quickly, but that's not sufficient evidence
-        // to increase our rate.
-        int ms_passed = (current_time_ms - last_adjustment_time_ms);
-        float maximum_permitted_requests = bucket.rate * ms_passed / 1000;
-
-        // Arbitrary threshold - require 50% of our current permitted rate to be used
```
KristaHyer: Why did we take this comment out? I think we should keep it

src/load_monitor.cpp
```diff
@@ -185,96 +185,93 @@ void LoadMonitor::request_complete(int latency)
...
+                   penalties,
+                   ms_passed,
+                   accepted,
+                   rejected);
       }
       else
       {
-        TRC_DEBUG("Maximum incoming request rate/second is unchanged at %f",
-                  bucket.rate);
+        TRC_STATUS("Maximum incoming request rate/second unchanged - only handled %d requests"
```
KristaHyer: Request rate is unchanged from what? I think we should state the bucket.rate in this log (like we do in the one below it).

src/load_monitor.cpp
eleanor-merry: I made the threshold into a constant rather than a magic number, so the comment is now in the include file.

KristaHyer: I think it was a key piece of info that this threshold was arbitrary, and the comment in the include file doesn't say that.

PR #695, opened 2017-10-11
Add IO hooks

AlexHockey:
This is support code to allow us to perform actions when a thread is about to do blocking IO. This PR also contains a small enhancement to allow a stopwatch to be stopped and restarted. No UTs yet - this PR is to get early feedback on the approach before I go any further.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -515,7 +515,8 @@ namespace Utils
         if (_ok)
         {
           result_us = (now.tv_nsec - _start.tv_nsec) / 1000L +
-                      (now.tv_sec - _start.tv_sec) * 1000000L;
+                      (now.tv_sec - _start.tv_sec) * 1000000L +
+                      _elapsed_us;
         }
         else
         {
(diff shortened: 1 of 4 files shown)


include/utils.h
```diff
@@ -623,6 +624,81 @@ namespace Utils
...
+  /// The destructor removes the hook from the stack. For this reason, it is
+  /// important that instances are destroyed in the opposite order they are
+  /// created in. This means that hooks should only be stored on the stack and
+  /// not on the heap.
+  class IOHook
+  {
+  public:
+    /// Type aliases for the callbacks. Each callback takes a string which
+    /// contains the reason that the I/O has been performed.
+    using IoStartedCallback = std::function<void(const std::string& reason)>;
```
eleanor-merry: I don't care as such, but could this be IO not Io?

AlexHockey: Yeah, that makes sense.

AlexHockey:
@eleanor-merry I'm going to make this PR point at overload-integration rather than master. I'm assigning to you to merge at a suitable time.

PR #696, opened 2017-10-12
[Reviewer Richard] Clear out old_db on flush_all too

johadalin:
flush_all wasn't actually flushing all. Now it does.


johadalin:
Discussed with @richardwhiuk , and agreed it seems sensible

PR #703, opened 2017-10-31
Prevent writing of large values to memcached

MatMeredith:
Part of fixing https://github.com/Metaswitch/clearwater-issues/issues/2110.

The issue was that Astaire had a bug which meant that it failed (ungracefully) if requests grew over 64k.  This is fixed under https://github.com/Metaswitch/astaire/pull/95.  

But this bug also reveals the fact that Sprout can be made to write arbitrarily large IMPI records to memcached by firing in lots of requests that get challenged.   This PR fixes that.   Once the IMPI grows to 
 64K, subsequent requests that need to be challenges will instead be rejected with a 500 response.  (You could also hit it by trying to register lots of bindings / subscriptions -- the same behaviour will apply there too).

Not sure what the log severity should be here -- we're going to get a lot of warning log spam if a device starts sending large numbers of registers.  On the other hand we do want to know when this happens...


MatMeredith:
Finally, updated (along with https://github.com/Metaswitch/clearwater-sas-resources/pull/235) to add the new SAS log.   @sebrexmetaswitch OK?

src/memcachedstore.cpp
```diff
@@ -430,15 +430,29 @@ Store::Status TopologyNeutralMemcachedStore::set_data(const std::string& table,
   TRC_DEBUG("Writing %d bytes to table %s key %s, CAS = %ld, expiry = %d",
             data.length(), table.c_str(), key.c_str(), cas, expiry);
 
-  if (data.length() > Store::MAX_DATA_LENGTH)
+  std::string fqkey = get_fq_key(table, key);
+
+  // Check whether this request is too big.  Note that neither Astaire nor
```
sebrexmetaswitch: Minor point, but its now Rogers, not Astaire, that these requests pass through

src/memcachedstore.cpp
```diff
@@ -432,6 +432,27 @@ Store::Status TopologyNeutralMemcachedStore::set_data(const std::string& table,
...
+  // it opens us up to DoS attacks where e.g. a subscriber firing in rapid
+  // REGISTERs leads to us storing arbitrarily large values in memcached.
+  uint32_t data_length = data.length();
+  if (data_length > Store::MAX_DATA_LENGTH)
+  {
+    if (trail != 0)
+    {
+      SAS::Event err(trail, SASEvent::MEMCACHED_REQ_TOO_LARGE, 0);
+      err.add_var_param(fqkey);
+      err.add_static_param(data_length);
```
sebrexmetaswitch: One other question: I don't really know what the difference between `static` and `var` parameters in SAS is, but I'm assuming that the length gets displayed is SAS correctly?

sebrexmetaswitch: @MatMeredith not sure if you've seen this

PR #704, opened 2017-11-02
[Reviewer: AJH] Homestead diameter request error

metaswitch-ak10:
Hi Alex, would you mind reviewing this?

This is a fix for GitHub [issue #2450](https://github.com/Metaswitch/clearwater-issues/issues/2540). I have added a "catch all" SAS log for when HOOK_MESSAGE_ROUTING_ERROR occurs and the booleans 'no_peers' and 'no_connected_peers' in the code are both false. The original scenario in which the issue arose seems to be tricky to reproduce, however I have managed to get the SAS log to be displayed in the trace during SIP UA registration (playing around with iptables to simulate network problems). 
I will assign to you a related PR in clearwater-sas-resources corresponding to the actual SAS log itself.

Also, I have intentionally not added HOOK_MESSAGE_DROPPED to invoke this fd_error_hook_cb routine, since we don't know the conditions under which it might occur (e.g. what if both HOOK_MESSAGE_ROUTING_ERROR and HOOK_MESSAGE_DROPPED occur at the same time?). Please let me know if you feel I should include it.


include/sasevent.h
```diff
@@ -111,6 +111,8 @@ namespace SASEvent {
   const int HTTP_HOSTNAME_DID_NOT_RESOLVE = COMMON_BASE + 0x000018;
   const int HTTP_HOSTNAME_DID_NOT_RESOLVE_DETAIL = COMMON_BASE + 0x000019;
 
+  const int DIAMETER_REQUEST_ERROR = COMMON_BASE + 0x00001A;
```
AlexHockey: I'm not sure that this log only applies to errors. Also it would be nice if it referred to routing errors (since there are presumably other errors that can occur that we are not catching). Maybe `DIAMETER_MSG_ROUTING_ERROR`?

metaswitch-ak10: The routine which makes the log is currently being invoked only by HOOK_MESSAGE_ROUTING_ERROR. So I think calling it DIAMETER_MSG_ROUTING_ERROR instead of DIAMETER_REQUEST_ERROR is appropriate.

metaswitch-ak10:
I talked to Steve Lloyd and he thinks it is too much of a risk to implement the message dropped hook, so I will leave it for now, and it can be added later when there is a need for it (e.g. not-detailed-enough SAS traces). 
P.S. Thank you for the review!

AlexHockey:
Sounds reasonable. No problem!

PR #706, opened 2017-11-07
Added support for using X-Span-Id as a SAS correlation header

oa-metaswitch:
Quicksilver microservices are expected to use X-Span-Id for correlating SAS trails. This PR adds support for logging this in clearwater components, if present on an inbound request.

See https://github.com/Metaswitch/cpp-common-test/pull/71 for the corresponding PR to cpp-common-test

### DIFF
--- a/src/httpstack.cpp
+++ b/src/httpstack.cpp
@@ -508,6 +508,7 @@ bool HttpStack::Request::get_local_ip_port(std::string& ip, unsigned short& port
 //
 // SasLogger methods.
 //
+
 void HttpStack::SasLogger::log_correlators(SAS::TrailId trail,
                                           Request& req,
                                           uint32_t instance_id)
@@ -543,7 +544,7 @@ void HttpStack::SasLogger::log_correlator(SAS::TrailId trail,
     SAS::Marker corr_marker(trail, marker_type, instance_id);
     corr_marker.add_var_param(correlator);
 
-    // Generic correlators have a uniqueness scope. Use UUIDs for HTTP headers
+    // Generic correlators have a uniqueness scope. Use UUIDs for HTTP requests
     if (marker_type == MARKED_ID_GENERIC_CORRELATOR) {
       corr_marker.add_static_param(
         static_cast<uint32_t>(UniquenessScopes::UUID_RFC4122));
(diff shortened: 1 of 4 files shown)


src/httpstack.cpp
```diff
@@ -525,6 +526,19 @@ void HttpStack::SasLogger::log_correlator(SAS::TrailId trail,
     // long delays in the call appearing in SAS.
     SAS::report_marker(corr_marker, SAS::Marker::Scope::Trace, false);
   }
+  std::string qs_correlator = req.header(SASEvent::HTTP_SPAN_ID);
```
richardwhiuk: This code is entirely duplicating the code above - I assume you did this by copy and pasting code 😒 .

Instead:

- Rename this function `log_sas_correlators` - as it's now logging two correlators
- Create a new function, `log_sas_correlator`, which takes a header name, trail ID, request, instance ID, and marker type.
- Have `log_sas_correlators` just call `log_sas_correlator` twice.

src/httpstack.cpp
```diff
@@ -525,6 +526,19 @@ void HttpStack::SasLogger::log_correlator(SAS::TrailId trail,
...
+
+  if (qs_correlator != "")
+  {
+    SAS::Marker corr_marker(trail, MARKER_ID_VIA_BRANCH_PARAM, instance_id);
```
richardwhiuk: You should really use `MARKED_ID_GENERIC_CORRELATOR`, and obey the concerns of doing so, rather than re-using VIA_BRANCH_PARAM.

We can't just make this change for `HTTP_BRANCH_HEADER_NAME` because of upgrade concerns.

oa-metaswitch:
Hi Richard,

Thanks for the quick review and sorry about the copy/paste. I believe I've implemented your markups now, though I'm not sure I understand your concern about upgrades. Could you please clarify that for future reference and let me know if you still think something needs doing here?

richardwhiuk:
Looks good now. My point about upgrade was just 'that's why we are using the Via branch parameter' currently, and encouraging you not to head down that path.

oa-metaswitch:
thanks for the clarification. Merging now.

PR #709, opened 2017-11-09
Fix memcached timeouts

MatMeredith:
See https://github.com/Metaswitch/astaire/pull/99

### DIFF
--- a/src/memcachedstore.cpp
+++ b/src/memcachedstore.cpp
@@ -54,7 +54,7 @@ BaseMemcachedStore::BaseMemcachedStore(bool binary,
   // connect and/or write to a failed memcached replica.
   // See comment on LOCAL_MEMCACHED_CONNECTION_LATENCY_MS to understand why a
   // relatively long poll timeout is needed even for a local store.
-  _options += " --POLL-TIMEOUT=" + (remote_store) ? "250" : "100";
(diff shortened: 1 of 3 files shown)


src/memcachedstore.cpp
```diff
@@ -46,8 +47,17 @@ BaseMemcachedStore::BaseMemcachedStore(bool binary,
...
   // significant length of time.
-  _options = "--CONNECT-TIMEOUT=10 --SUPPORT-CAS --POLL-TIMEOUT=250";
+  _options = "--CONNECT-TIMEOUT=10 --SUPPORT-CAS";
+
+  // When the MemcachedStore is being used to write to memcached via Rogers,
+  // the poll-timeout needs to be long enough to accomodate Rogers failing to
+  // connect and/or write to a failed memcached replica.
+  // See comment on LOCAL_MEMCACHED_CONNECTION_LATENCY_MS to understand why a
+  // relatively long poll timeout is needed even for a local store.
+  _options += (remote_store) ? " --POLL-TIMEOUT=300" : " --POLL-TIMEOUT=100";
```
sebrexmetaswitch: I think that the 300ms timeout for a remote store is 200ms for latency plus 100ms for Rogers' own intra-site retries. If that's correct, can we add that to the comment?

PR #711, opened 2017-11-10
[Reviewer: Richard] Tweaks to StopWatch

sebrexmetaswitch:
For the fix for https://github.com/Metaswitch/homestead/pull/496, we want to be able to do some things we can't currently:

* access the HTTPRequest's StopWatch from a Task
* add and subtract times to the StopWatch

This PR adds those.

**Note** - I haven't yet checked whether we need to add some more tests in cpp-common-test, so we should do that before this is merged.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -531,6 +531,18 @@ namespace Utils
       return _ok;
     }
 
+    /// Subtracts the specified time in microseconds from the elapsed time
+    inline void subtract_time(unsigned long subtract_us)
+    {
+      _elapsed_us -= subtract_us;
+    }
+
+    /// Adds the specified time in microseconds to the elapsed time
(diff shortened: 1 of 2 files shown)


include/utils.h
```diff
@@ -531,6 +531,18 @@ namespace Utils
...
     }
 
+    /// Subtracts the specified time in microseconds from the elapsed time
+    inline void subtract_time(unsigned long subtract_us)
+    {
+      _elapsed_us -= subtract_us;
+    }
+
+    /// Adds the specified time in microseconds to the elapsed time
+    inline void add_time(unsigned long subtract_us)
```
richardwhiuk: Feels like add should be before subtract, but it's quite a minor point.

PR #712, opened 2017-11-13
[Reviewer: KH1] Add table for queue success/fail by priority

sathiyan-sivathas:
Adds a new SNMP table that covers tracks attempts/successes/failures of events put on the queue for worker threads by priority level.

### DIFF
--- a/include/sip_event_priority.h
+++ b/include/sip_event_priority.h
@@ -1,0 +1,37 @@
+/**
+ * @file sip_event_priority.h
+ *
+ * Copyright (C) Metaswitch Networks 2017
+ * If license terms are provided to you in a COPYING file in the root directory
+ * of the source code repository by which you are accessing this code, then
+ * the license outlined in that COPYING file applies to your use.
+ * Otherwise no rights are granted except for those provided to you by
+ * Metaswitch Networks in a separate written agreement.
+ */
+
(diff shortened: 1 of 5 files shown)


include/sip_event_priority.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef SIP_EVENT_PRIORITY_H
+#define SIP_EVENT_PRIORITY_H
+
+// Allowable priority levels for SIP events. Levels with higher values correspond
+// to higher priorities.
+enum SIPEventPriorityLevel
+{
+  NORMAL_PRIORITY=0,
+  HIGH_PRIORITY_1,
```
KristaHyer: I don't get why the first one of these (NORMAL_PROPRITY) is set to 0, but after that none of them are assigned values?

sathiyan-sivathas: That's just how enums work in C++.

From http://en.cppreference.com/w/cpp/language/enum

"Each enumerator is associated with a value of the underlying type. When initializers are provided in the enumerator-list, the values of enumerators are defined by those initializers. If the first enumerator does not have an initializer, the associated value is zero. For any other enumerator whose definition does not have an initializer, the associated value is the value of the previous enumerator plus one."

So technically, I didn't even need the `=0` for the first one.

PR #721, opened 2017-11-24
Make exception handling more robust

AlexHockey:
Hi Ellie, please can you review this PR that makes our exception handling behavior more robust? 

Prior to these changes, every time we hit a handled exception we:

* Dumped a core file.
* Logged basic stack trace to the log file.
* Logged advanced stack trace to the log file. 

I've changed this so that we:

* Dump a core file on the **first** handled exception. 
* Log basic stack trace to the log file. 
* Log advanced stack trace for the **first** handled exception. This comes with some restrictions:
  * It is generated from the core dumping process (which means we only get the stack from the trace that hit the exception). This was necessary to avoid stopping the sprout process while generating the trace. 
  * Advanced stack trace is now logged to stderr (which gets captured to a log file). This is necessary because now that we're logging the stack trace from another process, the stack got completely interleaved with other logs being generated from the main process, making the stack basically unreadable. 

To test this I hacked sprout to crash in the sproutlet proxy, then ran the basic call live tests (`rake test[...'] SIGNUP_CODE=... TESTS='Basic Call - *' TRANSPORT=TCP`). The sprout log and stderr files looked sensible (and are attached to the PR) and I checked that we dumped a core on the first unhandled exception (though this is often truncated when monit restarts the sprout process). 

I've also tried aborting sprout `sudo service sprout abort` and checked that I get basic stack dump to the log file, a full stack dump (including all threads ) to stderr, and a non-truncated core file.

### DIFF
--- a/src/exception_handler.cpp
+++ b/src/exception_handler.cpp
@@ -14,6 +14,7 @@
 #include <unistd.h>
 #include <cstdlib>
 #include <signal.h>
+#include <string.h>
 
 #include "exception_handler.h"
 #include "health_checker.h"
@@ -109,6 +110,9 @@ void ExceptionHandler::dump_one_core()
       // Unset the SIGABRT handler so we don't try to handle the abort call
       // below.
       signal(SIGABRT, SIG_DFL);
+
+      // We're in the child process so we can safely get advanced stack trace.
+      TRC_BACKTRACE_ADV();
 
       // Ensure the log files are complete - the core file created by abort()
       // below will trigger the log files to be copied to the diags bundle
--- a/src/exception_handler.cpp
+++ b/src/exception_handler.cpp
@@ -100,6 +100,13 @@ void ExceptionHandler::dump_one_core()
       // Unset the SIGABRT handler so we don't try to print a stack trace.
       signal(SIGABRT, SIG_DFL);
 
+      // Dump an advanced backtrace.
+      TRC_BACKTRACE_ADV();
+
+      // Ensure the log files are complete - the core file created by abort()
+      // below will trigger the log files to be copied to the diags bundle
+      TRC_COMMIT();
+
       // Now abort to generate the corefile.
       abort();
     }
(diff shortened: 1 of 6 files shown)


AlexHockey:
[sprout_err.log](https://github.com/Metaswitch/cpp-common/files/1502009/sprout_err.log)
[sprout_current.txt](https://github.com/Metaswitch/cpp-common/files/1502010/sprout_current.txt)

src/exception_handler.cpp
```diff
@@ -67,17 +66,64 @@ void ExceptionHandler::delayed_exit_thread()
 
 void* ExceptionHandler::delayed_exit_thread_func(void* det)
 {
-  // Wait for a random time up to the _ttl. This thread was detached when it 
+  // Wait for a random time up to the _ttl. This thread was detached when it
   // was created, so we can safely call sleep
   int sleep_time = rand() % ((ExceptionHandler*)det)->_ttl;
+  TRC_WARNING("Delayed exit will shutdown this process in %d seconds", sleep_time);
```
eleanor-merry: We'd generally avoided having logs in the exception handling - is this safe?

AlexHockey: Hmmm, that's annoying as it's going to make it impossible to figure out if/when a core file gets produced. Do you have any concrete concerns here?

AlexHockey: We discussed. Your concern was that we take a lock in the log function, which is not allowed in a signal handler. I think what I'm going to do here is:

* Change TRC_BACKTRACE so that it takes a format string and prints this to the log file. 
* Remove the logs from this file that are executed in the signal handler and replace them with `fprintf` calls that print to stderr. 
* Raise an issue covering the fact that it would be nice to have a `TRC_SIGNAL` macro that is safe to use in signal handlers or when you're not sure what state you are in. 

(Note that I've also done some more digging, and I've discovered that `fprintf` is also not allowed in a signal handler either (`fprintf` is AS-Unsafe according to [this](https://www.gnu.org/software/libc/manual/html_node/Formatted-Output-Functions.html)). I'll add that to the issue I am raising.

PR #724, opened 2017-11-29
[Reviewer: Sathiyan] Have all requests cost a token

eleanor-merry:
This adds a method to the load monitor to allow all requests to cost a token, but have certain requests be allowed anyway even if there aren't any tokens.

This also smoothes out the latency calculations so that the average latency is calculated from the requests that make up the 20 requests before a recalculation.

Tested in UTs, and shortly under overload.

### DIFF
--- a/src/load_monitor.cpp
+++ b/src/load_monitor.cpp
@@ -79,7 +79,7 @@ LoadMonitor::LoadMonitor(uint64_t init_target_latency_us,
           init_token_rate_s,
           init_min_token_rate_s,
           init_max_token_rate_s),
-  _smoothed_latency_us(init_target_latency_us),
+  _smoothed_latency_us(0),
   _target_latency_us(init_target_latency_us),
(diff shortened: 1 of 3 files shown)


src/load_monitor.cpp
```diff
@@ -127,11 +127,11 @@ LoadMonitor::~LoadMonitor()
...
 
-bool LoadMonitor::admit_request(SAS::TrailId trail)
+bool LoadMonitor::admit_request(SAS::TrailId trail, bool allow_anyway)
 {
   pthread_mutex_lock(&_lock);
 
-  if (_bucket.get_token())
+  if (_bucket.get_token() || allow_anyway)
   {
     // Got a token from the bucket, so admit the request.
```
sathiyan-sivathas: Can you update this comment?

eleanor-merry:
We've discussed the diags, and @sathiyan-sivathas is going to change the Sprout SAS log wording.

PR #730, opened 2017-12-07
[Reviewer RJW2] Quote utility function

steelyvoid:
Hi Richard.

This is the "quote string" utility function that I wrote alongside the CF quoting function (see https://github.com/Metaswitch/sprout/pull/2042).

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -223,6 +223,7 @@ namespace Utils
       std::string& path);
 
   std::string url_unescape(const std::string& s);
+  std::string quote_string(const std::string& s);
   std::string url_escape(const std::string& s);
 
   std::string xml_escape(const std::string& s);
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -127,6 +127,32 @@ std::string Utils::url_unescape(const std::string& s)
   return r;
 }
 
+std::string Utils::quote_string(const std::string& s)
+{
+  std::string r = "\"";
+  r.reserve((2*s.length()) + 2); // Reserve enough space to avoid continually reallocating.
+
+  for (size_t ii = 0; ii < s.length(); ++ii)
+  {
+    char unquot = s[ii];
+    switch (unquot)
+    {
+      case '"':
+      case '\\':
+        r.append(1, '\\');
+        break;
+
+      default:
+        break;
+    }
+
+    r.append(1, unquot);
+  }
+
+  r.append(1, '"');
+
+  return r;
+}
 
 std::string Utils::url_escape(const std::string& s)
 {
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -127,6 +127,8 @@ std::string Utils::url_unescape(const std::string& s)
   return r;
 }
 
+// The following function quotes strings in SIP headers as described by RFC 3261 
+// Section 25.1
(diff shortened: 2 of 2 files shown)


src/utils.cpp
```diff
@@ -127,6 +127,32 @@ std::string Utils::url_unescape(const std::string& s)
   return r;
 }
 
+std::string Utils::quote_string(const std::string& s)
```
richardwhiuk: Is it worth adding some tests to cpp-common-tests to test this function explicitly? Feels like it'd be quite easy to write some fairly comprehensive tests.

include/utils.h
```diff
@@ -223,6 +223,7 @@ namespace Utils
       std::string& path);
 
   std::string url_unescape(const std::string& s);
+  std::string quote_string(const std::string& s);
```
richardwhiuk: Worth adding a comment saying that this is to quoted as per `quoted-string` as defined by SIP RFC 3261, Section 25.1 - given every protocol has different quoting rules (e.g. in lots of places, `'` is expected to be escaped).

src/utils.cpp
```diff
@@ -127,6 +127,32 @@ std::string Utils::url_unescape(const std::string& s)
...
+{
+  std::string r = "\"";
+  r.reserve((2*s.length()) + 2); // Reserve enough space to avoid continually reallocating.
+
+  for (size_t ii = 0; ii < s.length(); ++ii)
+  {
+    char unquot = s[ii];
+    switch (unquot)
+    {
+      case '"':
```
richardwhiuk: I think that `quoted-string` technically requires 0x00 to 0x20 to be quoted, but as they are unprintable, I doubt we'll ever care.

src/utils.cpp
```diff
@@ -127,6 +127,32 @@ std::string Utils::url_unescape(const std::string& s)
   return r;
 }
 
+std::string Utils::quote_string(const std::string& s)
+{
+  std::string r = "\"";
+  r.reserve((2*s.length()) + 2); // Reserve enough space to avoid continually reallocating.
```
richardwhiuk: Slightly irritating we have to do an initial allocation, and then resize the capacity, but I can't see a way of avoiding this.

src/utils.cpp
```diff
@@ -127,6 +127,32 @@ std::string Utils::url_unescape(const std::string& s)
...
+
+  for (size_t ii = 0; ii < s.length(); ++ii)
+  {
+    char unquot = s[ii];
+    switch (unquot)
+    {
+      case '"':
+      case '\\':
+        r.append(1, '\\');
```
richardwhiuk: It's probably neater to do `r.push_back(char)` instead of `r.append(1, char)`

src/utils.cpp
steelyvoid: I did think of doing this, but noticed that we don't have tests of the adjacent url_escape functions so assumed that cpp-common-test doesn't test utils.cpp.

Happy to do this though if you disagree

steelyvoid: Scrub that - there _are_ tests of the url_escape function (I must have been searching an earlier branch of cpp-common-test in which they didn't exist).  I'll add some.

PR #733, opened 2017-12-18
Add POST method which can take extra headers on the outbound request

NigelThorpe:
Required for Bifrost Twilio integration.  Added following the same pattern elsewhere in httpclient/httpconnection.


NigelThorpe:
Took me a while to get around to checking that, and there was indeed a coverage failure.  Fixed in metaswitch/cpp-common-test#73

PR #735, opened 2018-01-04
[Reviewer: AJH] Allow format parameter for SAS logging of memcached requests

Neil-Bruce:
Add parameter to allow caller to specify the format of data being added/retrieved from memcached for use when logging to SAS.

### DIFF
--- a/src/memcachedstore.cpp
+++ b/src/memcachedstore.cpp
@@ -462,7 +462,8 @@ Store::Status TopologyNeutralMemcachedStore::set_data(const std::string& table,
                                                       uint64_t cas,
                                                       int expiry,
                                                       SAS::TrailId trail,
-                                                      bool log_body)
+                                                      bool log_body,
+                                                      Store::Format data_format)
 {
   TRC_DEBUG("Writing %d bytes to table %s key %s, CAS = %ld, expiry = %d",
             data.length(), table.c_str(), key.c_str(), cas, expiry);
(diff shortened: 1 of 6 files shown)


src/memcachedstore.cpp
```diff
@@ -379,6 +380,12 @@ Store::Status TopologyNeutralMemcachedStore::get_data(const std::string& table,
         }
 
         got_data.add_static_param(cas);
+
+        if (log_body)
+        {
+          got_data.add_static_param(data_format);
+        }
```
AlexHockey: I'd really prefer it if the code that added the parameters looked more like this:

```cpp
if (log_body)
{
  got_data.add_var_param(fqkey);
  got_data.add_var_param(data);
  got_data.add_static_param(cas);
  got_data.add_static_param(data_format);
}
else
{
  got_data.add_var_param(fqkey);
  got_data.add_static_param(cas);
}
```

This makes it much easier to check that the parameters that are being sent matches up with what the resource bundle expects and makes it more difficult to introduce bugs, or misread the code. I don't care *too* much, and I appreciate this change might be too risky, but if you are able to make it that would be great!

Neil-Bruce: Recoded this to collapse the two log_body if tests but not duplicate the fqkey and cas lines:

  got_data.add_var_param(fqkey);
  got_data.add_static_param(cas);

if (log_body)
{
  got_data.add_var_param(data);
  got_data.add_static_param(data_format);
}

AlexHockey: OK

PR #736, opened 2018-01-05
[Reviewer: RKD] Comment out removal of zero entries in the sproutConnectedHomesteadsTable table

BennettAllen1:


src/http_connection_pool.cpp
```diff
@@ -113,8 +113,13 @@ void HttpConnectionPool::decrement_statistic(AddrInfo target, CURL* conn)
     // safe to access the table
     if (_stat_table->get(ip_address)->decrement() == 0)
     {
+      // Commenting out remove below.
+      // This is a workaround to a race condition between removing zero entries
+      // and trying to access something that has been removed. This causes
+      // sprout to crash.
```
rkd-msw: I'd probably combine this with the comment below and say "If the statistic is now zero, we'd like to remove it from the table, but this causes a race condition where we remove the row while someone else is using it. See issue #2905."

rkd-msw:
Looks good, but before merging I think you should confirm it fixes the problem - i.e. by rebuilding Sprout with this change and reinstalling it on the system. https://github.com/Metaswitch/sprout/blob/dev/docs/Development.md should tell you how to do this - ask in Slack if you get stuck.

BennettAllen1:
I tested this with a rebuilt sprout and while the problem did not appear again, I cannot categorically say this removed the issue.

rkd-msw:
As mentioned in scrum, I was seeing this, but uploaded your build to all the SPNs and that fixed the problem. Go ahead and merge.

PR #737, opened 2018-01-05
[Reviewer: Richard] DnsCachedResolver should respect TTL on SOA records when we get NXDOMAIN

sebrexmetaswitch:
See https://github.com/Metaswitch/clearwater-issues/issues/2906

Previously, we would get an SOA record with an NXDOMAIN response, but ignore the TTL and just set it to our default (300s) in the cache.

This changes that behaviour to check for an SOA and use the TTL from that if we have one. It's clamped to 300s so that we'll retry after 5 mins, like we used to, but this means we can retry faster if we should be doing so.

I've live tested:
* lookups that return NXDOMAIN and an SOA get the correct TTL in the cache
* this doesn't affect other errors from the DNS server (tested by misconfiguring the signaling dns server in config, and verifying that we don't go down this code path and we get the default expiry time)

### DIFF
--- a/src/dnscachedresolver.cpp
+++ b/src/dnscachedresolver.cpp
@@ -839,7 +839,7 @@ void DnsCachedResolver::dns_response(const std::string& domain,
     {
       // NXDOMAIN, indicating that the DNS entry has been definitively removed
       // (rather than a DNS server failure). Clear the cache for this
-      // entry.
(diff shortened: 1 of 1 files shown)


sebrexmetaswitch:
I've now added UTs in the linked PR

PR #738, opened 2018-01-09
[Reviewer: BPA] Add locking around access to a ManagedTable's row map

rkd-msw:
The basic design is:

* every time we try to modify the map, we must take a lock - this ensures only one thread can do it at once
* we don't want to try to take the same lock twice - this will cause deadlock. For this reason, where we call other functions (e.g. 'get' calling 'add'), I've added a 'locked_add' function which you call if you;re already holding the lock. This is protected, so users of this class can't call it - only this class and its subclasses.

I was reluctant to add a lock here, as that might make call processing less efficient just to update some statistics - but I don't think we'll take the lock very often (only when we connect to or disconnect from a DGN) and obviously crashing is worse.

### DIFF
--- a/include/snmp_internal/snmp_table.h
+++ b/include/snmp_internal/snmp_table.h
@@ -193,40 +193,37 @@ public:
 
   void add(TRowKey key, TRow* row)
   {
-    std::pair<TRowKey, TRow*> new_entry(key, row);
-    _map.insert(new_entry);
-
-    Table<TRow>::add(row);
+    pthread_mutex_lock(&_map_lock);
+    locked_add(key, row);
+    pthread_mutex_unlock(&_map_lock);
   }
 
   // Creates the row keyed off `key`.
   void add(TRowKey key)
   {
-    TRow* row = new_row(key);
-    if (row != NULL)
-    {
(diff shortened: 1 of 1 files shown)


mirw:
Don't we use `ManagedTable` for things like tracking numbers of different response types on Cx connections (via `CxCounterTableImpl`) - it's not just when we connect/disconnect a DGN?  It seems like taking an additional lock on every access to that (and it looks to be every read, not just every add/remove) could be pretty expensive.

Not saying it's the wrong thing to do, but I think we to double-check the performance implications here.

This feels like quite an obvious issue - do you know how we expected this to work when we designed it originally?

rkd-msw:
https://github.com/Metaswitch/cpp-common/blob/3b35e9c1d6438696905016cf680055742563d53f/src/http_connection_pool.cpp#L112 suggests we expected it to be protected by a lock, but it's not very clear.

I'll double-check how widely ManagedTable is used. In the interim, I'm going to ask @BennettAllen1 to finish and merge https://github.com/Metaswitch/cpp-common/pull/736/files as that does avoid the problem without a perf hit.

PR #743, opened 2018-01-17
[Reviewer: Seb] HttpClient/Connection rework

MatMeredith:


johadalin:
Co-opting this PR :)

This covers a few large chunks of change:

* Removing all the send_<method> methods in httpclient
* Creating new HttpRequest and HttpResponse classes as the 'proper' interface to the HttpClient
* Stripping basically everything out of HttpConnection, so it is now essentially just an HttpRequest factory with a bit of stored state on server info
* Adding/Updating Mock classes for the above
* Fixing ChronosConnection up to use the new infra

I'll link the PR in cpp-common-test that covers all of this, but we have got full test coverage on all this, so i'm fairly happy it's working as expected, bar possibly a few glitches we'll find in full system wide testing.

include/http_request.h
```diff
@@ -0,0 +1,75 @@
...
+              HttpClient::RequestType method,
+              std::string path);
+
+  virtual ~HttpRequest();
+
+  // SET methods will overwrite any previous settings
+  virtual void set_req_body(std::string body);
+  virtual void set_sas_trail(SAS::TrailId trail);
+  virtual void set_allowed_host_state(int allowed_host_state);
+  virtual void set_username(std::string username); //Unclear if we ever actually use this atm, may be unnecessary
```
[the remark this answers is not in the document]
johadalin: I'm not sure if we use this anywhere in the wider codebase, so was reluctant to remove it. Do you think it would be better to remove it for now, and someone can extend it later on? or leave it in as is, but without the comment?

test_utils/mock_http_request.h
```diff
@@ -0,0 +1,36 @@
...
+#define MOCK_HTTP_REQUEST_H__
+
+#include "http_request.h"
+#include "gmock/gmock.h"
+
+class MockHttpRequest : public HttpRequest
+{
+public:
+  MockHttpRequest();
+  ~MockHttpRequest();
```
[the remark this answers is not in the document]
johadalin: @sebrexmetaswitch can we chat about this? I'm not sure why it should be virtual, and none of the other mock classes here had it as such, so if this should be, feels like we should elsewhere too

include/http_request.h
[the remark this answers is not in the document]
sebrexmetaswitch: I think if we don't use it anywhere it's best to remove it. No point in leaving dead code around to rot

include/http_request.h
```diff
@@ -0,0 +1,84 @@
...
+{
+public:
+  HttpResponse(HTTPCode return_code,
+               const std::string& resp_body,
+               const std::map<std::string, std::string>& resp_headers);
+
+  virtual ~HttpResponse();
+
+  // GET methods return empty if not set yet
+  virtual HTTPCode get_return_code();
```
sebrexmetaswitch: Personally, I've found this slightly confusing when using this code.

All the other methods are shortened (`get_resp_headers()`, `add_req_header()` etc.), so I kept trying to write `get_rc()`. Do yoy have strong opinions on the naming here?

I could definitely get used to it, but would personally find it easier if it were changed.

sebrexmetaswitch:
As discussed, I'm tempted to say that we should be changing `HTTPCode` to `HttpCode` absolutely everywhere as part of this (and the related) PR(s). It's the most annoying thing about trying to use the HTTP/Http code.

I've found myself writing:
```
HttpResponse resp = req.send();
HTTPCode rc = resp.get_return_code();
```
which looks dreadful, and I keep forgetting which one should be SHOUTED and which is regular.

include/http_request.h
richardwhiuk: I don't think Return Code is a good fit for this anyway - I think Result Code or Response Code are more accurate.

I like the compromise of `get_rc`.

include/http_request.h
```diff
@@ -0,0 +1,84 @@
...
+  virtual std::string get_resp_body();
```
richardwhiuk: Is `get_body` not sufficient here?

include/http_request.h
```diff
@@ -0,0 +1,84 @@
...
+  virtual std::map<std::string, std::string> get_resp_headers();
```
richardwhiuk: Is `get_headers` sufficient here?

include/http_request.h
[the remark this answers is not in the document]
johadalin: Found it used in xdmconnection in sprout, so leaving it in

include/http_request.h
```diff
@@ -0,0 +1,84 @@
...
+  _method(method),
+  _path(path)
+{
+}
+
+
+  virtual ~HttpRequest();
+
+  // SET methods will overwrite any previous settings
+  virtual void set_req_body(const std::string& body);
```
richardwhiuk: Ditto `set_body`

include/http_request.h
```diff
@@ -0,0 +1,84 @@
...
+  virtual void set_sas_trail(SAS::TrailId trail);
+  virtual void set_allowed_host_state(int allowed_host_state);
+  virtual void set_username(const std::string& username);
+
+  // ADD methods
+  virtual void add_req_header(const std::string& req_header);
```
richardwhiuk: Ditto `add_header`

include/httpclient.h
```diff
@@ -59,6 +59,8 @@ class HttpClient
 public:
   // HttpConnectionPool requires access to the private Recorder class
   friend class HttpConnectionPool;
+  // HttpRequest requires access to the private send_request function
```
richardwhiuk: Should we just make `send_request` public?

richardwhiuk: Although maybe keeping it private discourages people assuming that's an API

PR #746, opened 2018-01-24
[Reviewer: BR1] Adding support for A records in dns.json

RobDover:
This pull request adds the ability to add A records to `dns.json` that can then be used as a static cache in preference to DNS lookups. It involves the following changes:

- Adding a StaticDnsCache class to handle reading from file (note that this is pulled out into `static_dns_cache.cpp` and `static_dns_cache.h` respectively).
- Sharing responsibility for populating the cache between `DnsCachedResolver` and `StaticDnsCache`. The user calls the existing `reload_static_records()` function in `DnsCachedResolver`. This is responsible for handling locking, and then hands off to the `StaticDnsCache` to do the actual populating.
- Enhancing the parsing and handling of `dns.json` to allow A records.


include/static_dns_cache.h
```diff
@@ -0,0 +1,47 @@
...
+ * of the source code repository by which you are accessing this code, then
+ * the license outlined in that COPYING file applies to your use.
+ * Otherwise no rights are granted except for those provided to you by
+ * Metaswitch Networks in a separate written agreement.
+ */
+
+#ifndef STATICDNSCACHE_H__
+#define STATICDNSCACHE_H__
+
+#include <string.h>
```
br1-msw: Don't use this: https://stackoverflow.com/questions/9257665/difference-between-string-and-string-h use cstring

include/static_dns_cache.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <map>
+#include <vector>
+
+#include "dnsrrecords.h"
+
+class StaticDnsCache
+{
+public:
+  StaticDnsCache(const std::string filename = "");
```
br1-msw: Why do we initialize with a blank filename? It's always going to be the same `dns.json` right? Even if this gets extended to other DNS types that will always be the same me thinks.

include/static_dns_cache.h
```diff
@@ -0,0 +1,47 @@
...
+  ~StaticDnsCache();
+
+  // Parse the _dns_config_file.
+  void reload_static_records();
+
+  // Returns the number of records in the cache.
```
br1-msw: probably worth clarifying static cache here, not the "real" cache. I disliked the cache naming here for this reason.

src/dnscachedresolver.cpp
```diff
@@ -205,184 +199,13 @@ DnsCachedResolver::~DnsCachedResolver()
...
-    }
-  }
-  catch (JsonFormatError err)
-  {
-    TRC_ERROR("Error parsing dns config file %s.", _dns_config_file.c_str());
-    CL_DNS_FILE_MALFORMED.log();
-  }
+  pthread_mutex_lock(&_cache_lock);
+  _static_cache.reload_static_records();
+  pthread_mutex_unlock(&_cache_lock);
```
br1-msw: I just wanted to say this all looks really nice :) 👍

src/dnscachedresolver.cpp
```diff
@@ -414,34 +239,27 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
...
-      }
+      // There were some DNS records in the static cache - we use these in
+      // preference to a DNS lookup.
+      results.push_back(static_result);
+    }
+    else
+    {
+      // The static cache didn't have any records that matched, so we'll need
+      // to do a DNS lookup.
+      query_required.push_back(canonical_domain);
```
br1-msw: I have some reservations about the `canonical_domain` naming of the variable here. `get_canonical_name` doesn't always guarantee you end up with it, you might just get back your original domain. I think the interface makes more sense here if `get_canonical_name` returned `NULL` if it's not in the cache, and we checked for that in this function, that way we'd name this variable correctly either `domain` or `canonical_domain` appropriately.

src/dnscachedresolver.cpp
```diff
@@ -405,7 +228,9 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
                                   std::vector<DnsResult>& results,
                                   SAS::TrailId trail)
 {
-  std::vector<std::string> new_domains;
+  // This will contain all of the domains we need to query the cache or perform
+  // a DNS lookup for.
+  std::vector<std::string> query_required;
```
br1-msw: `query_required` sounds like it should be a boolean. I think what this variable is is domains left to query, which might mean a name like `domains_to_query` or `unresolved_domains`.

src/static_dns_cache.cpp
```diff
@@ -0,0 +1,323 @@
...
+//     {
+//       "name": <hostname>,
+//       "records": [
+//         <record objects>
+//       ]
+//     }
+//   ]
+// }
+//
+// Currently, the only supported record object is CNAME:
```
br1-msw: Not anymore!

src/static_dns_cache.cpp
```diff
@@ -0,0 +1,323 @@
+/**
+ * @file static_dns_cache.cpp Implements that ability to read DNS records from a
+ * locally configured static cache.
+ *
+ * Copyright (C) Metaswitch Networks 2017
```
br1-msw: 2018? :)

src/dnscachedresolver.cpp
```diff
@@ -414,34 +239,27 @@ void DnsCachedResolver::dns_query(const std::vector<std::string>& domains,
...
-          TRC_VERBOSE("Static CNAME record found: %s -> %s",
-                      domain.c_str(),
-                      cname_record->target().c_str());
-          new_domain = cname_record->target();
-          break;
-        }
-      }
+      // There were some DNS records in the static cache - we use these in
+      // preference to a DNS lookup.
+      results.push_back(static_result);
```
rkd-msw: I think this causes a regression - previously, `domains` and `results` were in the same order, and `results[0]` was guaranteed to be the result for `domains[0]`. That's no longer true (if `domains[1]` is in the static cache, its result will be in `results[0]`), but some code relies on it: e.g. https://github.com/Metaswitch/cpp-common/blob/master/src/diameterresolver.cpp#L104

src/dnscachedresolver.cpp
rkd-msw: I'm not convinced - I think if there is no CNAME record, the domain you look up is the canonical name, and the interface reflects that. (It's not clear what a nicer interface would be - NULL isn't a valid value for std::string, so it would need to be a std::string* or const char*, and I think that makes memory management more complex.)

PR #749, opened 2018-01-26
increase the dns timeout to a multiple of dns servers used

br1-msw:
All UTs pass fine. It's a simple change so I've not run FVs over this too.

### DIFF
--- a/include/dnscachedresolver.h
+++ b/include/dnscachedresolver.h
@@ -83,8 +83,8 @@ public:
   // Reads DNS records from _dns_config_file and stores them in _static_records
   void reload_static_records();
 
-  // Default timeout for DNS requests over the wire (in milliseconds) per DNS server
-  static const int DEFAULT_TIMEOUT = 200;
+  // Default timeout for DNS requests over the wire (in milliseconds) for a single DNS server
+  static const int DEFAULT_TIMEOUT = 600;
 
   // Default port number for DNS requests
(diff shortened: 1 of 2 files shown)


rkd-msw:
RND is right.

rkd-msw:
RND is right.

br1-msw:
Oops. Is that right though? In your email you've said 2 servers should be 200ms (not 300ms)

RobDover:
I believe that was a typo - there is a later email in the trail where @rkd-msw corrects himself.

(See the email sent 18 January 2018 21:22)

br1-msw:
Ah yes I hadn't spotted that email. - @RobDover I've now corrected the code to give the right values. Could I get a re-review?

include/dnscachedresolver.h
```diff
@@ -83,12 +83,15 @@ class DnsCachedResolver
   // Reads DNS records from _dns_config_file and stores them in _static_records
   void reload_static_records();
 
-  // Default timeout for DNS requests over the wire (in milliseconds)
-  static const int DEFAULT_TIMEOUT = 200;
+  // Default timeout for DNS requests over the wire (in milliseconds) for a single DNS server
```
RobDover: While this is correct, I don't think it encompasses the full meaning of this field any more. The field now represents the total timeout we allow across all DNS requests for a given query, and it's probably worth updating the comment here to make that clear.

PR #750, opened 2018-01-30
Specify source address for HTTP and Memcached clients

AlexHockey:
This PR adds the capability for the user of the HTTP and Memcached clients to specify what IP address the client should establish TCP connections from. This is useful as:

* It allows the FV tests to simulate realistic network conditions such as netsplits and latency. 
* It opens the way for source-based IP routing which might be useful in future. 

Tested by running a simple FV test and checking the network traffic in wireshark.

### DIFF
--- a/include/http_connection_pool.h
+++ b/include/http_connection_pool.h
@@ -52,7 +52,8 @@ public:
   HttpConnectionPool(LoadMonitor* load_monitor,
                      SNMP::IPCountTable* stat_table,
                      bool remote_connection = false,
-                     long timeout_ms = -1);
+                     long timeout_ms = -1,
+                     const std::string& source_address = "");
 
   ~HttpConnectionPool()
   {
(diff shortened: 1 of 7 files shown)


include/http_connection_pool.h
```diff
@@ -84,5 +85,23 @@ class HttpConnectionPool : public ConnectionPool<CURL*>
   // Determines an appropriate absolute HTTP request timeout in ms given the
   // target latency for requests that the downstream components will be using
   long calc_req_timeout_from_latency(int latency_us);
+
+  // Callbacks that are uses when the user has specified that connections be
```
graemerobertson: typo *used

AlexHockey:
FTR, I think the best way to test this is in clearwater-fv-test, using the netsplit functionality that you have added. You are going to run with this. 

(I haven't fixed the typo as I'd have to update loads of submodule references and I can't be arsed).

PR #752, opened 2018-02-07
[Reviewer: Richard] Fix casing in 'X-Span-ID'

alango:


richardwhiuk:
Oh, you'll also need to fix the test as well prior to merging this: https://github.com/Metaswitch/cpp-common-test/blob/7322adedcc1287f4b99c0eeaae562e07dfb0eb79/src/ut/httpstack_test.cpp#L385-L411

alango:
Not sure about a spec, but the header we're trying to integrate with is defined at https://github.com/Metaswitch/swagger-rs/blob/68cf9c54a92fcef74f2b80b99bce7ed8ca67d10b/src/lib.rs#L42
This change was made so that Weatherwax could talk to Quicksilver microservices, which no other Clearwater component does (we found this bug while testing this happening for the first time), so there should be no interop issues.
PR for the test changes at https://github.com/Metaswitch/cpp-common-test/pull/80

PR #755, opened 2018-02-09
[Reviewer: AJH] Common sas_write log callback

MF1-MS:
Creation of a new logging callback that will be compatible with all sas-clients. 

- saslogger.cpp checks to see if we need to write the received log, and converts the sasclient log level to a string.

- log.cpp writes the sas_ip and log_id if they are present. If not, all the log information will be contained in the msg component, so that will be the only arg written. 

PR for changes to the PC sas-client to interact with the new callback is [here](https://github.com/Metaswitch/sas-client/pull/83).

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -162,13 +162,13 @@ void Log::write_sas_log(const char* sas_level_str,
 
   pthread_t thread = pthread_self();
 
-  if (!log_id == NULL)
+  if (log_id != NULL)
   {
-    written = snprintf(logline, MAX_LOGLINE - 2, "[%lx] %d,", thread, sas_level_str);
+    written = snprintf(logline, MAX_LOGLINE - 2, "[%lx] %s,", thread, sas_level_str);
     written += snprintf(logline + strlen(logline), MAX_LOGLINE - 2, "%.*s,", log_id_len, log_id);
   }
 
-  if (!sap_ip == NULL)
+  if (sas_ip != NULL)
   {
     written += snprintf(logline + strlen(logline), MAX_LOGLINE - 2, "%.*s,", sas_ip_len, sas_ip);
   }
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -94,7 +94,7 @@ void Log::_write(int level, const char *module, int line_number, const char *fmt
 
   pthread_t thread = pthread_self();
 
-  written = snprintf(logline, MAX_LOGLINE -2, "[%lx] ", thread);
+  written = snprintf(logline, MAX_LOGLINE -2, "[%lx] %s ", thread, log_level[level]);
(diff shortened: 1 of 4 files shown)


AlexHockey:
I'm not that keen that you've had to make the generic logging library aware of SAS. I think you should keep `sas_write` calling into `Log::write` but make the Log ID and SAS IP address into part of the message.

src/log.cpp
```diff
@@ -92,26 +92,34 @@ void Log::_write(int level, const char *module, int line_number, const char *fmt
...
+
+  // If no module is supplied then all the information in the log is supplied in the fmt
+  // parameter, so we skip to writing fmt.
+  if (module != NULL)
   {
-    written = snprintf(logline, MAX_LOGLINE - 2, "[%lx] %s %s: ", thread, log_level[level], module);
+    const char* mod = strrchr(module, '/');
+    module = (mod != NULL) ? mod + 1 : module;
+
+    if (line_number)
```
AlexHockey: It looks like you only print out the log level if module is not NULL. That doesn't seem right. We always have a level so should be logging it.

src/saslogger.cpp
```diff
@@ -11,44 +11,98 @@
...
+               int32_t log_id_len,
+               unsigned char* log_id,
+               int32_t sas_ip_len,
+               unsigned char* sas_ip,
+               int32_t msg_len,
+               unsigned char* msg)
 {
   int level;
-  va_list args;
+  std::string sas_level_str;
```
AlexHockey: Why do you need this. You should be letting `Log::_write` print out the level.

src/saslogger.cpp
```diff
@@ -11,44 +11,98 @@
...
+
+  // Set level to 0, so that the log level check in Log::write passes. Pass in an empty va_list
+  // to Log::_write, as we don't need to format our msg any further.
+  level = 0;
+  va_list empty_va_list;
+  Log::_write(level,
+              NULL,
+              0,
+              logline,
+              empty_va_list);
```
AlexHockey: Instead of having any of this code, why don't you just do

```cpp
Log::write(level, nullptr, 0, "%.*s %.*s %.*s", log_id_len, log_id, sas_ip_len, sas_ip, msg_len, msg);
```

MF1-MS: I still envision I need a check in Log::write() that we have a module, and print out only the level and msg if not? 

I hadn't thought of that line you've suggested, that's much nicer.

AlexHockey: Sorry I should have been more specific - I mean that most of the new code in saslogger.cpp is not needed. I agree you need the changes in the other files you've changed.

MF1-MS:
Made the markups requested, would you just have one more look please?

PR #761, opened 2018-02-15
Standardise ordering of DNS records when static cache is involved

RobDover:
This code makes sure that the results of `DnsCachedResolver::dns_query()` come through in the order that the queries were sent in. I've tested that this passes the test written for this case in cpp-common-test.


RobDover:
I spent a while trying to come up with a neater way to do this (because having multiple maps feels clunky), but I couldn't come up with an obvious improvement. Feel free to suggest one!

PR #765, opened 2018-02-23
Extra Diags and static dns cache from EANTC work

BennettAllen1:
Combination of changes in dns_static_cache and the apple_scrumble_master branch

This pull request adds the ability to add A records to dns.json that can then be used as a static cache in preference to DNS lookups. It involves the following changes:
Adding a StaticDnsCache class to handle reading from file (note that this is pulled out into static_dns_cache.cpp and static_dns_cache.h respectively).
Sharing responsibility for populating the cache between DnsCachedResolver and StaticDnsCache. The user calls the existing reload_static_records() function in DnsCachedResolver. This is responsible for handling locking, and then hands off to the StaticDnsCache to do the actual populating.
Enhancing the parsing and handling of dns.json to allow A records

This also includes extra diags from the EANTC work done by apple scrumble


src/dnscachedresolver.cpp
```diff
@@ -1112,10 +982,7 @@ DnsCachedResolver::DnsChannel* DnsCachedResolver::get_dns_channel()
     channel->resolver = this;
     struct ares_options options;
 
-    // ARES_FLAG_STAYOPEN implements TCP keepalive - it doesn't do
-    // anything obviously helpful for UDP connections to the DNS server,
-    // but it's what we've always tested with so not worth the risk of removing.
-    options.flags = ARES_FLAG_STAYOPEN;
+    options.flags = ARES_FLAG_USEVC;
```
BennettAllen1: Could this cause a regression?

include/memcached_connection_pool.h
```diff
@@ -31,8 +31,8 @@
 /// will block for three times the connect latency (which is one of the following
 /// two values): once when trying to create the connection, and then twice
 /// trying to use it (because libmemcached doesn't pass back the error).
-static int LOCAL_MEMCACHED_CONNECTION_LATENCY_MS = 25;
-static int REMOTE_MEMCACHED_CONNECTION_LATENCY_MS = 250;
+static int LOCAL_SITE_MEMCACHED_CONNECTION_LATENCY_MS = 30;
+static int REMOTE_SITE_MEMCACHED_CONNECTION_LATENCY_MS = 250;
```
BennettAllen1: Are we happy changing the default value in master?

PR #768, opened 2018-03-02
[Reviewer: SR2] Turns down logs which were too high

BennettAllen1:
This turns down a couple of per log calls which were at too high a level


sebrexmetaswitch:
LGTM 👍

PR #769, opened 2018-03-02
[Reviewer: Ellie] Diameter peer connection alarm

metaswitch-ak10:
This is an accompanying PR to https://github.com/Metaswitch/homestead/pull/512.


include/communicationmonitor.h
```diff
@@ -66,7 +67,6 @@ class CommunicationMonitor : public BaseCommunicationMonitor
 
 private:
   virtual void track_communication_changes(unsigned long now_ms);
-  unsigned long current_time_ms();
```
metaswitch-ak10: I have moved this routine to utils since I use it in realmmanager.cpp.

src/realmmanager.cpp
```diff
@@ -98,6 +105,70 @@ void RealmManager::stop()
...
+          ii != _peers.end();
+          ii++)
+    {
+      if ((ii->second)->connected())
+      {
+        num_connected_peers++;
+      }
+    }
+
+    const bool clear_alarm = (num_connected_peers >= _max_peers) ||
```
eleanor-merry: This check doesn't strictly match what the alarm text says (the clear and minor text is inconsistent). What's the correct behaviour?

src/realmmanager.cpp
```diff
@@ -413,5 +501,20 @@ void RealmManager::manage_connections(int& ttl)
...
   _peers = locked_peers;
+
+  // 8. Remove old _failed_peers.
+  const unsigned long current_time = Utils::current_time_ms();
+  for (std::map<AddrInfo, const unsigned long>::iterator ii = _failed_peers.begin();
+                                                         ii != _failed_peers.end();)
+  {
+    if (ii->second + _failed_peers_timeout_ms <= current_time)
+    {
+      _failed_peers.erase(ii++);
```
eleanor-merry: Is this safe? Erasing from something while iterating over it normally isn't!

src/realmmanager.cpp
```diff
@@ -118,21 +189,29 @@ void RealmManager::peer_connection_cb(bool connection_success,
                  realm.c_str());
         pthread_rwlock_unlock(&_peers_lock);
         pthread_rwlock_wrlock(&_peers_lock);
+        const int was_failed = _failed_peers.erase(peer->addr_info());
```
eleanor-merry: Hmm. This isn't very pleasant. I don't like how you're using an int and bool fairly interchangeably, and I don't like that you're using the rc of erase and insert to determine whether the failed peers have changed; I don't remember what the exact rcs are for these functions without looking it up - and I shouldn't have to just to understand this function. I think you should change this so:
* manage_alarm takes a bool, as that's what you're actually using the variable as. 
* there are two helper functions, remove_from_failed_peers and add_to_failed_peers, which modify the peer list and return if the peer was in the list or not (using the same logic as you've got here, but not breaking up the flow of this function).
* the bool was/nt_already_failed isn't defined as const - you have no need to do this.

src/realmmanager.cpp
metaswitch-ak10: It is safe for the case of a map:
* "References and iterators to the erased elements are invalidated. Other references and iterators are not affected." - http://en.cppreference.com/w/cpp/container/map/erase
* "Post-increment and post-decrement creates a copy of the object, increments or decrements the value of the object and returns the copy from before the increment or decrement." - http://en.cppreference.com/w/cpp/language/operator_incdec

src/realmmanager.cpp
```diff
@@ -44,12 +44,18 @@ RealmManager::RealmManager(Diameter::Stack* stack,
                            std::string realm,
                            std::string host,
                            int max_peers,
-                           DiameterResolver* resolver) :
+                           DiameterResolver* resolver,
+                           Alarm* alarm,
```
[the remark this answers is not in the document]
metaswitch-ak10: Fixed

metaswitch-ak10:
@eleanor-merry Sorry Ellie, I have accidentally based my fixed branch on the dev branch instead of master (for some reason there is a dev branch in cpp-common). I am closing this PR and opening a clean one with my branch cut from master.

PR #772, opened 2018-03-06
Additional Ram Recorder enhancements

kenferguson:


kenferguson:
Not taking this because we're now moving towards optional support for full logging so need this code back in.

PR #773, opened 2018-03-07
Add support for RAM recorder

MatMeredith:
Most of the change here is the corer RAM recorder functionality that we took from L3 previously.   The only code I've touched/added are the TRC... macros and the new enableRamRecordEverything function.


src/log.cpp
```diff
@@ -10,23 +10,823 @@
...
+    fprintf(output, "!!!Run off the end of the RAM trace buffer.  Trace corrupted!!!\n");
+    return;
+  }
+}
+
+void Log::dumpRamRecorder(std::string output_dir)
+{
+  // Dump out the RAM trace buffer
+  char ramname[64];
+  std::string filename_template = output_dir + "/" + "ramtrace.%ld.txt";
```
richardwhiuk: Could just do:

`std::string filename_template = output_dir + "/ramtrace." + std::to_string(time(NULL)) + ".txt";`

src/log.cpp
```diff
@@ -10,23 +10,823 @@
...
+  sprintf(ramname, filename_template.c_str(), time(NULL));
+  FILE *ramtrace = fopen(ramname, "w");
+  Log::ramDecode(ramtrace);
```
richardwhiuk: Should check if we managed to open `ramname` (i.e. if `fopen` returns `NULL`).

richardwhiuk: Should we merge ramDecode and dumpRamRecorder? Do we ever decode the ram buffer without dumping it?

src/log.cpp
```diff
@@ -10,23 +10,812 @@
...
 #include <string.h>
+#include <sys/time.h>
+#include <time.h>
+
+// Define the "Warn unused return" macro (to be nothing) as this macro is
+// needed by printf.h and is apparently undefined under -O2 optimisation
+//#define __wur
+#undef __wur
+#define __wur
+#include <printf.h>
```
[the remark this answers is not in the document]
kenferguson: That's partly my fault - I added part of that code when trying to get the ramrecorder code to compile (although the comment was Steve Lloyd's) and totally forgot about it .

src/log.cpp
kenferguson: (Just skimming through looking at the comments because I was curious)
I discussed this point with Steve and we felt they were conceptually separate. While at the moment we want to do both together, it felt perfectly reasonable to have an API where you requested the current RAM recorder contents and then returned them to the caller not just to disk. Feels reasonable to keep them separate to me.

include/log.h
```diff
@@ -47,6 +66,7 @@ namespace Log
 #endif
   }
   void setLoggingLevel(int level);
+  void enableRamRecordEverything();
```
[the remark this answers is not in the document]
MatMeredith: Maybe?  Not sure that I am convinced enough to change it though?   It's consistent with the config option too.

MatMeredith:
Thanks for comments.   I think I've applied the must fix markups outside of L3's code.

PR #774, opened 2018-03-07
[Reviewer: Ellie] Diameter peer connection alarm

metaswitch-ak10:
Hi Ellie.

Here is the PR for the diameter peer connection alarm we have discussed before. Would you mind reviewing it please?
As we agreed, the alarm gets raised directly from the realm manager without using the communication monitor. For the alarm mechanics see the block comment in realmmanager.h.

There are related PRs in homestead and ralf referenced bellow that implement the process specific details of the alarm.

I have also added some UTs to test my fix - see PR in cpp-common-test referenced bellow.

I have tested my fix by running make full_test in cpp-common-test, homestead and in ralf. I also did a live test to check the behaviour of the ENT logs is as expected.

### DIFF
--- a/include/realmmanager.h
+++ b/include/realmmanager.h
@@ -27,8 +27,8 @@ public:
                int max_peers,
                DiameterResolver* resolver,
                Alarm* alarm,
-               const PDLog* ALARM_CLEAR_LOG,
-               const PDLog1<const char*>* ALARM_ERROR_LOG);
+               const PDLog& peer_comm_restored_log,
+               const PDLog1<const char*>& peer_comm_error_log);
   virtual ~RealmManager();
 
   void start();
@@ -55,8 +55,8 @@ private:
   
   // Helper functions that modify the _failed_peers map. Both return true if
   // modified, false otherwise.
-  bool add_to_failed_peers(Diameter::Peer* peer);
-  bool remove_from_failed_peers(Diameter::Peer* peer);
+  bool try_add_to_failed_peers(Diameter::Peer* peer);
+  bool try_remove_from_failed_peers(Diameter::Peer* peer);
 
   void remove_old_failed_peers(unsigned long now_ms = 0);
 
@@ -82,9 +82,9 @@ private:
   DiameterResolver* _resolver;
   Alarm* _peer_connection_alarm;
   std::map<std::string, Diameter::Peer*> _peers;
-  std::map<AddrInfo, const unsigned long> _failed_peers;
-  const PDLog* _ALARM_CLEAR_LOG;
-  const PDLog1<const char*>* _ALARM_ERROR_LOG;
+  std::map<AddrInfo, unsigned long> _failed_peers;
+  const PDLog& _peer_comm_restored_log;
+  const PDLog1<const char*>& _peer_comm_error_log;
   volatile bool _terminating;
 };
 
--- a/include/realmmanager.h
+++ b/include/realmmanager.h
@@ -55,10 +55,10 @@ private:
   
   // Helper functions that modify the _failed_peers map. Both return true if
   // modified, false otherwise.
-  bool try_add_to_failed_peers(Diameter::Peer* peer);
-  bool try_remove_from_failed_peers(Diameter::Peer* peer);
+  bool add_to_failed_peers(Diameter::Peer* peer);
+  bool remove_from_failed_peers(Diameter::Peer* peer);
 
-  void remove_old_failed_peers(unsigned long now_ms = 0);
+  void remove_old_failed_peers();
 
   // We use a read/write lock to read and update the _peers map (defined below).
   // However, we read this map on every single Diameter message, so we want to
--- a/include/realmmanager.h
+++ b/include/realmmanager.h
@@ -23,7 +23,7 @@ public:
   RealmManager(Diameter::Stack* stack,
                std::string realm,
                std::string host,
-               unsigned int max_peers,
+               int max_peers,
                DiameterResolver* resolver,
                Alarm* alarm=NULL,
                std::string sender="",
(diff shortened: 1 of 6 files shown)


include/realmmanager.h
```diff
@@ -64,7 +80,11 @@ class RealmManager
   pthread_t _thread;
   pthread_cond_t _cond;
   DiameterResolver* _resolver;
+  Alarm* _peer_connection_alarm;
   std::map<std::string, Diameter::Peer*> _peers;
+  std::map<AddrInfo, const unsigned long> _failed_peers;
```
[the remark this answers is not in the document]
metaswitch-ak10: You are right. I think I should update the timestamp each time we fail to connect to the peer.

src/realmmanager.cpp
```diff
@@ -73,6 +79,110 @@ void RealmManager::stop()
...
+  {
+    failed_peers_string += sep + Utils::ip_addr_to_arpa(failed_peer.first.address);
+    sep = ", ";
+  }
+  return failed_peers_string;
+}
+
+bool RealmManager::add_to_failed_peers(Diameter::Peer* peer)
+{
+  // map::insert returns true if inserted, false otherwise
```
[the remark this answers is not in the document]
metaswitch-ak10: Would something like this be better?

// This function adds a peer to _failed_peers.
// Returns true if the peer wasn't failed before, otherwise returns false.
bool RealmManager::add_to_failed_peers(Diameter::Peer* peer)
{
  std::map<AddrInfo, unsigned long>::iterator it;
  it = _failed_peers.find(peer->addr_info());

  if (it == map::end)
  {
    // peer wansn't failed before so add it to _failed_peers
    _failed_peers.insert(std::pair<AddrInfo, unsigned long>(peer->addr_info(), Utils::current_time_ms()));
    return true;
  }
  else
  {
    // peer was already failed, update the timestamp
    it->second = Utils::current_time_ms();
    return false;
  }
}

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+    const bool clear_alarm = (num_connected_peers >= _max_peers) ||
+                              _failed_peers.empty();
+
+    if (_peer_connection_alarm->get_alarm_state() == AlarmState::ALARMED)
+    {
+      // Alarm is raised so consider clearing it
+      if (clear_alarm)
+      {
+        _peer_connection_alarm->clear();
+        _peer_comm_restored_log.log();
```
eleanor-merry: Why do you only make the log if the alarm is passed in?

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+  {
+    // peer was failed so remove it
+    _failed_peers.erase(it);
+    return true;
+  }
+}
+
+void RealmManager::remove_old_failed_peers(unsigned long now_ms)
+{
+  now_ms = now_ms ? now_ms : Utils::current_time_ms();
```
eleanor-merry: This is the same issue as I've commented on before - where you're using a variable as a different type. Yes this compiles, but it's incredibly confusing and it's a massive trap for the next person to edit this code.

In this case, I don't think you want to pass in the time to this function at all - it should always use the current time. I assume you done this for the UTs, but you can test this insetead by using cw_control_time instead.

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+
+// This function tries to remove a peer from _failed_peers. Returns a bool
+// indicating whether the peer is successfully removed (i.e. whether the peer
+// was failed already)
+bool RealmManager::try_remove_from_failed_peers(Diameter::Peer* peer)
+{
+  std::map<AddrInfo, unsigned long>::iterator it;
+  it = _failed_peers.find(peer->addr_info());
+
+  if (it == _failed_peers.end())
```
eleanor-merry: Replace it with _failed_peers.find(peer->addr_info())

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+  {
+    // peer wasn't failed before so nothing to remove
```
eleanor-merry: Can you capitalise your comments?

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+  }
+  else
+  {
+    // peer was already failed, update the timestamp
+    it->second = Utils::current_time_ms();
+    return false;
+  }
+}
+
+// This function tries to remove a peer from _failed_peers. Returns a bool
```
eleanor-merry: This function ensure that the peer is not in _failed_peers list, removing it if necessary. It returns whether the peer was in the list at the start.

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+// indicating whether the peer is successfully removed (i.e. whether the peer
+// was failed already)
+bool RealmManager::try_remove_from_failed_peers(Diameter::Peer* peer)
```
eleanor-merry: I don't think you want the try_ in this function name. To me, that implies that this function can fail.

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+  for (std::pair<AddrInfo, unsigned long> failed_peer : _failed_peers)
+  {
+    failed_peers_string += sep + Utils::ip_addr_to_arpa(failed_peer.first.address);
+    sep = ", ";
+  }
+
+  return failed_peers_string;
+}
+
+// This function tries to add a peer to _failed_peers. Returns a bool
```
eleanor-merry: This function adds a new peer to the _failed_peers map or, if the peer is already failed, updates the peer in the map. Returns whether the peer was in the map already.

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+// indicating whether the peer was successfully added (i.e. whether the peer
+// wasn't failed before)
+bool RealmManager::try_add_to_failed_peers(Diameter::Peer* peer)
```
eleanor-merry: Same comment as to the name as below.

src/realmmanager.cpp
```diff
@@ -73,6 +79,144 @@ void RealmManager::stop()
...
+{
+  std::map<AddrInfo, unsigned long>::iterator it;
+  it = _failed_peers.find(peer->addr_info());
+
+  if (it == _failed_peers.end())
```
eleanor-merry: Replace it with _failed_peers.find(peer->addr_info())

src/realmmanager.cpp
metaswitch-ak10: Yes, that was the reason, I'll have a look at using cw_control_time instead. Thank you.

src/realmmanager.cpp
metaswitch-ak10: I think "Returns whether the peer was in the map already" is a bit misleading. Shouldn't it be "Returns false if the peer was in the map already, true otherwise"? Or should I reverse the return values instead? However, I think it should return true if the peer was successfully added.

src/realmmanager.cpp
metaswitch-ak10: I think the ENT logs are supposed to provide more detail on the alarm. I think that if anyone is interested in monitoring peer communication failures, they would want to have the alarm as well. But if you think otherwise, I am willing to change it.

include/realmmanager.h
```diff
@@ -64,7 +80,11 @@ class RealmManager
   pthread_t _thread;
   pthread_cond_t _cond;
   DiameterResolver* _resolver;
+  Alarm& _peer_connection_alarm;
   std::map<std::string, Diameter::Peer*> _peers;
+  std::map<AddrInfo, unsigned long> _failed_peers;
+  const PDLog& _peer_comm_restored_log;
```
eleanor-merry: Why are these const? (I mentioned this in one of my previous comments)

src/realmmanager.cpp
```diff
@@ -73,6 +79,134 @@ void RealmManager::stop()
...
+
+  return failed_peers_string;
+}
+
+// This function adds a new peer to the _failed_peers map or, if the peer is
+// already failed, updates the peer in the map. Returns false if the peer was
+// in the map already, otherwise returns true.
+bool RealmManager::add_to_failed_peers(Diameter::Peer* peer)
+{
+  if (_failed_peers.find(peer->addr_info()) == _failed_peers.end())
```
eleanor-merry: Can you comment for all these functions that they're only safe to call if you're holding the peers lock

include/realmmanager.h
metaswitch-ak10: @eleanor-merry Since I am passing the PDLog structures by reference and they are declared as const in homestead_pd_definitions.h, I need the member variables to be const as well. Should I change it so that the PDLog structures are passed in by value instead?

metaswitch-ak10:
This PR was replaced by https://git.datcon.co.uk/clearwater-core/cpp-common/merge_requests/88.

PR #776, opened 2018-03-08
Naive RAM Recorder

richardwhiuk:
UTs are in https://github.com/Metaswitch/cpp-common-test/pull/88

This adds a naive RAM buffer which we store logs in from the SIP Stack and Diameter stack (and configurable from everywhere).

### DIFF
--- a/include/log.h
+++ b/include/log.h
@@ -72,6 +72,7 @@ namespace RamRecorder
 {
   extern bool record_everything;
   void recordEverything();
+  void _record(int level, const char* module, int lineno, const char* context, const char* format, va_list args);
   void record(int level, const char* module, int lineno, const char* format, ...);
   void record_with_context(int level, const char* module, int lineno, char* context, const char* format, ...);
   void write(const char* buffer, size_t length);
--- a/include/log.h
+++ b/include/log.h
@@ -75,6 +75,7 @@ namespace RamRecorder
   void _record(int level, const char* module, int lineno, const char* context, const char* format, va_list args);
   void record(int level, const char* module, int lineno, const char* format, ...);
   void record_with_context(int level, const char* module, int lineno, char* context, const char* format, ...);
+  void reset();
   void write(const char* buffer, size_t length);
   void dump(const std::string& output_dir);
 }
--- a/include/log.h
+++ b/include/log.h
@@ -70,16 +70,11 @@ namespace Log
 
 namespace RamRecorder
 {
-  bool record_everything = false;
-
-  void recordEverything()
-  {
-    record_everything = true;
-  }
-
+  extern bool record_everything;
+  void recordEverything();
   void record(int level, const char* module, int lineno, const char* format, ...);
-  void write(const char* buffer, int length);
-  void dump(std::string output_dir);
+  void write(const char* buffer, size_t length);
+  void dump(const std::string& output_dir);
 }
 
 #endif
--- a/include/log.h
+++ b/include/log.h
@@ -73,6 +73,7 @@ namespace RamRecorder
   extern bool record_everything;
   void recordEverything();
   void record(int level, const char* module, int lineno, const char* format, ...);
+  void record_with_context(int level, const char* module, int lineno, char* context, const char* format, ...);
   void write(const char* buffer, size_t length);
   void dump(const std::string& output_dir);
 }
--- a/include/log.h
+++ b/include/log.h
@@ -16,12 +16,25 @@
 #include "logger.h"
 #include <cstdarg>
 
-#define TRC_ERROR(...) if (Log::enabled(Log::ERROR_LEVEL)) Log::write(Log::ERROR_LEVEL, __FILE__, __LINE__, __VA_ARGS__)
-#define TRC_WARNING(...) if (Log::enabled(Log::WARNING_LEVEL)) Log::write(Log::WARNING_LEVEL, __FILE__, __LINE__, __VA_ARGS__)
-#define TRC_STATUS(...) if (Log::enabled(Log::STATUS_LEVEL)) Log::write(Log::STATUS_LEVEL, __FILE__, __LINE__, __VA_ARGS__)
-#define TRC_INFO(...) if (Log::enabled(Log::INFO_LEVEL)) Log::write(Log::INFO_LEVEL, __FILE__, __LINE__, __VA_ARGS__)
-#define TRC_VERBOSE(...) if (Log::enabled(Log::VERBOSE_LEVEL)) Log::write(Log::VERBOSE_LEVEL, __FILE__, __LINE__, __VA_ARGS__)
-#define TRC_DEBUG(...) if (Log::enabled(Log::DEBUG_LEVEL)) Log::write(Log::DEBUG_LEVEL, __FILE__, __LINE__, __VA_ARGS__)
+#define TRC_RAMTRACE(level, ...) RamRecorder::record(level, __FILE__, __LINE__, ##__VA_ARGS__)
+
+#define TRC_MAYBE_RAMTRACE(...)                                            \
+do {                                                                       \
+  if (RamRecorder::record_everything) {                                    \
+    TRC_RAMTRACE(__VA_ARGS__);                                             \
+  }                                                                        \
+} while (0)
+
+#define TRC_LOG(level, ...) if (Log::enabled(level)) Log::write(level, __FILE__, __LINE__, ##__VA_ARGS__)
+#define TRC_BASE(level, ...) do { TRC_MAYBE_RAMTRACE(level, ##__VA_ARGS__); TRC_LOG(level, ##__VA_ARGS__); } while (0)
+
+#define TRC_ERROR(...) TRC_BASE(Log::ERROR_LEVEL, ##__VA_ARGS__)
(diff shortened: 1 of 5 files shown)


include/log.h
```diff
@@ -55,4 +68,18 @@ namespace Log
...
+namespace RamRecorder
+{
+  bool record_everything = false;
+
+  void recordEverything()
+  {
+    record_everything = true;
+  }
+
+  void record(int level, const char* module, int lineno, const char* format, ...);
```
kenferguson: I really don't like the "record" and "write" functions because they sound, without further investigation like identical functions.

src/log.cpp
```diff
@@ -191,3 +209,136 @@ void Log::commit()
...
+  va_end(args);
+
+  int written;
+  int truncated;
+  char logline[MAXLOGLINE];
+
+  // Fill out the log line
+  log_helper(logline, written, truncated, level, module, lineno, format, args);
+
+  RamRecorder::record(logline, written);
```
kenferguson: I can't find this function anywhere in the review.

src/log.cpp
```diff
@@ -191,3 +209,136 @@ void Log::commit()
...
+void RamRecorder::record(int level, const char* module, int lineno, const char* format, ...)
+{
+  va_list args;
+  va_start(args, format);
+  RamRecorder::write(format, args);
+  va_end(args);
+
+  int written;
+  int truncated;
+  char logline[MAXLOGLINE];
```
kenferguson: Does this compile? I thought it was MAX_LOGLINE.

src/log.cpp
```diff
@@ -191,3 +209,136 @@ void Log::commit()
...
+}
+
+void RamRecorder::write(const char* buffer, int length)
+{
+  char* end = ram_buffer + RAM_BUFFER_SIZE;
+
+  while (length > 0)
+  {
+    size_t left = end - ram_buffer_end;
+    size_t write = std::min(length, left);
```
kenferguson: feels like write is a bad choice of variable name in a function called write.

src/log.cpp
```diff
@@ -191,3 +209,136 @@ void Log::commit()
...
+
+  // Fill out the log line
+  log_helper(logline, written, truncated, level, module, lineno, format, args);
+
+  RamRecorder::record(logline, written);
+
+  if (truncated)
+  {
+    char buf[128];
+    int len = snprintf(buf, 128, "Previous log was truncated by %d characters\n", truncated);
```
kenferguson: It might not be the previous log that's truncated - in principle, other threads could have jumped in before us. Is it worth saying "earlier log"? In general, it's probably reasonably obvious from the traceline unless you're unlucky. so I'll leave it up to you whether you think it's important enough.

src/log.cpp
kenferguson: For the cost of losing up to 8k, I'd have been tempted to just wrap if we were going to write more than the remaining buffer.

src/log.cpp
```diff
@@ -191,3 +209,136 @@ void Log::commit()
...
+    // Write until the end of the buffer
+    memcpy(ram_buffer_end, buffer, write);
+
+    // We've got less to write now
+    length -= write;
+
+    // We've also got less in the buffer to write
+    buffer += write;
+
+    if (write == left)
```
kenferguson: Please tell me that this was deliberate.

It would only be better if you'd written

 if ((3 * write) == left)

src/log.cpp
```diff
@@ -191,3 +209,136 @@ void Log::commit()
...
+
+  FILE *file = fopen(ram_file_name.c_str(), "w");
+
+  if (file)
+  {
+    fprintf(file, "RAM BUFFER\n==========\n");
+
+    if (ram_buffer_end == ram_buffer_start)
+    {
+      // No bufffered data
```
kenferguson: Feels like we should write something here to make it clear that there were no recorded logs. Otherwise you might worry that it had been truncated.

src/log.cpp
richardwhiuk: It's actually not that much less complex - we end up having to track the same number of things if we want the logs to be well-ordered.

src/log.cpp
richardwhiuk: Sadly not.

src/log.cpp
richardwhiuk: It's almost certainly an error case given we log on startup of the SIP and Diameter stacks.

src/diameterstack.cpp
```diff
@@ -694,8 +694,11 @@ void Stack::logger(int fd_log_level, const char* fmt, va_list args)
       log_level = Log::DEBUG_LEVEL;
       break;
   }
+
+  va_list argsb;
+  va_copy(argsb, args);
   Log::_write(log_level, "freeDiameter", 0, fmt, args);
-  RamRecorder::record(log_level, "freeDiameter", 0, fmt, args);
+  RamRecorder::_record(log_level, "freeDiameter", 0, nullptr, fmt, argsb);
```
kenferguson: Should you not have a call to va_end(argsb)?

(based on https://linux.die.net/man/3/va_copy)

src/log.cpp
```diff
@@ -213,133 +228,158 @@ void Log::commit()
...
+  bool record_everything = false;
+}
 
 void RamRecorder::recordEverything()
 {
   RamRecorder::record_everything = true;
 }
 
-void RamRecorder::record(int level, const char* module, int lineno, const char* format, ...)
+void RamRecorder::_record(int level, const char* module, int lineno, const char* context, const char* format, va_list args)
```
kenferguson: I think we need a comment explaining what context is somewhere - I couldn't find anything.

src/log.cpp
```diff
@@ -247,12 +248,26 @@ void RamRecorder::_record(int level, const char* module, int lineno, const char*
...
-  // Fill out the log line
-  log_helper(logline, written, truncated, level, module, lineno, context, format, args);
+  {
+    char timestamp[100];
+    timestamp_t ts;
+    struct timespec timespec;
+    clock_gettime(CLOCK_REALTIME, &timespec);
+    Logger::get_timestamp(ts, timespec);
+    Logger::format_timestamp(ts, timestamp, sizeof(timestamp));
+    RamRecorder::write(timestamp, strlen(timestamp));
```
kenferguson: This has threading issues - each time we write some data we lock but that can mean data is split up.

kenferguson: (would it be better to just timestamp EVERY log. Even if it's "start of file" "end of file"?

include/log.h
```diff
@@ -16,12 +16,25 @@
...
+#define TRC_RAMTRACE(level, ...) RamRecorder::record(level, __FILE__, __LINE__, ##__VA_ARGS__)
+
+#define TRC_MAYBE_RAMTRACE(...)                                            \
+do {                                                                       \
+  if (RamRecorder::record_everything) {                                    \
+    TRC_RAMTRACE(__VA_ARGS__);                                             \
+  }                                                                        \
+} while (0)
+
+#define TRC_LOG(level, ...) if (Log::enabled(level)) Log::write(level, __FILE__, __LINE__, ##__VA_ARGS__)
```
MatMeredith: This is late and not essential, but why isn't this TRC_MAYBE_LOG for consistency with TRC_MAYBE_RAMTRACE?

include/log.h
```diff
@@ -16,12 +16,25 @@
...
+#define TRC_BASE(level, ...) do { TRC_MAYBE_RAMTRACE(level, ##__VA_ARGS__); TRC_LOG(level, ##__VA_ARGS__); } while (0)
```
MatMeredith: Otherwise I think I'd assume from this line that we maybe ramtrace and definitely log...

MatMeredith: (I mean actually I wouldn't, but that's how it reads ;-)

src/log.cpp
```diff
@@ -191,3 +225,193 @@ void Log::commit()
...
+}
+
+void RamRecorder::write(const char* message, size_t length)
+{
+  pthread_mutex_lock(&RamRecorder::lock);
+  const char* end = buffer + RAM_BUFFER_SIZE;
+
+  while (length > 0)
+  {
+    size_t left = end - buffer_end;
```
MatMeredith: I find the names of the variables in this function _incredibly_ confusing...
- buffer_end is surely a pointer to the end of the buffer, but I think it's actually the current insert point?   `end - buffer_end` isn't at all obvious.  And `left` is just very ambiguous...   How about `buffer_space_left = buffer_end - current_insert_point`

src/log.cpp
```diff
@@ -191,3 +225,193 @@ void Log::commit()
...
+    size_t write = std::min(length, left);
+
+    // Write until the end of the buffer
+    memcpy(buffer_end, message, write);
```
MatMeredith: Write until the end of the buffer, but we start writing at buffer_end!?

src/log.cpp
```diff
@@ -191,3 +225,193 @@ void Log::commit()
...
+
+    // We've got less to write now
+    length -= write;
+
+    // We've also got less in the message buffer to write
+    message += write;
+
+    if (write == left)
```
MatMeredith: Sorry, but I'd need to do a search and replace of variable names to something clearer if I wanted to actually review the logic here...

PR #778, opened 2018-03-15
[Reviewer: Sathiyan] Remove the cache entry lock

eleanor-merry:
This PR removes the cache entry lock and makes it so that you have to be holding the cache lock in order to access any cache entries.

I've got a worry that this is less performant than the old (albeit broken) code. As there's just the one lock, any thread that's waiting on a pending result will get woken up by any release of the lock - which could happen multiple times before the cache entry is populated. It's the same worry that we have in the current DNSCachedResolver code (https://github.com/Metaswitch/cpp-common/blob/master/src/dnscachedresolver.cpp#L774).

I've only tested by running make full_test so far. This needs to be put on a system running load as well.

### DIFF
--- a/include/ttlcache.h
+++ b/include/ttlcache.h
@@ -72,20 +72,19 @@ public:
     pthread_condattr_t cond_attr;
     pthread_condattr_init(&cond_attr);
     pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC);
-    pthread_cond_init(&_cond, &cond_attr);
+    pthread_cond_init(&_result_available_cv, &cond_attr);
     pthread_condattr_destroy(&cond_attr);
   }
 
   ~TTLCache()
   {
     pthread_mutex_destroy(&_lock);
-    pthread_cond_destroy(&_cond);
+    pthread_cond_destroy(&_result_available_cv);
   }
 
   /// Get or create an entry in the cache.
   std::shared_ptr<V> get(K key, int& ttl, SAS::TrailId trail)
   {
-    bool retry_cache_lookup = true;
     std::shared_ptr<V> data_ptr = NULL;
 
     // Take the lock. We need to lock anytime we do anything with cache
@@ -95,8 +94,6 @@ public:
     // Evict any old entries.
     evict();
 
-    KeyMapIterator kmi = _cache.find(key);
-
     // We need to hold the lock if we're looking at cache entries, as otherwise
     // the evict method could be called which potentially destroys the entry.
     // However, we can't hold the lock while we're actually getting the DNS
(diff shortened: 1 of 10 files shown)


include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
+
+    // We need to hold the lock if we're looking at cache entries, as otherwise
+    // the evict method could be called which potentially destroys the entry.
+    // However, we can't hold the lock while we're actually getting the DNS
+    // record as this blocks for too long. Instead the logic is:
+    // - Look in the cache. The entry is either:
+    //   - Not present:
+    //     - Create an entry in the cache with the state set to PENDING.
+    //     - Store a shared_ptr to the DNS result. We have to release the lock
+    //       at this stage.
```
sathiyan-sivathas: What does release the lock _at_ this stage mean? Is that before or after the DNS lookup. I think it's before but that's not completely obvious. Can you make that clearer?

include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
+    //       no longer trust the entry we had at the start. Instead, re-get
+    //       the entry from the cache, creating it again if necessary. Set
+    //       the expiry time and the state to COMPLETE.
+    //     - Release the lock
+    //     - Return the shared_ptr
+    //   - Present in state pending:
+    //     - Wait on the lock. This releases the lock, then reclaims it when
+    //       it's next free
+    //     - As we've released the lock we can't look at the entry we had at
+    //       the start. Instead we just start again.
```
sathiyan-sivathas: I don't care if you use full stops or not, but at least do it consistently!

include/ttlcache.h
```diff
@@ -54,7 +54,6 @@ class TTLCache
   struct Entry
   {
     enum {PENDING, COMPLETE} state;
-    pthread_mutex_t lock;
```
AlexHockey: You've made the condition variable be at the scope of the cache rather than the entry. This means that when you do the broadcast you are going to wake up every thread that's waiting even if it's result isn't ready yet. Is that OK?

include/ttlcache.h
AlexHockey: Yeah I'm confused as well. How can you store a pointer to the DNS result when you haven't even done the query yet?

include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
+    //     - Store a shared_ptr to the DNS result. We have to release the lock
+    //       at this stage.
+    //     - We've now got the DNS result. As we'd released the lock we can
+    //       no longer trust the entry we had at the start. Instead, re-get
+    //       the entry from the cache, creating it again if necessary. Set
+    //       the expiry time and the state to COMPLETE.
+    //     - Release the lock
+    //     - Return the shared_ptr
+    //   - Present in state pending:
+    //     - Wait on the lock. This releases the lock, then reclaims it when
```
AlexHockey: Condition variable.

include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
+    //       it's next free
+    //     - As we've released the lock we can't look at the entry we had at
+    //       the start. Instead we just start again.
+    //     - There's an issue here where there's only one lock for every cache
+    //       entry. This means that this logic can be repeated many times before
+    //       there's a result
```
AlexHockey: Well that can happen anyway. Condition variables suffer from [spurious wakeups](https://en.wikipedia.org/wiki/Spurious_wakeup) so you always have to use a loop and recheck the condition after you wakeup.

Also, you use "lock" rather than "condition variable" again.

include/ttlcache.h
```diff
@@ -222,14 +246,45 @@ class TTLCache
...
+  {
+    KeyMapIterator kmi = _cache.find(key);
+
+    if (kmi == _cache.end())
+    {
+      entry = _cache[key];
+    }
+    else
+    {
+      entry = kmi->second;
```
AlexHockey: Erm, why do you pass entry in and then ignore it?

include/ttlcache.h
```diff
@@ -222,14 +246,45 @@ class TTLCache
...
-  /// Lock protecting the global structures in the cache.  This lock must be
-  /// held when accessing the global expiry list and key map structures.  It
-  /// must not be held when calling a factory get() method, but can be held
-  /// when calling an evict() method as these are assumed not to block.
+  /// Lock protecting the global structures in the cache. This lock must be
+  /// held when accessing the global expiry list, key map structures, or cache
+  /// entries. It must not be held when calling a factory get() method, but can
+  /// be held when calling an evict() method as these are assumed not to block.
   pthread_mutex_t _lock;
+  pthread_cond_t _cond;
```
AlexHockey: Worth a comment explaining what the cond var is for. Also you could give it a better name (e.g. `_result_available_cv`).

include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
+    while (retry_cache_lookup)
     {
-      TRC_DEBUG("Found the entry in the cache");
-      Entry& entry = i->second;
+      if (kmi == _cache.end())
+      {
+        TRC_DEBUG("Entry not in cache, so create new entry");
+        Entry& entry = _cache[key];
+        entry.state = Entry::PENDING;
+        entry.expiry_i = _expiry_list.end();
```
AlexHockey: I wonder if in the interests of safety we should not store a reference to the entry whenever we are about the drop the lock. That way there is no way to reference something that might be freed under your feet.

include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
-        // This cache entry is still being populated, so release the global
-        // lock and block on the entry's lock.
-        TRC_DEBUG("Cache entry is pending, so wait for the factory to complete");
-        pthread_mutex_lock(&entry.lock);
-        TRC_DEBUG("Entry is complete");
-
-        // The entry should now be complete, so release the lock on the entry.
-        pthread_mutex_unlock(&entry.lock);
+        TRC_DEBUG("Found the entry in the cache");
+        Entry& entry = kmi->second;
```
AlexHockey: This line is not obviously safe. `kmi` is initialized outside the loop, and if the loop is executed multiple times we will drop the lock while still keeping a reference. In fact, I all variables that are used inside the loop should be re-initialized on each loop iteration.

include/ttlcache.h
```diff
@@ -70,76 +69,101 @@ class TTLCache
...
+    //     - As we've released the lock we can't look at the entry we had at
+    //       the start. Instead we just start again.
+    //     - There's an issue here where there's only one lock for every cache
+    //       entry. This means that this logic can be repeated many times before
+    //       there's a result
+    //   - Present in state complete:
+    //     - Store off the shared_ptr to the DNS result in the entry
+    //     - Release the lock
+    //     - Return the shared_ptr
+    while (retry_cache_lookup)
```
AlexHockey: Do we *really* need a temporary variable here?

include/ttlcache.h
```diff
@@ -222,14 +246,45 @@ class TTLCache
     }
   }
 
+  // Populate the cache entry with the DNS record. There's a chance that the
+  // record has been deleted, so we might need to recreate the entry in the
+  // cache.
+  void populate_cache_entry(Entry& entry,
+                            K key,
```
AlexHockey: `K&` to avoid having to copy the key?

include/ttlcache.h
eleanor-merry: Well - that was the question in the PR.

It's similar to the code in the DNSCachedResolver - where we've accepted the cost of the extra lookups.

How can the condition variable be at the scope of the entry? Surely that's not safe in the same way that the lock can't be at the scope of the entry?

AlexHockey: The problem is not having locks or conditions at the scope of the entry - it's using references that you got before holding the lock after you have released and re-acquired it. 

The problem here though is that you cannot destroy a condvar while it is being waited on by another thread, and apparently you can't guarantee that by the time the broadcast returns that the other threads have become unblocked from the condvar. 

As the PR you linked to says, we can fix this by reference counting, which is why I suggested using shred pointers to do the ref counting for us. 

For now I'd suggest just merging this code as is as safe and slow is better than unsafe and fast.

PR #779, opened 2018-03-15
IO Trap Infrastructure. 

AlexHockey:
This PR contains the infrastructure to check that a process is correctly using the `CW_IO_STARTS` and `CW_IO_COMPLETES` macros. I'm calling this the "IO trap".

The idea is that:

* If a project wants to use the IO trap, it compiles `io_trap.cpp` into a separate shared object. 
* Any threads in that process that must use the `CW_IO_*` macros should call `CW_IO_CALLS_REQUIRED()` before doing any IO. This means the threads will be monitored by the IO trap. 
* When running the process you install the IO trap by using `LD_PRELOAD=/path/to/io_trap.so`.
* If one of the monitored threads does any IO without calling `CW_IO_*` the process will abort (and produce a core file and a stack trace). 

I've tested in sprout by hacking out various calls to `CW_IO_*` calls (specifically in the HTTP client, DNS resolver, and memcached client). When I run the live tests with the trap installed the process crashes and produces a stack trace. 

I also unit tested the infrastructure that is always compiled into the product code.

### DIFF
--- a/src/io_trap.cpp
+++ b/src/io_trap.cpp
@@ -9,6 +9,8 @@
  * Metaswitch Networks in a separate written agreement.
  */
 
+// LCOV_EXCL_START
+
 // C header files must be included with C linkage to prevent name mangling.
 extern "C" {
 #include <sys/types.h>
@@ -210,3 +212,5 @@ extern "C" int __poll(struct pollfd *fds, nfds_t nfds, int timeout)
 {
   HANDLE_NON_FD_CALL(__poll, fds, nfds, timeout);
 }
+
+// LCOV_EXCL_STOP
--- a/src/io_trap.cpp
+++ b/src/io_trap.cpp
@@ -1,0 +1,90 @@
+/**
+ * @file io_trap.cpp
+ *
+ * Copyright (C) Metaswitch Networks 2018
+ * If license terms are provided to you in a COPYING file in the root directory
+ * of the source code repository by which you are accessing this code, then
+ * the license outlined in that COPYING file applies to your use.
+ * Otherwise no rights are granted except for those provided to you by
+ * Metaswitch Networks in a separate written agreement.
+ */
+
+extern "C" {
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <stdlib.h>
+#include <dlfcn.h>
+}
+
+#include <atomic>
+#include "utils.h"
+
+static void maybe_abort()
+{
+  // Abort if we want to trap unexpected IO, and this IO is unexpected.
+  if (!Utils::IOMonitor::thread_doing_overt_io() &&
+      !Utils::IOMonitor::thread_allows_covert_io())
+  {
+    abort();
+  }
+}
+
+#define INTERPOSE(FUNCTION, ...)                                               \
+  using FunctionType = decltype(&FUNCTION);                                    \
+  static std::atomic<FunctionType> real_function(nullptr);                     \
+  if (!real_function)                                                          \
+  {                                                                            \
+    real_function = (FunctionType)dlsym(RTLD_NEXT, #FUNCTION);                 \
+  }                                                                            \
+                                                                               \
+  maybe_abort();                                                               \
+                                                                               \
+  return real_function(__VA_ARGS__)
+
+
+extern "C" ssize_t recv(int sockfd, void *buf, size_t len, int flags)
+{
+  INTERPOSE(recv, sockfd, buf, len, flags);
+}
+
+extern "C" ssize_t recvfrom(int sockfd, void *buf, size_t len, int flags,
+                            struct sockaddr *dest_addr, socklen_t* addrlen)
+{
+  INTERPOSE(recvfrom, sockfd, buf, len, flags, dest_addr, addrlen);
+}
+
+extern "C" ssize_t recvmsg(int sockfd, struct msghdr *msg, int flags)
+{
+  INTERPOSE(recvmsg, sockfd, msg, flags);
+}
+
+extern "C" ssize_t send(int sockfd, const void *buf, size_t len, int flags)
+{
+  INTERPOSE(send, sockfd, buf, len, flags);
+}
+
+extern "C" ssize_t sendto(int sockfd, const void *buf, size_t len, int flags,
+                          const struct sockaddr *dest_addr, socklen_t addrlen)
+{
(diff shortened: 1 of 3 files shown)


src/io_trap.cpp
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * Copyright (C) Metaswitch Networks 2018
+ * If license terms are provided to you in a COPYING file in the root directory
+ * of the source code repository by which you are accessing this code, then
+ * the license outlined in that COPYING file applies to your use.
+ * Otherwise no rights are granted except for those provided to you by
+ * Metaswitch Networks in a separate written agreement.
+ */
+
+// LCOV_EXCL_START
```
richardwhiuk: Do we need to exclude this from coverage?

It feels like we could either interpose, or trap abort and check that this works correctly? If we are going to expose this behind a shared config, I think we strongly should do that. If we are just expecting this to be niche and hard to enable, I'm less convinced, although I do worry about the decay factor.

src/io_trap.cpp
```diff
@@ -0,0 +1,216 @@
...
+/// block (e.g. is a socket is not in a particular state).
+static void about_to_block()
+{
+  // If the thread has not notified the IO monitor that it is about to do IO,
+  // and it is not allowed to do IO without notifying the IO monitor, we abort
+  // to generate a call stack and a core file.
+  if (!Utils::IOMonitor::thread_doing_overt_io() &&
+      !Utils::IOMonitor::thread_allows_covert_io())
+  {
+    abort();
```
richardwhiuk: Is it worth dropping an explicit printf here saying "Trapped IO called - aborting"?

We could also mention which of the two conditions we are in, which will be painful to determine from the core file.

There are other things we know (e.g. the function name could be passed to printf by using the `#` operator, but that's probably of little value given we'll definitely get the backtrace.

src/io_trap.cpp
```diff
@@ -0,0 +1,216 @@
...
+//
+// This macro does the following:
+//
+// -  Stores off a pointer to the real function being interposed.
+// -  Calls a function to handle the consequences of this function blocking.
+// -  Calls the real function.
+//
+// @param [in] FUNCTION - The function being interposed.
+// @param [in] ...      - The arguments passed to the function.
+#define HANDLE_NON_FD_CALL(FUNCTION, ...)                                      \
```
richardwhiuk: Generally you want to structure multi-line macros as `do { } while(0)` loops to avoid problems with early termination of if blocks, and to make them a single statement from a syntactical perspective - see https://wiki.sei.cmu.edu/confluence/display/c/PRE10-C.+Wrap+multistatement+macros+in+a+do-while+loop

It's less important here, because these are defined in a cpp file, and not a header file, but still good practice.

src/io_trap.cpp
```diff
@@ -0,0 +1,216 @@
...
+    abort();
+  }
+}
+
+
+// Helper macro to create a pointer to an interposed function and populate.
+//
+// @param [in] FUNCTION - The function being interposed.
+// @param [in[ VAR      - The variable in which to store the function pointer.
+#define STORE_REAL_FUNCTION(FUNCTION, VAR)                                     \
```
richardwhiuk: This is uglier than it could be. I don't see a particular need to store the function while we check whether we abort. I also don't like C++ variables having capital names, especially when that's the convention you are using for macro variables.

I think you can just do:

```c++
#define  RETURN_CALL_REAL_FUNCTION(FUNCTION, ...)  do {                        \
  using FunctionType = decltype(&FUNCTION);                                    \
  static std::atomic<FunctionType> func(nullptr);                              \
  if (!func)                                                                   \
  {                                                                            \
    func = (FunctionType)dlsym(RTLD_NEXT, #FUNCTION);                          \
  }                                                                            \
  return func(__VA_ARGS__);                                                    \
} while (0);
```

then your two macros below become:

```c++
#define HANDLE_NON_FD_CALL(...)  do {                                \
  about_to_block();                                                            \
  RETURN_CALL_REAL_FUNCTION(__VA_ARGS__)                             \
} while (0)

#define HANDLE_NON_FD_CALL(FUNCTION, FD, ...)  do {                                \
  if ((fcntl((FD), F_GETFL) & O_NONBLOCK) == 0)                                \
  {                                                                            \
    about_to_block();                                                          \
  }                                                                                \
  RETURN_CALL_REAL_FUNCTION(FUNCTION, FD, ## __VA_ARGS__)                             \
} while (0)
```

src/io_trap.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+extern "C" int accept4(int sockfd, struct sockaddr *addr, socklen_t *addrlen, int flags)
+{
+  HANDLE_FD_CALL(accept4, sockfd, addr, addrlen, flags);
+}
+
+extern "C" int select(int nfds, fd_set *readfds, fd_set *writefds,
+                      fd_set *exceptfds, struct timeval *timeout)
+{
+  HANDLE_NON_FD_CALL(select, nfds, readfds, writefds, exceptfds, timeout);
```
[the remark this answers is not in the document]
richardwhiuk: (this is a bit ugly, given some take `int timeout`, and others take `struct timespec* timeout`).

src/io_trap.cpp
AlexHockey: Plus they are only used in code that doesn't use single-statement ifs. But fair point, I will make this change.

src/io_trap.cpp
AlexHockey: I like it 😄

src/io_trap.cpp
AlexHockey: > We could also mention which of the two conditions we are in, which will be painful to determine from the core file.

What do you mean by this? We will only abort if both conditions are false. 

I'll add the first printf you suggest as that will make it clearer why we aborted. I don't think that printing out the function name is helpful - sure I'll know that `select` got called but it doesn't tell me why it was called. But the stack trace helps with that.

src/io_trap.cpp
AlexHockey: I ran this tool once and it didn't find anything, so I expect this is gong to fall into the category of "niche and hard to enable", and I'm not planning on adding any UTs. Happy?

src/io_trap.cpp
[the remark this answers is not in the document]
AlexHockey: I'm tempted to not bother until we know it's valuable. OK with you?

PR #781, opened 2018-03-16
Add PDLogs for when the sas.json file is invalid

jack-atack:
Adds PDlogs for parsing of invalid sas.json to flag when the on-the-box config file is awrt


ChrisRees:
We still need to add the ENT logs with details of which SAS connections we've lost the connection to.
But I'm expecting that in a separate PR.

